PR #6, opened 2023-01-20
Fix navigation.

### LINKED ISSUE #4: Navigation behaviour.
Right now, when navigating between bottom bar destinations, even when navigating to the same location again, it gets added to the back stack. Thus, when pressing the back button, it goes through all the locations passed by. I wonder whether it is a desirable behaviour, or whether it should be fixed.

KotlinGeekDev:
This fixes #4 .


KotlinGeekDev:
cc @vitorpamplona .

vitorpamplona:
Did you notice that with this change if you click twice on the bottom buttons the app scrolls to the top on all but the home page? I can't figure out why.

KotlinGeekDev:
Sorry @vitorpamplona just saw your comment now :( please was it fixed?

PR #63, opened 2023-02-02
Disable android backup as this app uses EncryptedSharedPrefs. Usage o…

clackbib:
Disable android backup as this app uses EncryptedSharedPrefs. Usage of EncryptedSharedPrefs can cause crashes if it's not properly excluded from backup.

https://stackoverflow.com/questions/61751264/autobackup-with-encryptedsharedpreferences-failing-to-restore


vitorpamplona:
Nice find, @clackbib

Issue #11, opened 2023-01-22
Profile button name "Post" should be "Save"


bitcoinuser:
Profile button name "Post" should be "Save"

![photo_2023-01-21_23-09-42](https://user-images.githubusercontent.com/1178856/213897488-9277e81c-3943-4405-b92b-585615e7d9b2.jpg)

maxmoney21m:
This issue can probably be closed?

PR #67, opened 2023-02-02
Add new button labeled 'Save'

kdunn926:
Addresses #11 . An alternative way is to generalize the `PostButton` into a `ConfirmButton` with display text as a parameter? That would collapse `Post`, `Create`, and now `Save` into a single function. I've never done Kotlin development so forgive me is this is sophomoric.


vitorpamplona:
But is it clear that Save is going to the published in the relay and not locally. That was why I kept Post, to make sure people know this is public.

kdunn926:
I hadn't considered the underlying "persistence" mechanism when opening this PR. I can see your point about using deliberate language to inform the user about both "what and how" rather than simply "what" is being done on that screen.

@bitcoinuser - what do you think?

dergigi:
I assume this is the same as #99?

kdunn926:
It seems like this may not be needed after all. I'm going to close the PR.

PR #84, opened 2023-02-06
Fix #83

### LINKED ISSUE #83: UI: Profile with long description lefts too little space for the content
When you open a profile having a long description, too little vertical space is left for the content. The profile header is not scrollable, so you have to view the content in a container too short or even not visible at all.

The orange arrow area is not scrollable. The blue arrow area is all the left to see the profile notes, followers, etc.

Radiokot:
This pull request fixes the issue #83 by making the profile header scrollable and hideable. Tabs, however, remain fixed

I used nested scrolling callback for this. Maybe there is a better way


vitorpamplona:
Nice! Thank you! I was struggling to find a solution here.

PR #118, opened 2023-02-13
Fix GIF image upload #115

### LINKED ISSUE #115: Image uploads get converted to .jpg
When you try to upload an animated gif the resulting link is to a jpg image, a single frame of the animated gif.

I know imgur supports animated gifs as well as some video formats. Can we preserve the filetype? That would enable us to post animated gifs and webms and the like. Those types already embed well on Amethyst

Radiokot:
- Image to upload is no more converted to a bitmap, which fixes #115 
- Image to upload is no more loaded into the memory
- The upload button indicates a running uploading
- An error message is shown if the uploading is failed

<p float="left">

</p>


vitorpamplona:
That was fast! Thank you!

PR #142, opened 2023-02-20
add leak canary to detect memory leaks in debug builds

terryschmidt:
this library does automatic memory leak detection only in debug builds. It's useful particularly because I've heard there were previous issues with OutOfMemory errors


terryschmidt:
https://square.github.io/leakcanary/

PR #167, opened 2023-03-01
WIP. Add search with NIP-50

brugeman:
This is an attempt to implement search for posts using NIP-50 like it's done in Snort/Iris. The only relay that supports that NIP properly atm is relay.nostr.band, so it had to be hard-coded.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -49,7 +49,7 @@ object LocalCache {
   val antiSpam = AntiSpamFilter()
 
   val users = ConcurrentHashMap<HexKey, User>()
-  val notes = ConcurrentHashMap<HexKey, Note>()
+  var notes = ConcurrentHashMap<HexKey, Note>()
   val channels = ConcurrentHashMap<HexKey, Channel>()
 
   fun checkGetOrCreateUser(key: String): User? {
@@ -85,7 +85,8 @@ object LocalCache {
   fun getOrCreateNote(idHex: String): Note {
     return notes[idHex] ?: run {
       val answer = Note(idHex)
-      notes.put(idHex, answer)
+      val newNotes = ConcurrentHashMap(mapOf(Pair(idHex, answer)) + notes)
+      notes = newNotes
       answer
     }
   }
@@ -146,7 +147,9 @@ object LocalCache {
       }
       return
     }
-
+    if(relay?.url == "wss://relay.nostr.band") {
+      Log.e("TAG", "event -> ${event.content}, -> ${event.id}")
+    }
     val note = getOrCreateNote(event.id.toHex())
     val author = getOrCreateUser(event.pubKey.toHexKey())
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
```diff
@@ -85,7 +85,8 @@ object LocalCache {
   fun getOrCreateNote(idHex: String): Note {
     return notes[idHex] ?: run {
       val answer = Note(idHex)
-      notes.put(idHex, answer)
+      val newNotes = ConcurrentHashMap(mapOf(Pair(idHex, answer)) + notes)
+      notes = newNotes
```
vitorpamplona, line 89: Is this change to immutable necessary? Rebuilding large maps (which is 10x heavier than an array copy) 1000s of times per second has been quite slow in entry-level phones in the past. 

That's why we use a Concurrent map for the main repository.

app/src/main/java/com/vitorpamplona/amethyst/service/relays/Client.kt
```diff
@@ -27,13 +27,21 @@ object Client: RelayPool.Listener {
     private var subscriptions = mapOf<String, List<TypedFilter>>()
 
     @Synchronized
-    fun connect(relays: Array<Relay>) {
+    fun connect(relays: Array<Relay>, searchRelays: List<Relay>) {
         RelayPool.register(this)
         RelayPool.unloadRelays()
         RelayPool.loadRelays(relays.toList())
+        RelayPool.loadSearchRelays(searchRelays)
         this.relays = relays
```
vitorpamplona, line 35: What if instead of an extra parameter, we had a new RelaySet Type for search? It would be similar to the Home, DM, Public Chat, Global split we have today, with an extra one for Search.

brugeman: I guess that's definitely the way to go in the future, but not sure it makes sense atm. There is only one such relay. 'Search' icon would occupy space in all other relays for nothing. Plus I didn't want users to be forced to have this relay added to their relay set, the way it's done now is that it's relied on only for search. 

Another way could be to have a 'search relay' setting that's pre-filled with this relay, and users could type another one, etc?

app/src/main/java/com/vitorpamplona/amethyst/service/relays/Relay.kt
```diff
@@ -76,9 +72,13 @@ class Relay(
                         val channel = msg[1].asString
                         when (type) {
                             "EVENT" -> {
-                                //Log.w("Relay", "Relay onEVENT $url, $channel")
-                                eventDownloadCounter++
```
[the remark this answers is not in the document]
brugeman: It went lower, after search events were redirected to proper listener.

app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
EvgenyPlaksin: For some reason, the elements were not added to the map in the previous way, when I tried to get the element by the desired id, I always returned null.  That's why I did it this way, and it solved the problem.

brugeman:
Thank you for the feedback, I think you're right. Marking it as WIP for now.

vitorpamplona:
Resolved in https://github.com/vitorpamplona/amethyst/commit/681e3eb44efe7ea6591815f3c29b851989e815e5

PR #168, opened 2023-03-01
Translated to brazilian portuguese

greenart7c3:


vitorpamplona:
Thank you @greenart7c3

PR #169, opened 2023-03-01
Update MainActivity.kt

rashedswen:
this line prevent app from crash when change language to Arabic

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/MainActivity.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/MainActivity.kt
@@ -62,6 +62,8 @@ class MainActivity : ComponentActivity() {
     }
 
     Client.lenient = true
+    
+    Locale.setDefault(Locale.ENGLISH)
   }
 
   override fun onResume() {


vitorpamplona:
I don't get it. Why is this crashing? It forces english, which then disables the translations, no?

rashedswen:
in my case it doesn't force English 

app crashing cause RoboHash return asset string in Arabic ( `sets/set1/red/01Body/red_body-٠٢.png` ) not English ( `sets/set1/red/01Body/red_body-02.png` ) this cause exception file not found.

vitorpamplona:
Well, if we force English then Autotranslations of the post fail... We have to figure out how to fix this robohash issue

rashedswen:
ok we need to focus in this line and functions behave it 

`      defaultAvatar = robots.imageForHandle(robots.calculateHandleFromUUID(UUID.nameUUIDFromBytes("aaaa".toByteArray()))).asImageBitmap()
`

vitorpamplona:
I just sent a quick fix for RoboHash, see if it still crashes

rashedswen:
nice this work well now

rashedswen:
thanks for all your hard work

vitorpamplona:
What's your npub, @rashedswen?

rashedswen:
sorry for late replay 

npub1twcvdqpwkgfvw7dakadzp5ehrde2hzrvgs8nexzpy3de304pfjzs3dy8z2

Issue #172, opened 2023-03-02
[FEATURE] Mark All Read for Direct Messages


derekross:
**Is your feature request related to a problem? Please describe.**
Amethyst does not have a feature to allow the user to mark all direct messages as read. All DMs that a user has previously read on other clients remain with a notification bubble as un-read even though they've been read. This makes it hard to know when you receive a new DM because a user could potentially have hundreds of DMs that have been previously read on another client.  Because of this, users have a constant notification bubble that won't go away.

**Describe the solution you'd like**
Implement a button in the Direct Message tab that marks all DMs as read. Tapping this button marks all DMs as read as seen on other clients like Snort and Astral.

**Bounty (in Bitcoin sats) offered for the implementation**
I'll start the bounty off at 50K sats and post on Nostr to raise more awareness.

maxmoney21m:
Considering taking a stab at this. Do you (or @vitorpamplona) think this would be better as:

1. A simple persistent button that shows whenever you have more than some threshold of unread messages (e.g. 10)
2. A popup button (style similar to zap amount selection buttons) that shows on long pressing any message. This method could allow for additional actions, perhaps: Mark Read, Mark All Read, Hide

vitorpamplona:
Thanks @maxmoney21m! 

Since this is for all DMs at once, the easiest design would be an action on the Tab itself that simply marks all DMs on that tab as Read. 

The most flexible design would be a per-row selector when you press and hold each user. In that way, users could mark many of them and then hit Mark all as Read in a similar way Telegram and WhatsApp do. 

In both cases, we could use the 3-dot menu icon after the time, in the same way we use on the home page, to also have these options (so people don't have to guess that press and hold does something)

derekross:
I wouldn't want to have to select 500 individual rows to mark all of my DMs as read. It's better than having to open 500 DMs and scroll down through each one, but it's still a very time consuming task that I just wouldn't do.

vitorpamplona:
Agree, looks like we are doing an action at the tab level, @maxmoney21m

maxmoney21m:
Reference images @vitorpamplona 

![image](https://user-images.githubusercontent.com/124794312/222466233-7bb7b9dc-cd1b-4001-8116-905979b88150.png)

![image](https://user-images.githubusercontent.com/124794312/222466281-05fd9dd5-e299-48e6-8d0e-66b94398ad82.png)

efraimbart:
Alternatively:

[...]

PR #174, opened 2023-03-02
WIP solution for #172

maxmoney21m:
This is the most direct way I could find to implement this. I separately tried going through the viewmodel layer, and that worked as well but I figured it'd be better to keep things simple unless you want it to go that route @vitorpamplona.

This almost works, the only thing I can't seem to get is how to update the state directly after clicking the button. If you click the menu button and then navigate away and back, the notifications are cleared as expected.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
@@ -1,31 +1,41 @@
 package com.vitorpamplona.amethyst.ui.screen.loggedIn
 
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
```diff
@@ -120,7 +157,37 @@ fun TabNew(accountViewModel: AccountViewModel, navController: NavController) {
...
+    onDismiss: () -> Unit,
+    onMarkKnownAsRead: () -> Unit,
+    onMarkNewAsRead: () -> Unit,
+) {
+    DropdownMenu(expanded = expanded, onDismissRequest = onDismiss) {
+        DropdownMenuItem(onClick = {
+            onMarkKnownAsRead()
+            onDismiss()
+        }) {
+            Text("Mark all Known as read")
```
maxmoney21m, line 177: Need to extract string resources

PR #177, opened 2023-03-03
Fix #172, add "Mark all as read" feature for DMs and chat rooms

### LINKED ISSUE #172: [FEATURE] Mark All Read for Direct Messages

maxmoney21m:
This solution is currently working, but there are at least two improvements that could be made:

1. Extracting the `route` name calculations, which are duplicated here from `ChatroomCompose.kt`. At the moment there is no obvious place to extract them.
2. Putting the `mark___AsRead` flags into the state management layer. For now they are just `MutableState` of the Composable and modified directly, which might be OK as they are just simple booleans.


app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
```diff
@@ -120,7 +157,37 @@ fun TabNew(accountViewModel: AccountViewModel, navController: NavController) {
...
+    onDismiss: () -> Unit,
+    onMarkKnownAsRead: () -> Unit,
+    onMarkNewAsRead: () -> Unit,
+) {
+    DropdownMenu(expanded = expanded, onDismissRequest = onDismiss) {
+        DropdownMenuItem(onClick = {
+            onMarkKnownAsRead()
+            onDismiss()
+        }) {
+            Text("Mark all Known as read")
```
maxmoney21m, line 177: Still need to extract string resources

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
```diff
@@ -1,88 +1,128 @@
...
-                selectedTabIndex = pagerState.currentPage,
-                indicator = { tabPositions ->
-                    TabRowDefaults.Indicator(
-                        Modifier.pagerTabIndicatorOffset(pagerState, tabPositions),
-                        color = MaterialTheme.colors.primary
-                    )
-                },
+    var moreActionsExpanded by remember { mutableStateOf(false) }
+    val markKnownAsRead = remember { mutableStateOf(false) }
+    val markNewAsRead = remember { mutableStateOf(false) }
```
maxmoney21m, line 56: We use `=` here to keep the mutability so that we can reset the values once they have been used.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/ChatroomListScreen.kt
```diff
@@ -1,88 +1,128 @@
...
+
+    Box(Modifier.fillMaxSize()) {
```
maxmoney21m, line 58: The `Column` layout is wrapped in a `Box` so that we can put the 3 dots menu overlaid on the top right of the screen.

app/src/main/res/values/strings.xml
```diff
@@ -177,4 +177,7 @@
     <string name="report_hateful_speech">Report Hateful speech</string>
     <string name="report_nudity_porn">Report Nudity / Porn</string>
     <string name="others">others</string>
+    <string name="mark_all_known_as_read">Mark all Known as read</string>
+    <string name="mark_all_new_as_read">Mark all New as read</string>
+    <string name="mark_all_as_read">Mark all as read</string>
```
maxmoney21m, line 182: Need translations.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -73,18 +88,53 @@ fun ChatroomListFeedView(viewModel: FeedViewModel, accountViewModel: AccountView
...
+                    } else {
+                        val replyAuthorBase = note.mentions?.first()
+                        var userToComposeOn = note.author!!
+                        if (replyAuthorBase != null) {
+                            if (note.author == account.userProfile()) {
+                                userToComposeOn = replyAuthorBase
+                            }
+                        }
+                        route = "Room/${userToComposeOn.pubkeyHex}"
+                    }
```
maxmoney21m, line 119: This code is duplicated from `ChatroomCompose`, but at the moment there is no obvious place to extract it for reuse.

app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
```diff
@@ -14,104 +14,109 @@ import nostr.postr.events.Event.Companion.getRefinedEvent
...
-      account.translateTo.let { putString("translateTo", it) }
-      account.zapAmountChoices.let { putString("zapAmounts", gson.toJson(it)) }
-      account.backupContactList.let { putString("latestContactList", Event.gson.toJson(it)) }
-    }.apply()
-  }
+    fun clearEncryptedStorage() {
+        encryptedPreferences.edit().apply {
+            encryptedPreferences.all.keys.forEach { remove(it) }
+        }.apply()
+    }
```
maxmoney21m, line 38: As part of the existing PR, I noticed that logging out and logging back in resulted in the notification dots staying hidden. Traced that to the `clearEncryptedStorage` method not properly clearing those keys. Rather than listing all the keys out, we can just remove all of them.

The following two commits f34dea8 and 75901d8 are just cleanup while I was touching the file.

maxmoney21m:
> The function should be just one and done. It should allow new notifications to be unread after it finishes marking everything as read.

The `markAsRead` parameter in `ChatroomListFeedView` is reset to `false` after the `LaunchEffect` runs once, which prevents it from running again unless the button is clicked again.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -99,8 +99,8 @@ private fun FeedLoaded(
     val notificationCache = notificationCacheState.value ?: return
     val context = LocalContext.current.applicationContext
 
-    if (markAsRead.value) {
-        LaunchedEffect(key1 = notificationCache) {
+    LaunchedEffect(key1 = markAsRead.value) {
+        if (markAsRead.value) {
```
maxmoney21m, line 103: Swapped the key to use the boolean flag directly for clarity. The behavior appears to be identical, including the number of calls and recompositions, etc.

app/src/main/java/com/vitorpamplona/amethyst/ui/note/ChatroomCompose.kt
```diff
@@ -82,7 +82,7 @@ fun ChatroomCompose(baseNote: Note, accountViewModel: AccountViewModel, navContr
         channel?.let { channel ->
             var hasNewMessages by remember { mutableStateOf<Boolean>(false) }
 
-            LaunchedEffect(key1 = notificationCache) {
+            LaunchedEffect(key1 = notificationCache, key2 = note) {
```
maxmoney21m, line 85: There was a pre-existing bug I noticed when coding this PR, that is, when a new message comes in while you are looking at the messages tab, the notification bubble does not re-appear unless you leave and come back, either by going to the other tab (Known/New), or going to another screen, etc.

If we key off of the note directly, the recomposition causes `hasNewMessages` to be re-evaluated with the new note so that the bubble can come back straight away.

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -254,7 +254,7 @@ fun ListContent(
         Spacer(modifier = Modifier.weight(1f))
 
         IconRow(
-            "Logout",
+            stringResource(R.string.log_out),
```
maxmoney21m, line 257: Since #182 was already merged before I could fix this and it's just a 1-liner, I am dropping it here for simplicity.

app/src/main/java/com/vitorpamplona/amethyst/ui/note/ChatroomCompose.kt
vitorpamplona: Nice find!

PR #178, opened 2023-03-03
Basic mark all as read implementation

efraimbart:
This implementation attempts to mark all as read within the composable context.

#172

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
@@ -4,21 +4,31 @@ import androidx.compose.animation.Crossfade
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.navigation.NavController
 import com.google.accompanist.swiperefresh.SwipeRefresh
 import com.google.accompanist.swiperefresh.rememberSwipeRefreshState
+import com.vitorpamplona.amethyst.NotificationCache
+import com.vitorpamplona.amethyst.R
 import com.vitorpamplona.amethyst.ui.note.ChatroomCompose
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.AccountViewModel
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+                    if (note.author == account.userProfile()) {
+                        userToComposeOn = replyAuthorBase
+                    }
+                }
+
+                "Room/${userToComposeOn.pubkeyHex}"
+            }
+
+            routeForLastRead?.let {
+                LaunchedEffect(key1 = notificationCache) {
```
efraimbart, line 157: I'm not sure what `LaunchedEffect` does or what should be used as the key.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+                    val createdAt = note.event?.createdAt
+                    if (createdAt != null) {
+                        NotificationCache.markAsRead(it, createdAt, context)
```
efraimbart, line 160: Should we consider marking as read with the current time instead of created at?

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+                    }
+                }
+            }
         }
     }
+
+
+    if (markedAllAsRead) {
+        markAllAsRead()
```
efraimbart, line 169: Should this flip `markedAllAsRead` back to false when complete?

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+                    item,
+                    accountViewModel = accountViewModel,
+                    navController = navController
+                )
+            }
+        }
+    }
+
+    @Composable
+    fun markAllAsRead() {
```
efraimbart, line 119: Not sure what the best placement for this function is, here? Same file but not nested? Extracted to another file?

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+    var markedAllAsRead by remember { mutableStateOf<Boolean>(false) }
+
+    Column {
+        TextButton(
+            modifier = Modifier.fillMaxWidth(),
+            shape = MaterialTheme.shapes.large,
+            onClick = { markedAllAsRead = true }
+        ) {
+            Text(stringResource(R.string.mark_all_as_read))
+        }
```
efraimbart, line 99: This is mostly a proof of concept, it could use a bit of design work and perhaps an if statement.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+                var userToComposeOn = note.author!!
+
+                if (replyAuthorBase != null) {
+                    if (note.author == account.userProfile()) {
+                        userToComposeOn = replyAuthorBase
+                    }
+                }
+
+                "Room/${userToComposeOn.pubkeyHex}"
+            }
```
efraimbart, line 154: Some of this is duplicated from elsewhere, consider extracting and reusing.

app/src/main/res/values-ar/strings.xml
```diff
@@ -173,4 +173,5 @@
     <string name="report_hateful_speech">أبلغ عن كلام يحض على الكراهية</string>
     <string name="report_nudity_porn">الإبلاغ عن عُري / إباحي</string>
     <string name="others">آخرون</string>
+    <string name="mark_all_as_read">Mark All As Read</string>
```
efraimbart, line 176: Needs translation

app/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -173,4 +173,5 @@
     <string name="report_hateful_speech">Denunciar discurso de ódio</string>
     <string name="report_nudity_porn">Denunciar nudez / pornografia</string>
     <string name="others">outros</string>
+    <string name="mark_all_as_read">Mark All As Read</string>
```
efraimbart, line 176: Needs translation

maxmoney21m:
Did you have any ideas for where to put the markAsRead route names so that they don't need to be duplicated here? My PR has the same problem.

efraimbart:
> Did you have any ideas for where to put the markAsRead route names so that they don't need to be duplicated here? My PR has the same problem.

@maxmoney21m What's the problem with the route names? They seem to be used like this across the board.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
```diff
@@ -77,19 +87,85 @@ private fun FeedLoaded(
...
+        val notificationCacheState = NotificationCache.live.observeAsState()
+        val notificationCache = notificationCacheState.value ?: return
+
+        val context = LocalContext.current.applicationContext
+
+        state.feed.value.forEach { baseNote ->
+            val noteState by baseNote.live().metadata.observeAsState()
+            val note = noteState?.note
+
+            val routeForLastRead = if (note?.event == null) {
```
efraimbart, line 132: Should probably change this to `routeForMarkAsRead`.

maxmoney21m:
> > Did you have any ideas for where to put the markAsRead route names so that they don't need to be duplicated here? My PR has the same problem.
> 
> @maxmoney21m What's the problem with the route names? They seem to be used like this across the board.

Not a problem per se, but just the code duplication, as you mentioned also in your comment on [ChatroomListFeedView.kt](https://github.com/vitorpamplona/amethyst/pull/178/files/7b42183f76e2719c1d0b6b282f74d537b5c0f7b7#diff-63c91192dfb7d3c678ad33f6a9ce2c3c9a7e7ed45832979fea8c00d0c5ab5751).

efraimbart:
> > > Did you have any ideas for where to put the markAsRead route names so that they don't need to be duplicated here? My PR has the same problem.
> > 
> > 
> > @maxmoney21m What's the problem with the route names? They seem to be used like this across the board.
> 
> Not a problem per se, but just the code duplication, as you mentioned also in your comment on [ChatroomListFeedView.kt](https://github.com/vitorpamplona/amethyst/pull/178/files/7b42183f76e2719c1d0b6b282f74d537b5c0f7b7#diff-63c91192dfb7d3c678ad33f6a9ce2c3c9a7e7ed45832979fea8c00d0c5ab5751).

Oh yeah, that would definitely be nice to consolidate.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
maxmoney21m: It allows running a side effect suspend function tied to the composable's lifecycle. The key ensures that it only runs once when the key changes. Since we only want it to run if the notification cache changes, this is correct.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
maxmoney21m: No, because you could see a new message later which has a createdAt time earlier than that timestamp, but you would not want to mark that one as read yet.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
maxmoney21m: Did not try your implementation yet, does it properly show notification dots for new messages that come in after you've flipped this to true?

efraimbart:
@maxmoney21m In response to your question, I was originally considering extracting some of the logic to Note.kt. Might be the best place for it currently.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
efraimbart: The same issue applies regardless of if we mark with the current time or with created at.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/ChatroomListFeedView.kt
efraimbart: Not without recomposition (if that's the term?), but that's also the current behavior on live once you've marked a chat as read the current way.

efraimbart: My other pull request (https://github.com/vitorpamplona/amethyst/pull/179) seems to resolve this though.

vitorpamplona:
Resolved on https://github.com/vitorpamplona/amethyst/pull/177

PR #186, opened 2023-03-03
temporary solution for mirroring arabic text

rashedswen:


vitorpamplona:
Did you try something like: 

```
   public static boolean isRTL(Locale locale) {
        final int directionality = Character.getDirectionality(locale.getDisplayName().charAt(0));
        return directionality == Character.DIRECTIONALITY_RIGHT_TO_LEFT ||
                directionality == Character.DIRECTIONALITY_RIGHT_TO_LEFT_ARABIC;
    }
```

rashedswen:
no, because sometimes notes will not translated so is not sense to force reverse to not RTL text

PR #182, opened 2023-03-03
Move nsec backup to drawer and organize drawer

maxmoney21m:
Regarding the first step mentioned in #59, this PR moves the nsec backup from the profile page to a separate screen using a Dialog in the drawer. I also organized the Drawer slightly and added some icons. See screenshots for reference.

There is no biometric check yet, but if the PR is OK so far I can add that.

image

image

image

PR #191, opened 2023-03-04
Require biometric with device lock fallback to copy nsec

maxmoney21m:
Following on #182, this PR prompts the user to authenticate after they click the "Copy nsec" button on the backup dialog.

- If user has fingerprint enabled, authenticate by fingerprint, otherwise by device lock
- From fingerprint screen, user can switch to device lock (PIN etc) if desired
- If fingerprint doesn't work, it falls back to PIN after the standard number of failures
- If no fingerprint or device lock mechanisms are in place, just copy the key anyway

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -73,6 +73,9 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-runtime-ktx:$lifecycle_version"
     implementation "androidx.lifecycle:lifecycle-livedata:$lifecycle_version"
 
+    // Biometrics
+    implementation "androidx.biometric:biometric-ktx:1.2.0-alpha05"
+
     // Swipe Refresh
     implementation 'com.google.accompanist:accompanist-swiperefresh:0.29.1-alpha'
 
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/MainActivity.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/MainActivity.kt
@@ -2,12 +2,12 @@ package com.vitorpamplona.amethyst.ui
 
 import android.os.Build.VERSION.SDK_INT
 import android.os.Bundle
-import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
 import androidx.compose.ui.Modifier
+import androidx.fragment.app.FragmentActivity
 import androidx.lifecycle.viewmodel.compose.viewModel
 import coil.Coil
 import coil.ImageLoader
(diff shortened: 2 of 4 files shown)


app/build.gradle
```diff
@@ -73,6 +73,9 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-runtime-ktx:$lifecycle_version"
     implementation "androidx.lifecycle:lifecycle-livedata:$lifecycle_version"
 
+    // Biometrics
+    implementation "androidx.biometric:biometric-ktx:1.2.0-alpha05"
```
maxmoney21m, line 77: Add biometric library

app/src/main/java/com/vitorpamplona/amethyst/ui/MainActivity.kt
```diff
@@ -22,7 +22,7 @@ import com.vitorpamplona.amethyst.ui.screen.AccountScreen
 import com.vitorpamplona.amethyst.ui.screen.AccountStateViewModel
 import com.vitorpamplona.amethyst.ui.theme.AmethystTheme
 
-class MainActivity : ComponentActivity() {
+class MainActivity : FragmentActivity() {
```
maxmoney21m, line 25: Biometrics are *only* available in FragmentActivity. Seems to be a hard requirement, no response from Google on issues open for more than 2 years already.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
```diff
@@ -121,3 +121,77 @@ private fun NSecCopyButton(
...
+fun Context.getFragmentActivity(): FragmentActivity? {
+    var currentContext = this
+    while (currentContext is ContextWrapper) {
+        if (currentContext is FragmentActivity) {
+            return currentContext
+        }
+        currentContext = currentContext.baseContext
+    }
+    return null
+}
```
maxmoney21m, line 134: We need to unwrap the wrapped composable context to get the fragment back, since the Biometrics accepts no other type of context.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
```diff
@@ -121,3 +121,77 @@ private fun NSecCopyButton(
...
+
+private fun authenticatedCopyNSec(
+    context: Context,
+    scope: CoroutineScope,
+    account: Account,
+    clipboardManager: ClipboardManager,
+) {
+    val fragmentContext = context.getFragmentActivity()!!
+    val authenticators = BIOMETRIC_STRONG or DEVICE_CREDENTIAL
```
maxmoney21m, line 143: `BIOMETRIC_STRONG` = fingerprint
`DEVICE_CREDENTIAL` = PIN or password

There is also `BIOMETRIC_WEAK` for face recognition, but I think the best practice would be to use only strong authentication for the nsec whenever possible.

Issue #183, opened 2023-03-03
Global feed icon


maxmoney21m:
The global feed is currently accessed through the magnifying glass icon at the bottom. This is typically the search icon. While the screen does have search, new users are often unaware that it is also the global feed.

I suggest replacing the magnifying glass with a globe icon to make this clear. The search bar is rather obvious once you are on the screen, and probably doesn't need it's own icon at the bottom.

3j2009:
I agree, globe icon for global feed is more intuitive (Jacob's Law).

Attached files are OS & royalty free.

![Global feed icon](https://user-images.githubusercontent.com/126127093/222864016-f40848b6-a400-479f-88cc-5ee4022415a8.png)
![Global feed icon](https://user-images.githubusercontent.com/126127093/222864020-b6c5f858-1d65-4451-a0d0-15250157857c.svg)
![Screenshot](https://user-images.githubusercontent.com/126127093/222864022-80e225a5-5c11-4409-b8ca-46341459ce6c.png)

vitorpamplona:
Looks like the icon fits really well with the other icons! We will use. Thank you!

PR #192, opened 2023-03-04
Change magnifying glass to globe for global/search tab on AppBottomBar

maxmoney21m:
Created properly sized pngs using svg provided by 3j2009 in #183

Reference screenshot:

image


vitorpamplona:
This is so much better...

3j2009:
looks great!

braydonf:
FWIW: For those that turn off all global so as to not be spammed, this tab really only functions to add new public keys.

maxmoney21m:
> FWIW: For those that turn off all global so as to not be spammed, this tab really only functions to add new public keys.

You can add paid relays as read-only for free. Try adding wss://eden.nostr.land and see if your global isn't spam-free.

Issue #159, opened 2023-02-27
zaps list: combine zaps from same user OR show related note of zap if not combining


KKA11010:
Hello,

i wanted to make this suggestion of combining zaps from the same user into one list entry. Also I am wondering if it would be possible to have the zaps separated like it is now but to show the related post after clicking on the zap list entry and not just the user profile.

![photo_2023-02-27_19-12-38](https://user-images.githubusercontent.com/40151034/221649860-b827e523-fc5b-4d74-a4e6-1cffa53947d0.jpg)

vitorpamplona:
This was implemented a couple of versions ago

PR #205, opened 2023-03-05
Refactor UserProfileZapsFeedFilter

Chemaclass:
## 🤔   Background

Extract the logic from `UserProfileZapsFeedFilter` into an isolated class, so we can write unit tests for it. 

This is inspired/related to https://github.com/vitorpamplona/amethyst/issues/159 . In this PR I didn't solve that issue, but we are getting closer to make it happen.

## 💡 Goal

Make the code more testable and add more automated tests.

## 📖 Changes

- Move the logic from the `UserProfileZapsFeedFilter` to a separated class `UserZaps.forProfileFeed()`
- Create `EventInterface` and `LnZapEventInterface`
  - Using the interface methods instead of the event's public values let the code behaving the same way as before
- Install "[mockk](https://mockk.io/)" to create mocks in kotlin test
- Added some characterization tests for `UserZapsTest`

The creation of the interfaces (`EventInterface`, `LnZapEventInterface`) is important in order to invert the dependencies and mock the `Secp256k1` behaviour [See Event:143]:

```kotlin
private val secp256k1 = Secp256k1.get()
```

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" project-jdk-name="Android Studio default JDK" project-jdk-type="JavaSDK">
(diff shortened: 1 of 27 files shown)


vitorpamplona:
Nice work!

PR #208, opened 2023-03-06
Biometric support down to app's minimum SDK level 26, fixes #204

### LINKED ISSUE #204: [BUG] Fingerprint crashes on Android 9
From report on nostr:

> Amethyst now crashes when I try to backup nsec private key on my old andoid 9 Samsung phone...

The problem is here:

https://github.com/vitorpamplona/amethyst/blob/main/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt#L176

From the [docs](https://developer.android.com/training/sign-in/biometric-auth):

> The following combinations of authenticator types aren't supported on Android 10 (API level 29) and lower: DEVICE_CREDENTIAL and BIOMETRIC_STRONG | DEVICE_CREDENTIAL. To check for the presence of a PIN, pattern, or password on Android 10 and lower, use the KeyguardManager.isDeviceSecure() method.

The quick fix will be to disable this logic for Android <= 10. Proper fix would be to have separate logic as described in the docs for those versions. If someone doesn't get to it first I will do when I have time.

maxmoney21m:
Needed to add a fallback to the `KeyguardManager` to support SDK below level 30 (Android 9 and below).

Fingerprint with PIN unlock and PIN-only (no fingerprint set) tested and working on emulators for SDKs 26, 28, 29, 30, 33.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
@@ -1,8 +1,16 @@
 package com.vitorpamplona.amethyst.ui.screen.loggedIn
 
+import android.app.Activity
+import android.app.KeyguardManager
 import android.content.Context
 import android.content.ContextWrapper
+import android.content.Intent
+import android.os.Build
 import android.widget.Toast
+import androidx.activity.compose.ManagedActivityResultLauncher
+import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.activity.result.ActivityResult
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.biometric.BiometricManager
 import androidx.biometric.BiometricManager.Authenticators.BIOMETRIC_STRONG
 import androidx.biometric.BiometricManager.Authenticators.DEVICE_CREDENTIAL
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
```diff
@@ -105,10 +113,17 @@ private fun NSecCopyButton(
     val context = LocalContext.current
     val scope = rememberCoroutineScope()
 
+    val keyguardLauncher =
+        rememberLauncherForActivityResult(ActivityResultContracts.StartActivityForResult()) { result: ActivityResult ->
+            if (result.resultCode == Activity.RESULT_OK) {
+                copyNSec(context, scope, account, clipboardManager)
+            }
+        }
```
maxmoney21m, line 121: Use the old-fashioned `StartActivityForResult` and pass a `KeyguardManager` intent on older devices.

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountBackupDialog.kt
```diff
@@ -141,10 +156,33 @@ private fun authenticatedCopyNSec(
...
+            context.getString(R.string.copy_my_secret_key),
+        )
+
+        keyguardLauncher.launch(intent)
+    }
+
+    if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.Q) {
+        keyguardPrompt()
+        return
+    }
```
maxmoney21m, line 182: Use the KeyguardManager on older devices. Only pass through this to the BiometricPrompt above version Q (= 29).

PR #219, opened 2023-03-07
Added Privacy on Relays to Readme.MD

alltheseas:
Added a small section on user privacy implications on Amethyst & relays use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,10 +121,12 @@ openssl base64 < <my-release-key.keystore> | tr -d '\n' | tee some_signing_key.j
 7. Add your CHANGE LOG to the description of the new release
 8. Download the `aab` file and upload it to the` PlayStore.
 
-# Privacy on Relays
+# Privacy on Relays & nostr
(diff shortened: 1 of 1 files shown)


vitorpamplona:
We should probably make this broader. Relays have all your data in raw text. They know your IP, your name, your location (guessed from IP), your pub key, all your contacts, and other relays, and can read every action you do (post, like, boost, quote, report, etc) with the exception of Private Zaps and Private DMs.

maxmoney21m:
Also important IMO to point out that:

- Content posted to one relay can (and probably will) be shared to other relays. Those relays will not necessarily see your IP address, but they will see some or all of the other information listed.
- That information is publicly visible to anyone else reading from any of the relays that have your information. The relays keep it, but anyone can look it up.
- There is no way to guarantee deleting or editing any content once posted. Any posts or actions should be considered permanent (on any client, whether or not the client has a delete or edit button).

PR #220, opened 2023-03-07
Add ktlint

Chemaclass:
## 📚 Description

A [linter](https://en.wikipedia.org/wiki/Lint_(software)) is basically a small program that checks no errors/bugs are introduced, additionally, it defines a set of rules that the code must follow. For Kotlin, [ktlint](https://github.com/pinterest/ktlint) is the most popular.

**Why?** This is really helpful to create a convention about the code style over the whole project, ideal for an OSS where many different people can contribute, so there is a uniform code style everywhere.

## 📖  Changes

- Install ktlint (kotlin linter) plugin
- Add a git pre-commit hook when building the project
  - Run linter checker when doing a commit, so it forces you to keep the same code style 
- The build CI will check the code style
- Apply the linter to the whole project

> **Note**: If you want to skip the linter (while doing a commit), you can add the --no-verify (-n) option when doing your commit. Eg: `git commit -m "your message" --no-verify`.


app/build.gradle
```diff
@@ -1,6 +1,7 @@
 plugins {
     id 'com.android.application'
     id 'org.jetbrains.kotlin.android'
+    id 'org.jlleitschuh.gradle.ktlint' version "11.3.1"
```
Chemaclass, line 4: This is the actual change: adding the `ktlint` plugin. 
The rest of the file-changes are just after applying the linter behavior.

PR #226, opened 2023-03-08
Redesign note long click menu, URL sharing

maxmoney21m:
This is a first pass at a new style of quick action menu for taking actions on notes. Right now it has 5 buttons, the but this could easily go to 6 or 8 as you can imagine. Next steps would be to add a screen for the Block/Report feature and add a button to that. Ultimately I would like to remove the 3-dots menu on the note if others agree. Note that the Mention and Quote buttons include the `@` in the copied text.

Here is how the quick action menu looks:

image

Pressing Select Text button gives me this popup for easy text selection:

image

(The "Select All" button is missing, apparently this is a known issue in Compose at the moment. I didn't find an easy fix.)

Share button gives me the URL to copy or put in another app:

image

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
@@ -25,6 +25,7 @@ class LocalPreferences(context: Context) {
         const val TRANSLATE_TO = "translateTo"
         const val ZAP_AMOUNTS = "zapAmounts"
         const val LATEST_CONTACT_LIST = "latestContactList"
+        const val HIDE_DELETE_REQUEST_INFO = "hideDeleteRequestInfo"
         val LAST_READ: (String) -> String = { route -> "last_read_route_$route" }
     }
 
@@ -49,6 +50,7 @@ class LocalPreferences(context: Context) {
             account.translateTo.let { putString(PrefKeys.TRANSLATE_TO, it) }
             account.zapAmountChoices.let { putString(PrefKeys.ZAP_AMOUNTS, gson.toJson(it)) }
             account.backupContactList.let { putString(PrefKeys.LATEST_CONTACT_LIST, Event.gson.toJson(it)) }
+            putBoolean(PrefKeys.HIDE_DELETE_REQUEST_INFO, account.hideDeleteRequestInfo)
         }.apply()
     }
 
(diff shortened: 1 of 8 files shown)


maxmoney21m:
> I think the 3-dots must stay. Lots of people don't realize they can click and hold for the menu options, including PlayStore reviewers that need to find the Block and Report options very easily. Otherwise, they reject the update.

That makes sense. In that case I suggest keeping the more “advanced” features like Broadcast in the menu, in addition to the Report button, and using the quick action menu for more common things. If we ever get one of those “first use tutorial” type flows, it would be good to call attention to the long press action so more people are aware of it.

@vitorpamplona I was considering how to do the Follow quick action button, it's easy enough for the general case but do you have any suggestions for what should happen when the user is already followed or it's one's own note? It's possible to just remove it, but that would change the position of any buttons that came after.

vitorpamplona:
Follow/Unfollow? It would be the same as the buttons in the Profile Page. 

People seem to use this button in the Global/Search view to quickly follow a list of people.

maxmoney21m:
> Follow/Unfollow? It would be the same as the buttons in the Profile Page.
> 
> People seem to use this button in the Global/Search view to quickly follow a list of people.

What about on your own note?

maxmoney21m:
> Delete? Feel right...

Great. I will push it in a bit.

maxmoney21m:
@vitorpamplona there are two commits here, the first (83841d5) includes the complete button functionality for Follow/Unfollow/Delete. It is usable if you don't want the second. The second (10a692d) includes a dialog box that pops up the first time you try to delete a note:

image

If you press Delete, it will show up again the next time. If you press Don't show again, we save a flag and ignore it for all future requests. There is no way to unset it without logout/login.

vitorpamplona:
Nice!

maxmoney21m:
What's going on with these tests?

vitorpamplona:
My bad on the test... 

There are a few improvements to be made here: 

1. Mention and Quote both copy to the clipboard. We need to either change the names or open a new message with a mention or the quote. 
2. If we keep the 3-dot menu, then these options should be there as well, especially the Share option
3. See if it's better with text in white, like with other buttons in the app.

Chemaclass:
I don't know why they fail now, but I know how to fix them. Preparing a PR with the fix...

vitorpamplona:
I would just keep the old language then. Right now, calling this mention and quote doesn't make much sense. 


Also keep in mind that many (majority?) phones do not show the Toast message when the key is copied to the clipboard. It shows nothing when you click on it. If the person doesn't know that this button copies the key/note id, they won't know what's happening.

maxmoney21m:
> Also keep in mind that many (majority?) phones do not show the Toast message when the key is copied to the clipboard. It shows nothing when you click on it. If the person doesn't know that this button copies the key/note id, they won't know what's happening.

It doesn't show the toast even though I have explicitly launched a Toast?

PR #230, opened 2023-03-08
Fix LN invoice in expandable text component

maxmoney21m:
The original text extraction function in the unexpanded version of the ExpandableRichTextViewer blindly pulled 350 characters out of the string. This fix does a pre-check for an invoice before extracting characters. If an index is found within the first 350 characters, we make sure to include the whole invoice.

Original problem was reproducible. Fix tested with Zeus in emulator.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/components/ExpandableRichTextViewer.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/components/ExpandableRichTextViewer.kt
@@ -26,7 +26,10 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.navigation.NavController
 import com.vitorpamplona.amethyst.R
+import com.vitorpamplona.amethyst.service.lnurl.LnInvoiceUtil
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/components/ExpandableRichTextViewer.kt
```diff
@@ -40,7 +43,16 @@ fun ExpandableRichTextViewer(
 ) {
     var showFullText by remember { mutableStateOf(false) }
 
-    val text = if (showFullText) content else content.take(350)
+    val text = if (showFullText) {
+        content
+    } else {
+        val (lnStart, lnEnd) = LnInvoiceUtil.locateInvoice(content)
+        if (lnStart < SHORT_TEXT_LENGTH && lnEnd > 0) {
+            content.take(lnEnd)
```
maxmoney21m, line 51: Ensures we do not chop LN invoices in half.

vitorpamplona:
Good quick fix, but we need to figure out how to do this Expandable view with the size of the container, not the number of characters in it.

PR #233, opened 2023-03-08
Restructuring User/Note

vitorpamplona:
I restructured some of the basic roles of Note and User, passing more responsibility to the Event hierarchy and killing some memory-intensive indexes (Sets) in User and Note. In practice, the follow/follower lists, user notes and notification lists are reassembled every time the app needs them from the Event object instead of OO-friendly per-user cache.

I think (if I am not going crazy) this is faster. It's a bit more low-level, but maybe better for the long term. 

@maxmoney21m @Chemaclass @Radiokot Can you folks test this PR, please? Just run it and see if you find any issue. If it's faster or slower than the current main.


Radiokot:
I've tested the branch. It works, I haven't noticed any issues with notes, following and followers – the content is the same, all the counters are the same 👌🏻 
However, I can't tell whether it is faster or not, as the elephant in the room is the overall UI lag (even if the debug cable is dicsonnected). This is not the issue with the branch, the app performance is like this all the time. In the following video, I compare Amethyst to Twitter:
<video src="https://user-images.githubusercontent.com/5675681/223999158-8a128bac-c23d-49cd-80f7-599bb2bc0fb7.mp4" controls="controls" style="max-width: 400px;">
</video>

vitorpamplona:
So, the lag is still there even at start up? With my user this lag shows up only after some time, usually after going to a few profiles and downloading all their followers and notes, filling up the memory. The lag happens because the app waits for garbage collection

Radiokot:
> So, is the lag is there even at start up?

Unfortunately yes, maybe it is due to differences in memory management on different devices.

maxmoney21m:
I am seeing similar to @Radiokot. I have 2 devices, Android 10 & Android 13 (Pixel 7). The Pixel 7 is pretty smooth on either version, and the old phone is pretty choppy on either version, looks very similar to the screen recording. (On start up and initial browsing.)

Did you do some profiling/benchmarks to see which methods CPU is spending the most time in, etc?

vitorpamplona:
Yep, the issue is that this version uses less memory and more CPU. Since the choppiness in my phone comes from the garbage collector, the idea was to reduce memory use but increase CPU usage recomputing things in Profile Pages and Notifications. 

It's not clear. The only benefit I can see is on long-term use. Because this uses less memory, it takes longer to reach an OutOfMemory error for heavy users.

maxmoney21m:
So the app eventually runs out of memory? What is filling it up?

vitorpamplona:
With heavy use yes, but before it happens, it keeps trying to release memory with the garbage collector. 

Right now mostly Zaps, Badges, Contact Lists (follower counts), Tons of small reaction events, heavy images/videos, spam (when you have) and DMs (if you use a lot of DMs)

maxmoney21m:
Do you have a strategy already in mind for dealing with it, like moving some things to persistent storage, or...?

vitorpamplona:
Persistent storage is in the feature list. I don't think it solves much (besides avoiding re-downloads), though. In the end, we need a good memory cache with nice pruning capabilities.

maxmoney21m:
It might allow for keeping a lot less stuff in memory generally, since it's much faster to search for and pull things out of the disk than to get them again over the wire. BTW check DMs if you want to discuss sync instead of async here.

vitorpamplona:
I expect it to either be slow if it's actually pulling from disk (we do too many queries all the time), or have a similar memory problem in the DB memory cache.

vitorpamplona:
But in the long term, we will need to work offline, so DB is a must.

maxmoney21m:
@vitorpamplona I wonder if the general choppiness is related to https://issuetracker.google.com/issues/228995735 (too many `combinedClickable` items on all of our screens)

vitorpamplona:
Updating this as we speak. Some of the issue was the incredibly slow `markAsRead` that must run on IO

maxmoney21m:
> Some of the issue was the incredibly slow markAsRead that must run on IO

That makes sense. I was thinking it might be possible to keep instances of all the preference files opened at app startup. Or have you got a better idea?

We should probably switch to DataStore and Flow instead of SharedPreferences, eventually...

vitorpamplona:
> I was thinking it might be possible to keep instances of all the preference files opened at app startup.

The issue is marking things as read (and saving to disk) as the user navigates, especially in the Chat list Screen. Delaying the save to disk can help but not sure by how much. Either way, I don't think this is the problem if the choppiness is still there. 

> We should probably switch to DataStore and Flow instead of SharedPreferences, eventually...

Agree

PR #245, opened 2023-03-10
Keep feed position while navigating different screens

maxmoney21m:
This PR does not persist feed positions across opening/closing the app, just across navigating different screens. Prevents things from jumping around so much. Currently applies only go Global, Home, and Chatrooms, but can apply to anything that uses a LazyListState. We just store a keyed map of index/offset and look them up when loading the lazy list state.

The second 2 commits replace SwipeRefresh (deprecated) with PullRefresh (replacement), just to remove some deprecation warnings and simplify the code a tiny bit. They are not related to keeping the feed position, and you can use the first 2 commits independently if you don't want this change.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
@@ -1,11 +1,17 @@
 package com.vitorpamplona.amethyst.ui.navigation
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.livedata.observeAsState
+import androidx.lifecycle.viewmodel.compose.viewModel
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
+import com.vitorpamplona.amethyst.ui.dal.GlobalFeedFilter
 import com.vitorpamplona.amethyst.ui.screen.AccountStateViewModel
+import com.vitorpamplona.amethyst.ui.screen.NostrGlobalFeedViewModel
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.AccountViewModel
+import com.vitorpamplona.amethyst.ui.screen.loggedIn.SearchScreen
 
 @Composable
 fun AppNavigation(
@@ -14,9 +20,28 @@ fun AppNavigation(
     accountStateViewModel: AccountStateViewModel,
     nextPage: String? = null
 ) {
+    val accountState by accountViewModel.accountLiveData.observeAsState()
+    val account = accountState?.account ?: return
+
+    GlobalFeedFilter.account = account
+    val globalFeedViewModel: NostrGlobalFeedViewModel = viewModel()
+
     NavHost(navController, startDestination = Route.Home.route) {
+        Route.Search.let { route ->
+            composable(route.route, route.arguments, content = {
+                SearchScreen(
+                    accountViewModel = accountViewModel,
+                    feedViewModel = globalFeedViewModel,
+                    navController = navController,
+                    scrollToTop = it.arguments?.getBoolean("scrollToTop") ?: false
+                )
+            })
+        }
+
         Routes.forEach {
-            composable(it.route, it.arguments, content = it.buildScreen(accountViewModel, accountStateViewModel, navController))
+            it.buildScreen?.let { fn ->
+                composable(it.route, it.arguments, content = fn(accountViewModel, accountStateViewModel, navController))
+            }
         }
     }
 
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
@@ -7,9 +7,15 @@ import androidx.lifecycle.viewmodel.compose.viewModel
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
+import com.google.accompanist.pager.ExperimentalPagerApi
+import com.google.accompanist.pager.rememberPagerState
 import com.vitorpamplona.amethyst.ui.dal.GlobalFeedFilter
+import com.vitorpamplona.amethyst.ui.dal.HomeConversationsFeedFilter
+import com.vitorpamplona.amethyst.ui.dal.HomeNewThreadFeedFilter
 import com.vitorpamplona.amethyst.ui.screen.AccountStateViewModel
 import com.vitorpamplona.amethyst.ui.screen.NostrGlobalFeedViewModel
+import com.vitorpamplona.amethyst.ui.screen.NostrHomeFeedViewModel
+import com.vitorpamplona.amethyst.ui.screen.NostrHomeRepliesFeedViewModel
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.AccountViewModel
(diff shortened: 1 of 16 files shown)


vitorpamplona:
I will test this soon, but does it keep the behavior of clicking in the currently-selected bottom icon to refresh and navigate all the way up to the most recent item?

maxmoney21m:
> I will test this soon, but does it keep the behavior of clicking in the currently-selected bottom icon to refresh and navigate all the way up to the most recent item?

Good catch - no. I will fix it and update the PR.

maxmoney21m:
Alright @vitorpamplona go ahead and test it. There is a still one minor bug around tapping the Home icon if you are in the 2nd tab, it takes you back to the top of the 1st tab. This is the current behavior, and I haven't fixed that part in this PR. I believe the pager state also needs to be hoisted out, but I don't have time to do it right now because it will require a little bit more work on the router.

Otherwise, Global & Home work as expected, the scroll position is saved while navigating around, and you can refresh the feed and jump to the top by tapping the icon again.

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/Routes.kt
```diff
@@ -35,17 +35,32 @@ sealed class Route(
...
-        "Search",
+        "Search?forceRefresh={forceRefresh}",
         R.drawable.ic_globe,
-        buildScreen = { acc, accSt, nav -> { _ -> SearchScreen(acc, nav) } }
+        arguments = listOf(navArgument("forceRefresh") { type = NavType.BoolType; defaultValue = false }),
+        buildScreen = { acc, accSt, nav ->
+            { backStackEntry ->
+                SearchScreen(acc, nav, backStackEntry.arguments?.getBoolean("forceRefresh", false))
+            }
+        }
```
maxmoney21m, line 61: This is where the magic happens now. For any screens that need this functionality, just add the `forceRefresh` argument and pass it to the `___Screen` Composable as seen here.

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/Routes.kt
```diff
@@ -35,17 +35,32 @@ sealed class Route(
     val arguments: List<NamedNavArgument> = emptyList(),
     val buildScreen: (AccountViewModel, AccountStateViewModel, NavController) -> @Composable (NavBackStackEntry) -> Unit
 ) {
+    val base: String
+        get() = route.substringBefore("?")
```
maxmoney21m, line 39: It turns out we sometimes use the route without any optional params added (e.g. for comparing whether or not we are on the same route), so we can use this `base` delegated property whenever we need to check that.

vitorpamplona:
Great work here! I think the only issue I found is that when you click in the bottom button to refresh, the pulldown refreshing animation shows up as well. That feels weird.

maxmoney21m:
> Great work here!

Thanks

> I think the only issue I found is that when you click in the bottom button to refresh, the pulldown refreshing animation shows up as well. That feels weird.

I left it that way because I thought it would be good to have an indicator that some loading was happening (even if it already happened, it's good visual cue to the user that they are jumping to the top now). If you want to disable it, just wrap [FeedView.kt#L91](https://github.com/vitorpamplona/amethyst/blob/90147ce557a8948330ea9e33ace33f9bf10aaaf2/app/src/main/java/com/vitorpamplona/amethyst/ui/screen/FeedView.kt#L91) like this:

```kotlin
if (!forceRefresh) {
    PullRefreshIndicator(refreshing, pullRefreshState, Modifier.align(Alignment.TopCenter))
}
```

Not tested, but I think that should do it.

maxmoney21m:
@vitorpamplona I think I found out what the problem is here: it looks like we need to hoist all of the `___FeedViewModel` instances into the `MainScreen`, otherwise any navigation action causes the `___Screen` composable to be recreated, and that ends the `___FeedViewModel` lifecycle since it's declared at the top of the `___Screen`. A new one gets created, and the screen blinks to black for a moment while that happens.

By hoisting and passing down, the blink goes away and we can finally do the scroll animation when you tap the button to force a refresh. Unfortunately there is a [separate issue](https://issuetracker.google.com/issues/204328131) that causes the scroll animation to be very laggy which we can't do anything about. But using the non-scrolling version is actually quite nice, the screen does not blink, and it feels very smooth.

I am just tidying up the code a bit and will push this change later, so don't merge yet.

vitorpamplona:
Nice! Yes, that has been the pain of my existence for quite some time. Looking forward to seeing what you found as a solution

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/Routes.kt
```diff
@@ -16,52 +14,34 @@ import com.vitorpamplona.amethyst.model.Account
...
-import com.vitorpamplona.amethyst.ui.screen.loggedIn.ProfileScreen
-import com.vitorpamplona.amethyst.ui.screen.loggedIn.SearchScreen
-import com.vitorpamplona.amethyst.ui.screen.loggedIn.ThreadScreen
 
 sealed class Route(
     val route: String,
     val icon: Int,
     val hasNewItems: (Account, NotificationCache, Context) -> Boolean = { _, _, _ -> false },
-    val arguments: List<NamedNavArgument> = emptyList(),
-    val buildScreen: (AccountViewModel, AccountStateViewModel, NavController) -> @Composable (NavBackStackEntry) -> Unit
```
maxmoney21m, line 36: Removed this lambda parameter for two reasons:

1. The various `___Screen` components have different needs, and this was janky.
2. We can better manage state hoisting now in the `AppNavigation` component

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
```diff
@@ -1,22 +1,120 @@
...
+        Route.Search.let { route ->
+            composable(route.route, route.arguments, content = {
+                SearchScreen(
+                    accountViewModel = accountViewModel,
+                    feedViewModel = globalFeedViewModel,
+                    navController = navController,
+                    scrollToTop = it.arguments?.getBoolean("scrollToTop") ?: false
+                )
+            })
+        }
```
maxmoney21m, line 60: This block replaces the `buildScreen` function. We need one per Route, but it's much simpler to build the screens directly.

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
```diff
@@ -1,22 +1,120 @@
...
+    val account = accountState?.account ?: return
+
+    GlobalFeedFilter.account = account
+    HomeNewThreadFeedFilter.account = account
+    HomeConversationsFeedFilter.account = account
+
+    val globalFeedViewModel: NostrGlobalFeedViewModel = viewModel()
+    val homeFeedViewModel: NostrHomeFeedViewModel = viewModel()
+    val homeRepliesFeedViewModel: NostrHomeRepliesFeedViewModel = viewModel()
+    val homePagerState = rememberPagerState()
```
maxmoney21m, line 48: We can probably hoist more state up here, at least for the other 2 bottom tabs (notifications / channels)

maxmoney21m:
Alright @vitorpamplona , the new technique is in and it's pretty solid in my testing. It also allowed me to fix the "Home" button to not always jump you back to the "New Threads" tab, so now clicking Home will keep you in your current tab. And the scroll position of the feed in both Home tabs as well as the Global screen will all be saved as you navigate the app.

Issue #54, opened 2023-01-29
proprietary dependencies added with v0.11.4


IzzySoft:
Looks like today's release added a bunch of proprietary dependencies:

```
Offending libs:
---------------
* Firebase Data Transport (/com/google/android/datatransport): NonFreeNet
* Google Mobile Services (/com/google/android/gms): NonFreeDep
* Firebase (/com/google/firebase): NonFreeNet,NonFreeDep
* ML Kit (/com/google/mlkit): NonFreeDep,Tracking

4 offenders.
```

My guess is you've added MLKit, and the other 3 were dragged in by that. Now I don't know what to do:

* remove today's version and disable updates (in the hope you'll revert the above)
* adding 3 anti-features (NonFreeNet,NonFreeDep,Tracking) which will make your app looking weird (privacy and tracking?) and scare some new users away

So what do you think, @vitorpamplona?

vitorpamplona:
Those are for the QR code scanning. I tried other ones but it wasn't really working well. 

We need to search for a good way to scan since Google's default camera doesn't send nostr:// uris to Amethyst. :(

IzzySoft:
Maybe look what other apps use? [Conversations](https://codeberg.org/iNPUTmice/Conversations) has QR Code scanning as well and is fully FOSS (they [use zxing](https://codeberg.org/iNPUTmice/Conversations/src/branch/master/build.gradle#L57) it seems), for example. I could check what other QR Code libraries my library definitions know about.

PS: I see the APK size also more than doubled, having reached 25 MB now. With a limit of 30 MB per app, my repo will only be able to keep 1 APK then.

IzzySoft:
Apart from ZXing I could only find [zBarLibary](https://github.com/bertsir/zBarLibary) (Java) in my list.

IzzySoft:
> since Google's default camera doesn't send nostr:// uris to Amethyst.

That camera app isn't even installed on all devices. Maybe try [SecScanQR](https://github.com/Fr4gorSoftware/SecScanQR) and see what they are using. I'm using that scanner for many years now and am quite happy with it. It's of course available at F-Droid, so you could even recommend it to your users. They btw use Zxing (`com.google.zxing:core:3.4.0`).

IzzySoft:
[...]

PR #249, opened 2023-03-10
Replace mlkit with zxing QR scanner

maxmoney21m:
I wanted to give this as an option for #54. It works fine for me on the few devices I've tested, and the scanner documentation says it supports SDK 24+ by default, so there shouldn't be any problems. True, the interface is not quite as polished as the mlkit version, but for scanning a QR it's probably not a big deal.

This does not completely solve the file size issue. A large part of that is in RoboHash images, which would be fixed with my [svg branch](https://github.com/maxmoney21m/amethyst/tree/215-robohash-identicons) (needs a little work for PR). The other part of it would be doing a split APK for each platform. With this PR, replacing bitmap with svg robohash, and doing a split APK, the file size should go down to under 30 MB.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
@@ -6,7 +6,6 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
@@ -34,13 +33,13 @@ fun QrCodeDrawer(contents: String, modifier: Modifier = Modifier) {
         createQrCode(contents = contents)
     }
 
-    val foregroundColor = MaterialTheme.colors.onSurface
+    val foregroundColor = Color.Black
 
     Box(
         modifier = modifier
             .defaultMinSize(48.dp, 48.dp)
             .aspectRatio(1f)
-            .background(MaterialTheme.colors.background)
+            .background(Color.White)
     ) {
         Canvas(modifier = Modifier.fillMaxSize()) {
             // Calculate the height and width of each column/row
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
```diff
@@ -34,13 +33,13 @@ fun QrCodeDrawer(contents: String, modifier: Modifier = Modifier) {
...
 
-    val foregroundColor = MaterialTheme.colors.onSurface
+    val foregroundColor = Color.Black
 
     Box(
         modifier = modifier
             .defaultMinSize(48.dp, 48.dp)
             .aspectRatio(1f)
-            .background(MaterialTheme.colors.background)
+            .background(Color.White)
```
maxmoney21m, line 42: QR code spec says that the codes must be dark on a light background. With the previous version, our QR codes were improperly formatted in dark mode.

Also note that zxing does not support QR codes that aren't up to spec.

app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeScanner.kt
```diff
@@ -1,156 +1,57 @@
...
+        setDesiredBarcodeFormats(ScanOptions.QR_CODE)
+        setPrompt(stringResource(id = R.string.point_to_the_qr_code))
+        setBeepEnabled(false)
+        setOrientationLocked(false)
+    }
+
+    DisposableEffect(lifecycleOwner) {
+        qrLauncher.launch(scanOptions)
+        onDispose { }
+    }
```
maxmoney21m, line 56: This is the entirety of the new code needed to use zxing, and replaces all of the previous QR scanner code.

app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeDrawer.kt
vitorpamplona: Are you sure this isn't an old spec? The last time I read the text, it could be high-contrast color.

maxmoney21m: > old spec

That could be.

vitorpamplona: That can be a problem. Damus uses their blue-purple GRADIENT as the QR code background :(

vitorpamplona: Do you want to try to separate this work into a full-on flavor of the app? Even if it doesn't work that well, F-Droid folks would love it. 

The idea is: the current package name has Google dependencies and is shipped to the PlayStore. The new flavor would remove Google's deps and add this work and ship to F-Droid

vitorpamplona: Like here: https://stackoverflow.com/questions/18196974/how-to-define-different-dependencies-for-different-product-flavors

maxmoney21m:
@vitorpamplona I missed [this issue](https://github.com/journeyapps/zxing-android-embedded/issues/693) the first time round, but it seems everything is working now. I can scan dark/light/color QR codes with no issue from my old phone using this library.

Doing flavors is fine, but IMO it should be sort of the last choice, and this library seems to be working fine now with the correct settings.

app/src/main/java/com/vitorpamplona/amethyst/ui/qrcode/QrCodeScanner.kt
```diff
@@ -48,6 +49,7 @@ fun QrCodeScanner(onScan: (String?) -> Unit) {
         setPrompt(stringResource(id = R.string.point_to_the_qr_code))
         setBeepEnabled(false)
         setOrientationLocked(false)
+        addExtra(Intents.Scan.SCAN_TYPE, Intents.Scan.MIXED_SCAN)
```
maxmoney21m, line 52: That was it. Works fine now for all kinds of QR codes that I could test (inverted white-black and colors)

vitorpamplona:
I will test this tomorrow with my iPhones at the office

vitorpamplona:
This now works great with Damus!

PR #258, opened 2023-03-11
Generate robohash using SVGs

### LINKED ISSUE #215: [BUG] Static robohash as the default avatar
I knocked together a quick demo of this just to see if it would work, and turns out to be very easy. The [robohash](https://robohash.org/) icons are available as svg under MIT license [here](https://github.com/nimiq/robohash).

Some quick and dirty splicing them together directly in the code gives me a nice looking avatar:

image

@vitorpamplona if you would accept a PR for this I will polish up the code and submit.

The end result is just a function:

```
@Composable
fun Robohash(npub: String): ByteBuffer { ... }
```

Where the `ByteBuffer` can be passed directly to `Coil` data loader instead of a URL:

```
AsyncImage(
  model = RoboHash(user.npub),
  ...
)
```

I didn't benchmark anything but I think it should be pretty performant, it's just some simple string formatting and manipulation.

maxmoney21m:
@vitorpamplona mentioned [here](https://github.com/vitorpamplona/amethyst/issues/215#issuecomment-1457937943) that creating too many bitmaps was slow. This PR would fix #215, it replaces all of the bitmap robohashing with on demand SVG generation with simple string manipulation. All avatars (including chat rooms, relay icons, etc.) have been replaced with robohash SVGs.

I don't know if it's "fast enough" but it feels OK on my test devices & emulator. If you want to test loading tons of robohash images, just hardcode `loading = true` and every avatar will be a robohash, see this snippet:

```kotlin
// RobohashFallbackAsyncImage.kt

@Composable
fun RobohashFallbackAsyncImage(
    robot: String = "aaaa",
    // ...
) {
    /* var loading by remember { mutableStateOf(false) } */
    val loading = true // use this instead
    var error by remember { mutableStateOf(false) }

    Box {
        AsyncImage(
            // comment out
            onLoading = { /* loading = true */ },
            onSuccess = { /* loading = false; */ error = false },
            // ...
        )
```

Robot SVGs from https://github.com/nimiq/robohash


app/src/main/java/com/vitorpamplona/amethyst/ui/components/Robohash.kt
```diff
@@ -0,0 +1,299 @@
...
+    val faceIndex = byteMod(hash[11], 10)
+    val eyesIndex = byteMod(hash[12], 10)
+    val mouthIndex = byteMod(hash[13], 10)
+    val accIndex = byteMod(hash[14], 10)
+    val background = backgrounds[bgIndex]
+    val body = bodies[bodyIndex]
+    val face = faces[faceIndex]
+    val eye = eyes[eyesIndex]
+    val mouth = mouths[mouthIndex]
+    val accessory = accessories[accIndex]
```
maxmoney21m, line 44: Hash a message and pick some values from the bytes (not sure if there is a preferred way to do this, I just did it arbitrarily...)

app/src/main/java/com/vitorpamplona/amethyst/ui/components/Robohash.kt
```diff
@@ -0,0 +1,299 @@
...
+            <defs>
+                <style>
+                    .cls-fill-1{fill:${toHex(fgColor)};}.cls-fill-2{fill:${toHex(fgColor)};}.bg-1{fill:${toHex(bgColor1)};}.bg-2{fill:${toHex(bgColor2)}}
+                    ${body.style}${face.style}${eye.style}${mouth.style}${accessory.style}
+                </style>
+            </defs>
+            <title>RoboHash $hashHex</title>
+            ${background}${body.paths}${face.paths}${eye.paths}${mouth.paths}${accessory.paths}
+        </svg>
+    """.trimIndent()
```
maxmoney21m, line 57: Generate the SVG string by hand. All the component part SVG strings are embedded directly in this file, no additional assets required.

app/src/main/java/com/vitorpamplona/amethyst/ui/components/Robohash.kt
```diff
@@ -0,0 +1,299 @@
...
+        return ImageRequest
+            .Builder(context)
+            .data(
+                ByteBuffer.wrap(
+                    svgString(message).toByteArray()
+                )
+            )
+            .build()
+    }
+}
```
maxmoney21m, line 71: This `ImageRequest`-wrapped-SVG is passed to Coil as the data model.

app/src/main/java/com/vitorpamplona/amethyst/ui/components/RobohashAsyncImage.kt
```diff
@@ -0,0 +1,128 @@
...
+    var error by remember { mutableStateOf(false) }
+
+    Box {
+        AsyncImage(
+            model = model,
+            contentDescription = contentDescription,
+            modifier = modifier,
+            onLoading = { loading = true },
+            onSuccess = { loading = false; error = false },
+            onError = { error = true },
```
maxmoney21m, line 69: Use simple loading/error state flags to show or hide the robohash SVG on top of the base image.

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -135,12 +133,10 @@ fun ProfileContent(baseAccountUser: User, modifier: Modifier = Modifier, scaffol
...
 
         Column(modifier = modifier) {
-            AsyncImageProxy(
+            RobohashAsyncImageProxy(
+                robot = accountUser.pubkeyHex,
                 model = ResizeImage(accountUser.profilePicture(), 100.dp),
                 contentDescription = stringResource(id = R.string.profile_image),
-                placeholder = BitmapPainter(RoboHashCache.get(ctx, accountUser.pubkeyHex)),
-                fallback = BitmapPainter(RoboHashCache.get(ctx, accountUser.pubkeyHex)),
-                error = BitmapPainter(RoboHashCache.get(ctx, accountUser.pubkeyHex)),
```
maxmoney21m, line 143: I removed placeholder/fallback/error parameters on all calls to the image composables. I'm not sure if they should still be provided, maybe with some static image like the Amethyst logo, just in case.

vitorpamplona:
This is very cool. 

Two things: 
- Is it possible to remove the background? With a colorful background, it's hard to know when its a real user picture and when it's updating... Maybe for background as black at all times? 
- I believe the SVG blinks in black when switching between the user and the SVG. Maybe a cross fade?

maxmoney21m:
> Is it possible to remove the background? With a colorful background, it's hard to know when its a real user picture and when it's updating... Maybe for background as black at all times?

Makes sense, just means that there will be a lot less options. I wonder if there is something we could do to make it obvious they are generated and still use them. The color (esp. the two different colors between the background & the robot) make them useful as identifiers. I'm thinking if we're going to have the robohash bots, they might as well be identifiable at a glance so they can serve their purpose properly.

I have 2 alternative suggestions before we remove them altogether:

1. Add a purple Amethyst (logo-shaped) diamond behind the robot. (No ostrich cutout, just the diamond)
2. Maybe just use a flat color instead of the various patterns.

> I believe the SVG blinks in black when switching between the user and the SVG. Maybe a cross fade?

Good idea, I'll give it a try.

maxmoney21m:
image

image

image

They don't look that good, but probably a designer could come up with something that would make them identifiably "Amethyst generated robohash bots", so we can keep the color backgrounds.

maxmoney21m:
Ok, these actually don't look too bad. Remove the background, and just leave the diamond, but with a random color:

image

image

image

image

image

I don't think these will be easily confused with users who are actually using robohash or similar avatars, but they are still more easily recognizable than everything just on a black background. WDYT @vitorpamplona ?

vitorpamplona:
These are good. My only thing is to make sure there is a distinction between loaded and unloaded information.

maxmoney21m:
> These are good. My only thing is to make sure there is a very clear distinction between loaded and unloaded information.

How about this:

1. If there's no image URL or we get an onError, we use the robot directly 
2. If there is an image URL, then initially and onLoading we just use the Amethyst logo.

That way if a robot shows up it means we definitely don't have an image, so it won't flicker from robot to PFP.

vitorpamplona:
> If there is an image URL, then initially and onLoading we just use the Amethyst logo.

We should be able to show the robot also when loading, no? So the transition becomes robot -> PFP instead of robot - black - pfp

maxmoney21m:
> We should be able to show the robot also when loading, no? So the transition becomes robot -> PFP instead of robot - black - pfp

Oh I see, sorry I misunderstood, I thought that this:

> a very clear distinction between loaded and unloaded information

meant you wanted to differentiate between PFPs that have not yet been loaded, and knowing that we won't get a PFP.

Ok, then let me see if I can fix the blink.

app/src/main/java/com/vitorpamplona/amethyst/ui/components/RobohashAsyncImage.kt
```diff
@@ -56,37 +52,22 @@ fun RobohashFallbackAsyncImage(
...
-            )
-        }
-    }
+    AsyncImage(
+        model = model,
+        contentDescription = contentDescription,
+        modifier = modifier,
+        placeholder = painter,
+        fallback = painter,
+        error = painter,
```
maxmoney21m: Apparently we can just do this to use a fallback and remove the makeshift Box. It seems to load faster for me and no noticeable blink, can you check it out again @vitorpamplona?

vitorpamplona:
Hey @maxmoney21m, can you merge from main again? Looks like the AccountSwitchBottomSheet needs to use your new RoboHash here

PR #261, opened 2023-03-11
Refactor LocalCache

Chemaclass:
## 📖  Changes

- Remove the unused value `checkHex`
- Applying extract to method refactoring
- Invert conditional in `fun consume(event: ChannelCreateEvent)` to get rid of one indentation level

Screenshot 2023-03-11 at 20 53 32

Screenshot 2023-03-11 at 20 54 29

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -58,7 +58,7 @@ object LocalCache {
 
     fun checkGetOrCreateUser(key: String): User? {
         return try {
-            val checkHex = Hex.decode(key).toNpub() // Checks if this is a valid Hex
+            checkIfValidHex(key)
             getOrCreateUser(key)
         } catch (e: IllegalArgumentException) {
             Log.e("LocalCache", "Invalid Key to create user: $key", e)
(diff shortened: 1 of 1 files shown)


JesusValera (approved):

app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
```diff
@@ -99,14 +99,18 @@ object LocalCache {
...
+            checkIfValidHex(key)
             getOrCreateChannel(key)
         } catch (e: IllegalArgumentException) {
             Log.e("LocalCache", "Invalid Key to create channel: $key", e)
             null
         }
     }
 
+    private fun checkIfValidHex(key: String) {
+        Hex.decode(key).toNpub()
```
maxmoney21m, line 111: I'm not at PC right now, but this method throws? I would put the try catch here and return a Boolean from this function instead.

maxmoney21m (changes requested):

app/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
vitorpamplona: Agree, it can be a `isValidHexNpub` (it throws in both parsing and creating if the key is not valid) with a try catch inside, returning a simple boolean to each check function. 

We will do this in the new refactoring :)

Chemaclass: Thanks for the feedback! I will do that later :)

PR #270, opened 2023-03-14
Update NL strings.xml

Bardesss:
Added missing strings an fixed some mistakes. We we're using notes en notities (Dutch translation), changed it all to notes for now.


matata2140:
thought about that too so +1 on that

PR #275, opened 2023-03-14
Add separate screen for report & block

### LINKED ISSUE #216: [FEATURE] Merge all “Report” options
The post drop-down menu is pretty long already. It might make sense to join all of the “Report ___” options together in a single “Report” button that opens a dialog to pick the report type.

maxmoney21m:
The quick action menu & dropdown menu both have a single button for block/report now:

image

image

Clicking the button opens this page:

image

Tap in the reason field to select a reason from the menu:

image

After selection, the Post Report button is enabled. Add additional context in the text field, attached as `content` field in the report event.

image

Also re-reverted ChatroomMessageCompose back to using the quick action menu since this is now in place.

Resolves #216

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ render.experimental.xml
 .idea/**/caches/
 .idea/**/libraries/
 .idea/**/shelf/
+.idea/**/codeStyles
 .idea/**/workspace.xml
 .idea/**/tasks.xml
 .idea/**/.name
--- a/app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
@@ -47,7 +47,8 @@ private object PrefKeys {
     const val TRANSLATE_TO = "translateTo"
     const val ZAP_AMOUNTS = "zapAmounts"
     const val LATEST_CONTACT_LIST = "latestContactList"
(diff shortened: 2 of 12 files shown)


vitorpamplona:
So, block and report are two things for PlayStore reviewers. We might need to name the button Block and Report or just have two buttons that go to the same place.

maxmoney21m:
> So, block and report are two things for PlayStore reviewers. We might need to name the button Block and Report or just have two buttons that go to the same place.

If they are using the 3-dots menu, it's called “Block / Report”, and if they click Block on the quick action items, the screen title is “Block and Report”. I wonder if that is sufficient?

vitorpamplona:
I don't think so, we have been rejected because the menu only had one name ... :(  That's why the "Block / Report" is spelled out like that. My concern is that they won't even click in the button to know that there is a second screen (has happened many times before)

maxmoney21m:
> I don't think so, we have been rejected because the menu only had one name ... :( That's why the "Block / Report" is spelled out like that. My concern is that they won't even click in the button to know that there is a second screen (has happened many times before)

What do you think about having the quick action block button actually just do the block directly, while the report button goes to this screen? I'm just thinking users would be confused if they both linked to the same screen.

maxmoney21m:
Added block button and "don't show again" dialog.

image

image

PR #276, opened 2023-03-14
Make release channel flavors, remove mlkit from fdroid channel

maxmoney21m:
This commit adds a flavor dimension for release channels, and defines two flavors, play store and fdroid. The mlkit dependency is only added to the play flavor, and the `TranslateableRichTextViewer` is defined as equal to the `ExpandableRichTextViewer` for the fdroid flavor. It also splits the builds by abi for x86 and x86_64.

No other code changes were made besides that definition for fdroid. Language service related tests were moved to the appropriate flavor directory as well, so the tests should be passing.

Both builds tested locally and perform as expected - no translation UI on the fdroid version. The x86_64 fdroid build is now coming in at 28 MB.

Let's see if this solves #54.


vitorpamplona:
Super cool! 

1. Do we need different package names for them? Will it create an issue of users going back and forth between a PlayStore/APK installation and an f-droid installation? 

2. We need to update the github actions to test both flavors, generate assets and upload to Github on new releases.

maxmoney21m:
I believe different package names would allow them to be installed in parallel, and we can keep the same name to only allow installing one.

The testing happens automatically (Gradle understands the flavors; check the logs). You just need to fix the artifact names for the upload action.

IzzySoft:
A practice I've seen is to append `.foss` or `.fdroid` to the "normal package name".

PR #278, opened 2023-03-14
Update strings.xml

matata2140:
A small addition I don't know how to properly correct in the file wrt the translation of the 'auto-translation pop-up'. 
Referring specifically to line 162 and 163, "Laten zien in" and "eerste".

In English: "Show in [language] first" = grammatically ok
In Dutch: "Laten zien in [language] eerste" = grammatically wrong  **--> should be "Eerst laten zien in [language]"**

### DIFF
--- a/app/src/main/res/values-nl/strings.xml
+++ b/app/src/main/res/values-nl/strings.xml
@@ -161,7 +161,7 @@
     <string name="to">naar</string>
     <string name="show_in">Laten zien in</string>
     <string name="first">eerste</string>
-    <string name="always_translate_to">Altijd vertalen naar</string>
+    <string name="always_translate_to">Altijd vertalen naar </string>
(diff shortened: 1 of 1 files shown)


maxmoney21m:
@matata2140 good catch. These need to be parameterized like:

```
// en
<string name="show_in_lang_first">Show in %1$s first</string>

// nl
<string name="show_in_lang_first">Eerst laten zien in %1$s</string>
```

but it will require some code changes so you can't do it just yet. I'll fix it later. @vitorpamplona do you know of any other strings that are using these "split strings" to do a paramterization workaround? I'll fix them all at the same time if there are others.

matata2140:
there's also a tiny added change (a space) in the file that should be merged one way or the other

maxmoney21m:
@matata2140 this should be fixed in the current release. I added the space with a few other edits.

PR #279, opened 2023-03-15
Add .profm sorting to build.gradle

maxmoney21m:
For #54, @IzzySoft is this what you mean? I'm not that familiar with the gradle build process, or with creating reproducible builds. @vitorpamplona I built it and result looks normal to me, but please don't merge until IzzySoft can confirm.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -181,3 +181,35 @@ dependencies {
 ktlint {
     disabledRules.set(["no-wildcard-imports"])
 }
+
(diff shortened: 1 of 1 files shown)


IzzySoft:
@maxmoney21m thanks – I'd say that is it (note that code is not from me, nor is that gist – that's all the work of obfusk :wink:

Once a new release has this, I'll just update the MR once more, then compare again. With some luck we're ready then :crossed_fingers:

PR #280, opened 2023-03-15
Remove ignored file

maxmoney21m:
This file was removed in #265 but it's back... do you need to remove it locally or something @vitorpamplona?

```
git rm --cached .idea/misc.xml
```


vitorpamplona:
hum... my bad... probably a merge conflict with that older pr I had going for a while

PR #283, opened 2023-03-16
Polls: UI RFC

toadlyBroodle:
Here is the basic UI screen for creating a new poll event. 
Thoughts on how it looks so far and tips on how to proceed?


vitorpamplona:
Hi @toadlyBroodle 

This looks ok. In the end, I think we will need to move the new Pool button to be inside the new Post Dialog to allow people to create new pools while writing a post about it. But this should work for tests. We can work in aligning the interface later. 

So, from what I see here, we are missing the event generation itself, the parsing back, and UI views to vote and show pools. Do you need any help with those?

toadlyBroodle:
Sounds good, the fab polls button and icon are just there for testing for sure. I'm still figuring out how the codebase works, so will take time, but I will be taking a stab at creating/handling new poll event type and rest of UI and logic. I'll see how far I can get before running into problems.

vitorpamplona:
This is getting really interesting... 

I would do Zap buttons beside each option. Anything that removes friction from participating.

toadlyBroodle:
@vitorpamplona I think the basic functionality is now ready for your code review and testing now! Looking forward to your feedback. Would you like me to merge with main branch first?

vitorpamplona:
This looks great! Any chance you can update it with the latest from main?

toadlyBroodle:
@vitorpamplona Thanks. Merged into main and will submit new PR.

PR #288, opened 2023-03-17
Fix for hashtag icon resolutions

believethehype:
+ 2 additional icons for #zap (and popular variations) and #amethyst


vitorpamplona:
Wrong logos, though :) 

there is an Amethyst logo in R.drawable.amethyst and a lightning logo at R.drawable.lightning

vitorpamplona:
We should probably fix the btc logo which is off-centered, to the left.

believethehype:
Updated Amethyst and Lightning icons to the existing ones.

PR #310, opened 2023-03-23
adding in support for clicking on lightning withdrawal links intenting out to a wallet 

lgleasain:
and having it go to a wallet. Right now this shows the link as clickable text but for a follow on I would like to actually get the withdrawal json payload and show details about it. I did this as part of a POC to be able to allow other mechanisms of receiving money via Nosr (for example through Azteco) at the nostrica hackathon that I thought would be useful.


vitorpamplona:
Super nice! Thank you!

PR #313, opened 2023-03-24
Create UserInterface

Chemaclass:
## 🤔 Context

This refactoring is motivated by this open-issue that I want to do: https://github.com/vitorpamplona/amethyst/issues/159
However, it is really hard to write tests when the dependencies are coupled to each other. So, before starting doing that task, I wanted to focus on refactoring the current status inverting the dependencies introducing interfaces for the collaborators.

> Extra thoughts: I know, ideally interfaces should be small (Interface Segregation Principle), but for this particular refactor phase, it's way easier to keep it simple and "extract" the whole User as an interface. We could work into segregating the interface if we consider. But I wouldn’t bother to do that know. There are other parts that are more important at this moment.

## 🔖  Changes

Introduced `UserInterface` Tt decouple the `"concrete" User` behaviour from other parts. 
This will definitely help in order to be able to write tests whenever the User collaborator is involved.

> **Note:** Similar as I already did here introducing the `EventInterface` https://github.com/vitorpamplona/amethyst/pull/205

## 🧪 Tech Hint 

After merging this PR, current PR's will luckily need to rebase/merge main in their others to solve the conflicts. They should be pretty straightforward: instead of using the `User` public properties, use the public methods from the `UserInterface`. Or, worst scenario, you can always cast your `UserInterface` as a concrete `User`, and then you can use  the `User` properties.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/model/User.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/model/User.kt
@@ -2,21 +2,11 @@ package com.vitorpamplona.amethyst.model
 
 import androidx.lifecycle.LiveData
 import com.vitorpamplona.amethyst.service.NostrSingleUserDataSource
-import com.vitorpamplona.amethyst.service.model.BookmarkListEvent
-import com.vitorpamplona.amethyst.service.model.ContactListEvent
-import com.vitorpamplona.amethyst.service.model.LnZapEvent
-import com.vitorpamplona.amethyst.service.model.MetadataEvent
-import com.vitorpamplona.amethyst.service.model.ReportEvent
+import com.vitorpamplona.amethyst.service.model.*
 import com.vitorpamplona.amethyst.service.relays.Relay
 import com.vitorpamplona.amethyst.ui.note.toShortenHex
 import fr.acinq.secp256k1.Hex
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.NonCancellable
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.*
 import nostr.postr.Bech32
 import nostr.postr.toNpub
 import java.math.BigDecimal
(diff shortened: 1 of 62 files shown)


vitorpamplona:
hum.. Can't we use abstract properties in the Interface instead of recreating methods for every property?

Chemaclass:
Oh, I didn't know we can have (abstract) properties in the interface. I am also learning kotlin on the fly, so let me try that 👍🏼

vitorpamplona:
You can have them directly in the interface, which I think creates the same effect you want, just without having to code all these methods

https://kotlinlang.org/docs/interfaces.html#properties-in-interfaces

Chemaclass:
Thanks for the hint! Jap, I found the way, and this is gonna be much better 🧠

app/src/main/java/com/vitorpamplona/amethyst/model/User.kt
```diff
@@ -25,66 +15,56 @@ import java.util.regex.Pattern
...
-    var notes = setOf<Note>()
-        private set
-
-    var reports = mapOf<User, Set<Note>>()
-        private set
-
-    var latestReportTime: Long = 0
-
-    var zaps = mapOf<Note, Note?>()
-        private set
```
Chemaclass, line 43: The only counter-side from moving the properties to the interface is that we cannot specify the visibility of the setters. Eg: `Private setters are not allowed for abstract properties`.
Screenshot 2023-03-25 at 17 56 20

What do you think, @vitorpamplona? Do we want to have methods for those vars? Or it is ok to keep it like this?

vitorpamplona: Not great since there is a reason to force these fields to be private (an update must trigger a livedata refresh) 

Maybe we take a step back and only do the interface for methods that are going to be tested? When can then have Interfaces for parts of the User class that matter for each test case (and hopefully parts of the app as well)

Chemaclass: What about creating an interface method for those properties as I had before? This way we will have the benefits from both parts.

vitorpamplona: It's hard to judge it without actually using the interface for something useful. The PR feels like we are doing this just for the sake of code styling than for any useful reason. Because of that, the new methods and classes feel like bloat. 

I think it would make more sense to start designing interfaces when test cases need it. If we are testing Zaps, we might need a UserZappingInterface with just the methods to collect, update and show zaps.

Chemaclass: Yes, you're right, @vitorpamplona. Ok, I will re-think the approach as you are suggesting 🤔

Chemaclass:
Close because the following argument (https://github.com/vitorpamplona/amethyst/pull/313#discussion_r1148568289)
> I think it would make more sense to start designing interfaces when test cases need it. If we are testing Zaps, we might need a UserZappingInterface with just the methods to collect, update and show zaps.

PR #316, opened 2023-03-25
Updated Chinese (Traditional) Translation

miseelu:
Updated Chinese (Traditional) Translation
CJK characters are shorter, so align the translation menu to the right to make it easier to be clicked.


vitorpamplona:
> CJK characters are shorter, so align the translation menu to the right to make it easier to be clicked.

But here you are aligning to the right in English as well, right? I am not sure if it makes sense.

PR #317, opened 2023-03-26
Group zaps by author

Chemaclass:
## 🚀 Goal

Issue: https://github.com/vitorpamplona/amethyst/issues/159 Combine the zaps from the same user into one list entry

## 📚 Description

First group all zaps by author, and then aggregate the zaps amount for each author, cloning the original notes to avoid altering the original data.

### BEFORE

Screenshot 2023-03-26 at 23 10 19

Or even this annoying "multiple 1 sats"...
Screenshot 2023-03-26 at 23 10 40

### AFTER

Now it's grouped by user and sorted accordingly 🚀 

Screenshot 2023-03-26 at 23 07 53

Screenshot 2023-03-26 at 23 12 16

### 💭 Extra...

I wrote the logic in `UserZaps` in an imperative approach, although I am sure it could be written in a more functional "kotlin way". I am open for suggestions, refactoring or any kind of improvement to the current logic 🧠


app/src/main/java/com/vitorpamplona/amethyst/service/model/zaps/ZapAmount.kt
```diff
@@ -0,0 +1,14 @@
+package com.vitorpamplona.amethyst.service.model.zaps
+
+import java.math.BigDecimal
+
+interface ZapAmountInterface {
+    fun total(): BigDecimal?
+}
+
+class ZapAmount(var amount: BigDecimal?) : ZapAmountInterface {
```
Chemaclass, line 9: I am not sure if this is needed anymore. It was an idea before, but I think I can safely remove it. I will try removing it to keep the code simpler.

vitorpamplona:
Good work. My only issue is with the idea of cloning a Note as opposed to creating a new UI object (similar to how we do the Notification Cards) to keep the Zap information. 

The issue with cloning notes is that, throughout the entire application, there is only one Note object per event_id. LocalCache enforces that. It makes it easy to deal with sets and maps because the same object references mean the same underlying `event.id` without having to compare the id string or the whole object. 

Here we should create a new UI Card object and make the Zap FeedView, model and filters return this new object instead of the clone of note.

Chemaclass:
Thanks for the feedback! I am not 100% sure how to do what you are saying, so I see two possible solutions:

1.- We can merge this PR and you can finalise it on your own, because you seems to know what is left.
2.- We can work on it together doing pair-programming; tomorrow evening (27), for example, or the Wed(29) or Thur(30) next week evening would work as well for me. I would love to do it as driver myself with your help navigating the solution.

What do you say, @vitorpamplona? 

>**Note**: Although I prefer the 2nd (pair-prog), if you choose the 1st (you do it) I will still learn form your diff-changes.

vitorpamplona:
Closing this idea for now. We may remove zaps from profiles altogether.

PR #319, opened 2023-03-26
Custom Messages for Zaps

believethehype:
- Includes adding a custom message to zaps in Backend Code (uses content field)
- Added a simple window on double tap on the bolt symbol for custom amount + message
- ! The popup/process might need some beautification
- ! Messages are not shown yet in Amethyst
- Window could provide options for non-zaps and private zaps in the future

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/note/ZapOptionsDialog.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/note/ZapOptionsDialog.kt
@@ -1,0 +1,180 @@
+package com.vitorpamplona.amethyst.ui.note
+
+import android.widget.Toast
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/note/ZapOptionsDialog.kt
```diff
@@ -0,0 +1,180 @@
...
+import com.vitorpamplona.amethyst.model.Account
+import com.vitorpamplona.amethyst.model.Note
+import com.vitorpamplona.amethyst.ui.actions.CloseButton
+import com.vitorpamplona.amethyst.ui.screen.loggedIn.AccountViewModel
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+class ZapOptionstViewModel : ViewModel() {
+    private var account: Account? = null
+
+    var CustomAmount by mutableStateOf(TextFieldValue(""))
```
KotlinGeekDev, line 45: Please, would it not be better to start the field name here(and below as well) with a small letter?

PR #330, opened 2023-04-04
Moving new pool button to NewPost

vitorpamplona:


toadlyBroodle (commented):
Looks good. Much cleaner UI without the optional extras and user input fields. Probably best to test the most basic functionality first.

PR #339, opened 2023-04-07
Prevent author vote manipulation

toadlyBroodle:
don't allow poll authors to vote in their own polls


vitorpamplona:
Shouldn't we allow people to vote on their own polls but disqualify them on when tallying side? We can't be sure clients are following the spec and people can always send events by hand.

toadlyBroodle:
> Shouldn't we allow people to vote on their own polls but disqualify them on when tallying side? We can't be sure clients are following the spec and people can always send events by hand.

Yes, most of the checks remain to be written on the tallying still. While we're still the only client emitting poll events, this still works, but all remaining tally side protocol checks will have to be written asap. I think it's best we don't allow authors to vote on their own polls, just so there's no confusion on what will be included in tallies, similar to how users aren't allowed to send zaps that won't be included in tallies.

Issue #351, opened 2023-04-15
[BUG]Quote repost is tied up to a post which is replyed a little while ago


ryogrid:
**Describe the bug**
When a user does quote repost, the puote rpost is tied up to a not related post with the quoted post.
It is not shown on TL but shown on a thread of the not related post.
(Probably, a post which is wrongly tie uped is one of being mentioned by the user most recent)

**To Reproduce**
1. do mention (reply) to some post which has no special condition
2. do quote repost ('Quote' on UI on amethyst)
3. quote repost is shown on thread of post mentioned at step-1 

**Expected behavior**
Quoted repost is shown on TL and it is not tied up to not related post.

**Video and Screenshots**
![quote-repost-bug-reproduction](https://user-images.githubusercontent.com/24614/232226802-efd7fcfa-9caa-4370-8bae-f69d4a2d8c0f.png)
post at bottom of above image is a reply post for reproduction.

**Device (please complete the following information):**
 - Phone Brand/Model : Sony Xperia SO-51A
 - Android Version: 12 
 - App Version: v0.32.3 and other older versions
   - apk is universal one and download from Release section of this repository

**Bounty (in Bitcoin sats) offered for a solution**
- 1k sats :)

vitorpamplona:
I believe this has been fixed as well. Let me know if it still happens.

PR #353, opened 2023-04-16
fix bug: quote repost is tied up to a post which is replyed a little while ago

ryogrid:
fix for #351

- When doing quote repost, NewPostViewModel::replyTos property often contains a Note object which is used for prior reply post
- In this modification, when the NewPostViewModel::load method's replyingTo argument is null (this means posting kind is 'Quote'),  the property is cleared by run call at line 75


vitorpamplona:
hum... wait... why is this property already not cleaned up? 

I think the issue is 5 lines before... It should be: 

```
} ?: run {
            replyTos = null
            mentions = null
        }
```

the `run` call ensures this block runs, instead of just returning it to the `replyingTo?.` expression

ryogrid:
@vitorpamplona 
I undestund.
I revise my modification according to your comment and do testing  :)

vitorpamplona:
Thank you @ryogrid!

PR #357, opened 2023-04-20
wip: add tor socks support

greenart7c3:
- [ ] Create a checkbox to use tor in the login screen
- [ ] Add a tor icon in the drawer
- [ ] Change the icon color or text when tor is enabled/disabled
- [ ] Test to see if the connection is using the proxy
- [ ] Add an pop up explaining how to setup
- [ ] In the pop up add two options with host and port of the proxy

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
@@ -46,6 +46,7 @@ import androidx.navigation.NavHostController
 import coil.compose.AsyncImage
 import com.vitorpamplona.amethyst.BuildConfig
 import com.vitorpamplona.amethyst.R
+import com.vitorpamplona.amethyst.ServiceManager
 import com.vitorpamplona.amethyst.model.Account
 import com.vitorpamplona.amethyst.model.User
 import com.vitorpamplona.amethyst.ui.components.ResizeImage
@@ -100,7 +101,12 @@ fun DrawerContent(
 }
 
 @Composable
-fun ProfileContent(baseAccountUser: User, modifier: Modifier = Modifier, scaffoldState: ScaffoldState, navController: NavController) {
+fun ProfileContent(
+    baseAccountUser: User,
+    modifier: Modifier = Modifier,
+    scaffoldState: ScaffoldState,
+    navController: NavController
+) {
     val coroutineScope = rememberCoroutineScope()
 
     val accountUserState by baseAccountUser.live().metadata.observeAsState()
@@ -197,11 +203,17 @@ fun ProfileContent(baseAccountUser: User, modifier: Modifier = Modifier, scaffol
                     })
             ) {
                 Row() {
-                    Text("${accountUserFollows.cachedFollowCount() ?: "--"}", fontWeight = FontWeight.Bold)
+                    Text(
+                        "${accountUserFollows.cachedFollowCount() ?: "--"}",
+                        fontWeight = FontWeight.Bold
+                    )
                     Text(stringResource(R.string.following))
                 }
                 Row(modifier = Modifier.padding(start = 10.dp)) {
-                    Text("${accountUserFollows.cachedFollowerCount() ?: "--"}", fontWeight = FontWeight.Bold)
+                    Text(
+                        "${accountUserFollows.cachedFollowerCount() ?: "--"}",
+                        fontWeight = FontWeight.Bold
+                    )
                     Text(stringResource(R.string.followers))
                 }
             }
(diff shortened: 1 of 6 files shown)


moonsettler:
got to watch for side leaks as well! where ever the HTTP client is used in the code, needs to be examined if pictures or other media or whatever is loaded it doesn't happen on clearnet!

if someone sends an image url in dm and the client pulls it from his server he basically got your ip even if the feed is queried in a privacy preserving manner.

greenart7c3:
> Do we need per-relay proxy setups?

No, I think it's better we create it for each Account

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -272,6 +306,44 @@ fun ListContent(
     if (backupDialogOpen) {
         AccountBackupDialog(account, onClose = { backupDialogOpen = false })
     }
+
+    if (openDialog.value) {
+        AlertDialog(
+            text = { Text(text = "Do you really want to disable tor?") },
```
vitorpamplona, line 312: Move to Strings.xml

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -272,6 +306,44 @@ fun ListContent(
...
+            onDismissRequest = { },
+            confirmButton = {
+                TextButton(
+                    onClick = {
+                        openDialog.value = false
+                        checked = false
+                        enableTor(account, false, openDialog)
+                    }
+                ) {
+                    Text(text = "Yes")
```
vitorpamplona, line 322: strings.xml

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -272,6 +306,44 @@ fun ListContent(
...
+                }
+            },
+            dismissButton = {
+                TextButton(
+                    onClick = {
+                        openDialog.value = false
+                    }
+                ) {
+                    Text(text = "No")
```
vitorpamplona, line 331: strings.xml

app/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedOff/LoginScreen.kt
```diff
@@ -221,6 +222,15 @@ fun LoginPage(
                 }
             }
 
+            Row(verticalAlignment = Alignment.CenterVertically) {
+                Checkbox(
+                    checked = useProxy.value,
+                    onCheckedChange = { useProxy.value = it }
+                )
+
+                Text("Enable Tor")
```
vitorpamplona, line 231: strings.xml: Probably change to "Connect via Tor" to clarify

app/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
```diff
@@ -259,6 +277,22 @@ fun ListContent(
             onClick = { backupDialogOpen = true }
         )
 
+        var textTorProxy = if (checked) "Disable" else "Enable"
```
vitorpamplona, line 280: Strings.xml

vitorpamplona:
nice work! This will need a Tor already running, correct?

moonsettler:
> This will need a Tor already running, correct?

I think it just needs Orbot installed and will start it automatically you just need to wait a bit for the circuit to build. Would have to try it out to be sure.

vitorpamplona:
> I think it just needs Orbot installed and will start it automatically you just need to wait a bit for the circuit to build. Would have to try it out to be sure.

If this is true, is there any way to see if Orbot is installed and not show the option to everyone else?

moonsettler:
> > I think it just needs Orbot installed and will start it automatically you just need to wait a bit for the circuit to build. Would have to try it out to be sure.
> 
> If this is true, is there any way to see if Orbot is installed and not show the login option (drawer is fine) to everyone else? Then, It would be nice if the option said: "Connect through your Orbot setup".

no guarantees, but maybe something like this? (i can't try it out)

```Kotlin
fun isPackageExist(context: Context, target: String): Boolean {
     return context.packageManager.getInstalledApplications(0).find { info -> info.packageName == target } != null
}

isPackageExist(context, "org.torproject.android")
```

greenart7c3:
Changed the strings and now i'm checking if orbot is installed in the login screen.

greenart7c3:
Should we have a configuration dialog or should we always use the default values?
For now I added it to the drawer but not the login

moonsettler:
This looks almost done! Awesome!

PR #365, opened 2023-04-21
should fix the notes own zaps from influencing the poll percentages

djschnei21:


djschnei21:
damn you're fast... :D

PR #372, opened 2023-04-23
Update for Custom Zaps

believethehype:
Based on last selection in ZapCustomDialog, Reaction Row Zaps will use the same zap type

Currently Zaps in Polls are always public.


believethehype:
Additionally in this PR, Private Zaps get decrypted and shown in the notifications. 

Zap Messages are now also shown in Notifications.

PR #381, opened 2023-04-27
Add support for non-english hashtags too

vivganes:
**Background:**
Right now, we can use only english hashtags.  This can be limiting for non-english users.

Example:
image

**Screenshot after fix:**

image

Edit: I took the solution from https://stackoverflow.com/a/27343333


vitorpamplona:
Nice!

PR #386, opened 2023-04-30
updated descriptions for Image uploader

believethehype:
removed "less secure" from titles and adapted descriptions. While it is true that Nip94 is more secure, it might be too confusing for users that simply want to upload a cat image. Also added a hint that media might not be visible for people using clients that have not implemented NIP94 yet.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -327,19 +327,19 @@
     <string name="file_server">File Server</string>
     <string name="zap_forward_lnAddress">LnAddress or @User</string>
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
A few things here: 

1. We need to warn people that they are at the mercy of their image servers for 1 and 2. Many people come to Nostr because it's decentralized. They trust us to keep it so. Image servers are not decentralized. 

2. We need to explain what Verifiable means. What we are doing to fix the problem with 1. 

3. NIP-95's "without using a third-party host" is incorrect. The relay is a third-party host. What we won't have in NIP-95 is a fixed URL.

PR #391, opened 2023-05-02
Added example for Image upload on nostr.build

believethehype:


believethehype:
added NIP94 support for nostr.build

PR #412, opened 2023-05-15
Swedish translation

Pextar:


vitorpamplona:
Thank you, @Pextar! What's your pubkey?

PR #445, opened 2023-06-08
WIP: Nip26

greenart7c3:
WIP DO NOT MERGE !!!

Making this draft so everyone can see what decisions we are making while developing the nip 26 in amethyst.

1. Qr code data should be black QR data on a white background so the HWW can read it
2. The timestamps in the delegation string should be in seconds, not milliseconds

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -193,10 +193,14 @@ dependencies {
     // immutable collections to avoid recomposition
     implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.5")
 
+
     // Video compression lib
     implementation 'com.github.AbedElazizShe:LightCompressor:1.2.3'
     // Image compression lib
     implementation 'id.zelory:compressor:3.0.1'
+
+    implementation "androidx.documentfile:documentfile:1.0.1"
+
 
     // Automatic memory leak detection
     debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.11'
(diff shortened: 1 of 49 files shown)


vitorpamplona:
Keep in mind that NIP-26 might turn into a deprecated NIP soon. People are migrating to NIP-46 "delegations" instead with the nSec bunker that Pablo has been pushing for. 

We can still implement it for completeness, but I don't expect many new supporters for this NIP (unless public feeling turns 180).

weex:
> Keep in mind that NIP-26 might turn into a deprecated NIP soon. People are migrating to NIP-46 "delegations" instead with the nSec bunker that Pablo has been pushing for.

Where can we find more info about this alleged deprecation? I couldn't find anything about that in https://github.com/nostr-protocol/nips

Delegation ala NIP-26 seems superior to requiring a hardware device or additional service to interactively sign each message.

vitorpamplona:
There is some stuff here: https://fiatjaf.com/4c79fd7b.html But essentially no one likes NIP-26 because you have to add custom cases to display the delegator information in ALL event types. What happens in voting delegation with Zap Polls? Imagine using NIP-95 with delegated keys. There is so much that can go wrong. 

The second issue is that once the delegated keys leak, the attacker that got the keys can now post on behalf of the delegator forever. Even if dates are included in the token (`"kind=1&created_at>1674834236&created_at<1677426236"`)  that only means that relays cannot accept before and after those dates. All an attacker needs to do is to keep posting as the delegator with created_at dates inside that time period. The attacker can create DMs in the past, that are valid. It can create reports, vote on polls, etc. 

NIP-26 always raised a ton of security questions for me. As Nostr grows, more and more use cases will need to specialize their behavior to account for NIP-26 delegations and rules (expirations).

weex:
Excellent points, thanks for the summary.

greenart7c3:
I'll close this since most of the relays and clients doesn't support nip 26.
It's also a really bad user experience using nip 26 if we can't filter for the delegation tags from the relays.
Maybe in the future we can have a better experience for those who want to leave their keys in a HWW

PR #447, opened 2023-06-08
Add weed hashtag

TonyGiorgio:
Feel free to ignore if you don't want it but I kinda like it plus some people were asking me for it. Thought I would see if you'd like it too. 

signal-2023-06-08-111340_002


vitorpamplona:
This is great! Thanks Tony!

PR #462, opened 2023-06-21
Relay information dialog

greenart7c3:


vitorpamplona:
If you want to keep going, you can replace the links in the Relay icons with your new view :)

vitorpamplona:
Just added :)

PR #486, opened 2023-07-03
Optional media pre-loading

### LINKED ISSUE #435: [FEATURE] Don't load media when not connected to WiFi
Amethyst is difficult to use while traveling, as poor cell phone reception can cause timelines with several media attachments at the time to load slowly or not at all.

A non-default option to not load media when not connected to WiFi, so I can scroll past those posts while on cellular data. The posts should display a button, similar to the recent nsfw/sensitive button, to click thru this and load the media. That way I can choose what posts to load media on when I'm on the go.

Dont have sats to offer for this

**Additional Details** 
This is similar to something offered by the Fedilab app, which is a similar app to Amethyst but for ActivityPub instances

greenart7c3:
DO NOT MERGE !

Fixes #435

Current ui:

image

if you have any ideas how the ui should look please comment here with some example


greenart7c3:
cc @vitorpamplona 

Do you think I should create a settings page or we can continue without one for a while?

Also should i load profile pictures, banners etc or just stop loading the content of the notes?

vitorpamplona:
FYI, the Checkbox is quite ugly. You can use the Switch compose on the right of the text to make it similar to the Android settings page.

greenart7c3:
image

davotoula:
Wow... looking great so far. Loving being able to follow the development.

Can we consider adding preview of links to this feature? It would make a truly snappy and lightweight experience on slow mobile connection.... However happy to see it being added later if it speeds up the release of this.

well done!

greenart7c3:
@vitorpamplona @davotoula 
Should I move the zaps and custom emojis configuration to this settings page?

greenart7c3:
I'm also still loading profile pictures and verifying nip 5. Should i disable that too?

davotoula:
Personally I would prefer that. 

It's more natural (from other apps) to single click to do default emoji/zap and hold down to select alternative. 

I often make mistake to edit custom zaps/emojis instead of selecting them when holding down. 

If custom emoji/zap lives in settings there is less chance for error. 

Let's see what @vitorpamplona says, I understand he may be concerned about settings bloat. 

Also, maybe better as a future enhancement to ship the original media loading feature quicker.

davotoula:
Small, well defined enhancements with shorter development time are so much easier to code review and to release!

vitorpamplona:
This is getting good! 

A few comments: 
- I wouldn't move the Account Preferences (Relays, filters, backup, Tor) to this page yet. Apologies if my earlier post confused things. They could stay in the sidebar for now to avoid breaking the Settings Screen into two UI concepts: (i) buttons to open dialogs and (ii) combo boxes with save button. We will have to add Zap/NWC setup, Reactions Setup, and Translations setup to these Account Preferences. So, we might leave it for later. 
- Can we make these combo boxes look better? On the image upload screens (NewPostView.ImageVideoDescription), we use a TextSpinner for servers. That looks better, I think. 
- The save button should either be at the top, in a similar shape that other dialogs/new posts/etc do, with a way to cancel or the change should take effect immediately (no save action).

greenart7c3:
image

KotlinGeekDev:
This is really nice @greenart7c3 . But isn't it better to have an explicit save button? IMHO it is useful for visual feedback. I guess you could implement it like the Post button, disabled when settings are saved, enabled when Settings are modified.

greenart7c3:
Yes, I find the button better. I also need to improve the design of the video player and the load image placeholder

image
image

KotlinGeekDev:
@greenart7c3 Yeah. IMHO the 'load image' button could be much smaller. As for video, I am thinking of two options: Either using a button just like the image option, or having a blurred(or not) thumbnail. 
cc @vitorpamplona

vitorpamplona:
We won't have a thumbnail without downloading the entire image (only available in NIP-94/95 images). :( 

Maybe display the URL with a small inline Load button?

KotlinGeekDev:
@vitorpamplona that seems to be a good idea. The button could be put at the bottom right, for example, with the URL link in the center.

greenart7c3:
You mean show the images and videos like this?

image

vitorpamplona:
This is the LoadingAnimation we have that displays while the image is loading. We could do a similar size loading button

greenart7c3:
image

vitorpamplona:
That works. It is hard to press the button? Too small?

greenart7c3:
For me with wasnt hard, but i think for other people might be depending on the size of the finger and the device size

KotlinGeekDev:
The logos are more appealing than the simple button.

vitorpamplona:
@greenart7c3 let me know when you are ready for a merge on this

Issue #390, opened 2023-05-01
[BUG] Japanese text is misidentified as URL.


mattn:
**Describe the bug**
Some Japanese text is unexpectedly misidentified as URL.

**To Reproduce**
1. Open input dialog
2. Type some Japanese characters contains `。`  (`\u3002`)
3. Post the note
4. See the text is misidentified as URL.

**Expected behavior**
The  text should be normal text.

**Device (please complete the following information):**
 - Android Version: 0.37.4 

This is related on linkedin/URL-Detector. https://github.com/linkedin/URL-Detector/issues/39

URL-Detector handle `。` as dot. This is not a bug because IDN allow to use `。` as dot. However, most of Japanese text are often misidentified.

![image](https://user-images.githubusercontent.com/10111/235552699-846bd525-d71d-4a7c-b6fc-bbcced9db926.png)

afternooncurry:
Ideographic full stop and full width period are not handled as dot in IDNA2008 while IDNA2003 does. A current recommendation is IDNA2008. linkedin/URL-Detector looks implemented based on IDNA2003 about IDN which may causes the issue.

mattn:
@vitorpamplona Many users in the East Asian region have been waiting for this fix for a long time.

vitorpamplona:
We are waiting for the library to be able to support these additional characters. Until that library is fixed, there is not much we can do :(

mattn:
Thanks. FYI, This issue can be reproduced with English speakers.

![image](https://github.com/vitorpamplona/amethyst/assets/10111/9a519d4d-6261-41be-837c-be627f0c4f22)

vitorpamplona:
Interesting. That is a different "bug". 

We have a separate procedure to linkify all `yyy.xxx` texts. It requires the domain separator `.` and should not affect the `\u3002` character. Is that an issue for you. I see it here and there during the week but always just ignore it because it works most of the time.

vitorpamplona:
And now that urls can have any Unicode character, I am not really sure how to solve it. Because the `Thanks.<emoji>` is a valid URL these days.

mattn:
I don't make sure what is wrong in the code but this is rendered as URL in only amethyst.

![image](https://github.com/vitorpamplona/amethyst/assets/10111/b9d956ab-9f47-47a2-b260-6fe01c0c218f)

mattn:
Is this related on this issue?

[...]

PR #491, opened 2023-07-05
Fixed a bug in Japanese string display

ShinoharaTa:
Nice to meet you. I am an engineer in Japan, usually making web applications, etc.

I was wondering about the bug in issue #390 and thought I would fix it.

The following images are before and after the fix.

[before 1]

[after 1]

[before 2]

[after 2]

The fix is to make it more strict to determine if it is a URL or not.
First of all, "." is used. is mistaken for Japanese ". is now corrected by hard-coding.
In addition, URLs without a scheme are now allowed to use the characters used in URLs.

Note that this fix affects all users of the timeline.
We have prepared a pre-built APK so that you can check it before release.

https://drive.google.com/drive/folders/1vTfyPIsvP58_X3Xx26nSc8aGfRX6XonN?usp=sharing

The other day I was looking at the Amethyst timeline and saw a bug where a Japanese string was interpreted as a URL.

About 6 hours ago, I saw it posted and decided to fix it.

nevent1qqsdu79vq5wradvquk32upfd5amgrla448v4vv69sa3sr4dy6w6j0vcpz9mhxue69uhkummnw3ezuamfdejj7qf9waehxw309a6ku6tkv4e8xefwdehhxarjd93kstnvv9hxgteld3skueeadfsslwjtwm

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/service/CachedRichTextParser.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/service/CachedRichTextParser.kt
@@ -70,8 +70,7 @@ class RichTextParser() {
             } else if (it.originalUrl.contains("。")) {
                 null
             } else {
-                val pattern = "^(http|https)://([A-Z0-9][A-Z0-9_-]*(?:.[A-Z0-9][A-Z0-9_-]*)+):?(d+)?/?".toRegex(RegexOption.IGNORE_CASE)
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Nice! Thank you @ShinoharaTa !

ShinoharaTa:
Thanks for the quick merge.

PR #507, opened 2023-07-14
change branding from NIP-05 to Nostr address

derekross:
This changes the string values from "NIP-05" to "Nostr address" to reflect a less technical and more user friendly branding for users that wish to edit this field.


vitorpamplona:
Sorry, I just saw this PR now.... :( Closing since it was also implemented here: https://github.com/vitorpamplona/amethyst/commit/64a5071345065c690e1b9e46f3539e519ecffc4c

PR #510, opened 2023-07-16
Update strings.xml in values-ta

AutumnSunshine:
Tamil strings for Settings page & Nostr Address


AutumnSunshine:
Please ignore this - there seems to be an error though I can't spot it.
Will check & submit a proper file later.
Thanks!

PR #519, opened 2023-07-24
Update PRIVACY.md

Ben3EDF5301:
Made the terms of use more appropriate for free and open source software that has been distributed with an MIT license.


vitorpamplona:
These terms are not for the code. They are for the distributed version through Google Play.

PR #531, opened 2023-07-29
Grouped notifications.

### LINKED ISSUE #501: I want to group Amethyst notifications.
My device is a Pixel 7a with Android 13.
I am concerned about the behavior of Amethyst when running in this environment, so I have created this Issue.

Each Amethyst notification that arrives at my terminal is displayed separately.
Ideally, I think notifications from the same application should be grouped together, like Gmail.
When I receive a lot of Zaps as well as DMs, the notification box is sometimes overflowing with unread messages, and I suggest that this should be improved.

*npub:npub1l60d6h2uvdwa9yq0r7r2suhgrnsadcst6nsx2j03xwhxhu2cjyascejxe5
*my account: shino3

KotlinGeekDev:
Potentially fixes #501 . 
[Per the docs](https://developer.android.com/develop/ui/views/notifications/group#group-notification), using the same group key can solve this(especially since the app's min. target is API 26).

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/service/notifications/NotificationUtils.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/service/notifications/NotificationUtils.kt
@@ -18,7 +18,8 @@ import com.vitorpamplona.amethyst.ui.MainActivity
 object NotificationUtils {
     private var dmChannel: NotificationChannel? = null
     private var zapChannel: NotificationChannel? = null
-    const val NOTIFICATION_GROUP_KEY = "com.vitorpamplona.amethyst.NOTIFICATION"
+    private const val DM_GROUP_KEY = "com.vitorpamplona.amethyst.DM_NOTIFICATION"
+    private const val ZAP_GROUP_KEY = "com.vitorpamplona.amethyst.ZAP_NOTIFICATION"
 
     private fun getOrCreateDMChannel(applicationContext: Context): NotificationChannel {
         if (dmChannel != null) return dmChannel!!
@@ -71,7 +72,7 @@ object NotificationUtils {
         val zapChannel = getOrCreateZapChannel(applicationContext)
         val channelId = applicationContext.getString(R.string.app_notification_zaps_channel_id)
 
-        sendNotification(id, messageBody, messageTitle, pictureUrl, uri, channelId, applicationContext)
+        sendNotification(id, messageBody, messageTitle, pictureUrl, uri, channelId, ZAP_GROUP_KEY, applicationContext)
     }
 
     fun NotificationManager.sendDMNotification(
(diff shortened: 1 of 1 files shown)


KotlinGeekDev:
cc @vitorpamplona

vitorpamplona:
Doesn't that group everything? Like zaps and DMs together from multiple people?

KotlinGeekDev:
@vitorpamplona True. Will fix that right away.

KotlinGeekDev:
@vitorpamplona done :+1:

vitorpamplona:
Thanks @KotlinGeekDev

vitorpamplona:
Did this work for you @KotlinGeekDev ? I now only receive the first notification. It doesn't add the others in on my Pixel 7

KotlinGeekDev:
@vitorpamplona I barely receive notifications, so I have not seen it in action yet. However, if this does not work, it means the docs are incorrect. I'll send another PR to correct this.

vitorpamplona:
So, the original code was supposed to group DMs by the author of each message: 3 messages from the same author were supposed to be grouped already. The new one groups everyone in the same DM notification, which I think is less desirable. 

Nevertheless, neither the original implementation, nor the new one group people correctly for me. :( I am not sure what's missing

KotlinGeekDev:
@vitorpamplona It seems we may have to create a seperate grouping(summary) notification, that does just that, per the docs.

KotlinGeekDev:
@vitorpamplona there are two potential solutions:
1. We use the same channel ID, while keeping the group keys, or
2. We remove the setGroup() and the keys altogether.

vitorpamplona:
> @vitorpamplona It seems we may have to create a separate grouping(summary) notification, that does just that, per the docs.

Maybe we create one by querying active notifications? We could group DMs by user, get all their messages, sort by created_at and join them with a `\n`

KotlinGeekDev:
> > @vitorpamplona It seems we may have to create a separate grouping(summary) notification, that does just that, per the docs.
> 
> Maybe we create one by querying active notifications? We could group DMs by user, get all their messages, sort by created_at and join them with a `\n`

That is a very interesting solution, but it seems it will be more complex than adding a summary notification(as it seems we will need it anyway).

vitorpamplona:
hum... what I described is to assemble the summary notification. Otherwise, where do you get the Notification text from?

KotlinGeekDev:
Oh, I see what you mean @vitorpamplona . I was instead thinking of using a generalized message for the summary notification text, such as "New DMs received", and "New Zaps received".

vitorpamplona:
That doesn't sound very informative for the user. We should copy what WhatsApp and Telegram do. 

I will revert the change for now until we find a better solution

KotlinGeekDev:
@vitorpamplona  you're right. No need to revert the change though. I can just add the proposed changes to it.

KotlinGeekDev:
@vitorpamplona How about zaps? How could the summary notification look like?

vitorpamplona:
Maybe zaps need to always be a new notification. The issue is with DMs, that the same person sends 5-10 messages in the same minute. That doesn't happen with zaps.

PR #533, opened 2023-07-30
Replace hard coded text with resource text and update translations

davotoula:
Added missing translations cs/de/se
Replaced hard coded text with resource 
Added new translations for cs/de/se
Corrected "occurred"


vitorpamplona:
Thank you @davotoula

PR #534, opened 2023-07-31
Fixes group notifications.

KotlinGeekDev:


KotlinGeekDev:
cc @vitorpamplona

KotlinGeekDev:
@vitorpamplona Please, when receiving a DM, see whether the sender's image appears correctly. :sweat_smile:

vitorpamplona:
It will only appear if the profile is loaded in the app.

vitorpamplona:
Are you planning to split the groups per user? So we can provide a summary that includes the messages themselves?

KotlinGeekDev:
@vitorpamplona yeah, I am planning to do that. But I think it will be better to do it in another PR. :sweat_smile:

KotlinGeekDev:
@vitorpamplona I think we can merge this for now(except if you have some concerns). :sweat_smile:

vitorpamplona:
I am not sure if you saw that when I reverted the original change, I removed the setting of a group because Android tries already to do that automatically. This is how it groups for Zaps: 

Screenshot 2023-08-02 at 1 31 57 PM

It seems to work for messages ass well. And if that is the case, then grouping like this might not actually need a custom Summary implementation.

KotlinGeekDev:
@vitorpamplona that's true. I guess this PR can be closed then.

PR #542, opened 2023-08-07
fix signed byte used for array slice with + added test

frnandu:
In case L was above 127 it would be signed int and ArrayIndexOutOfBoundsException would be thrown on the slice array.
Added test with nevent1 to reproduce and fix with .toUByte().toInt()


vitorpamplona:
ohhh, very cool! Thanks @frnandu

PR #543, opened 2023-08-07
Add Amber support

greenart7c3:
In this fist version we will always open amber to get the signature from the event and in the future we need to do something in amber to always allow some kinds for an application using contentResolver, services or getting the result of the intent


greenart7c3:
I'll try to find a way to get from where the intent is coming

greenart7c3:
Using rememberLauncherForActivityResult i can get the package name

vitorpamplona:
Check this out as well 

https://developer.android.com/reference/android/app/Activity.html#getReferrer()

greenart7c3:
This one returns an uri with the package name com.vitorpamplona.amethyst.debug but only if i use the rememberLauncherForActivityResult.
With this i can also send the result back to Amethyst and set the event signature without using the clipboard

vitorpamplona:
Not sure if you will need all events on Amber, but if you do now you can load them all from here via jitpack.io

```
    implementation('com.github.vitorpamplona.amethyst:quartz:main-SNAPSHOT')
```

greenart7c3:
> Not sure if you will need all events on Amber, but if you do now you can load them all from here via jitpack.io
> 
> ```
>     implementation('com.github.vitorpamplona.amethyst:quartz:main-SNAPSHOT')
> ```
I get this error when trying to decrypt nip44 using quartz. Had to copy the quartz folder to Amber to work

java.lang.unsatisfiedlinkerror: native library (com/sun/jna/android-aarch64/libjnidispatch.so) not found in resource path

vitorpamplona:
> libjnidispatch.so) not found in resource path

Try adding `implementation 'net.java.dev.jna:jna:5.13.0@aar'`. I don't know why the implementation is not passing these native libs along :(

greenart7c3:
I think you can start the review now.
It's working but theres a lot to refactor later
I don't know if we should do it this way but the nip24 dms, private zaps and nip04 dms are always trying to decrypt so the app keeps opening amber when new dms arrives or when you open the dms screen

PR #545, opened 2023-08-09
fixing global relays selection

jeremyd:
Global feed was not filtering by relay global toggle settings.  This patch fixes that.  Now when you have relays set to Global feed, the feed only displays messages from those relays.  Tested and working, I'm a kotlin noob so I don't know if this is the best way to fix it but it works!


vitorpamplona:
Thanks @jeremyd!

Issue #536, opened 2023-08-02
[BUG] URLs containing "-" may not become clickable link.


shion72:
**Describe the bug** 
The URLs that containing FQDN written in the following format does not become clickable link.
https:// **aaa.b-b.ccc** /

 **To Reproduce** 
Post the URLs that containing FQDN written in the following format.
https:// **aaa.b-b.ccc** /

**Expected behavior** 
URLs become clickable.

**Video and Screenshots** 
<img src="https://github.com/vitorpamplona/amethyst/assets/132278089/da0ab69b-3b23-4762-926d-a22cab00720a" width = "270px" height = "100px"> </br> <img src="https://github.com/vitorpamplona/amethyst/assets/132278089/a964e8d1-5046-4cc7-9c19-7a24b47a1491" width = "270px" height = "100px">

**Device (please complete the following information):**
 - Phone Brand/Model [Sony Xperia SO-52A]: 
 - Android Version [12]: 
 - App Version [0.71.0-PLAY]:

KotlinGeekDev:
@vitorpamplona I think this can be closed.

PR #549, opened 2023-08-15
fix url regex pattern.

ShinoharaTa:
Fixed in response to issue #536.
Two screenshots are attached to the evidence.

before

after

My npub is here.
npub1l60d6h2uvdwa9yq0r7r2suhgrnsadcst6nsx2j03xwhxhu2cjyascejxe5


KotlinGeekDev:
Hi @ShinoharaTa . Thanks! It seems the issue has been fixed [here](https://github.com/vitorpamplona/amethyst/commit/4bdb5a28d0854ade1f32b58f265ffd5576ffd2d8#diff-93c37930e7b438d0320a081e4260867b76b6485ac95674c3df3b8af351b34670). Please can you check the latest version to see if it is OK?

vitorpamplona:
Thank you, @ShinoharaTa! This works.

KotlinGeekDev:
> Hi @KotlinGeekDev
> 
> Yes, I am aware of this commitment. But my point of correction is different.
> 
> The problem was due to a misdescription of the RegEx pattern, which was not resolved in the latest main branch. I have built and verified it on the actual machine.
> 
> 

Nice :+1:

PR #552, opened 2023-08-21
fix crash when getting firebase token without play services

greenart7c3:


vitorpamplona:
Thanks @greenart7c3 !

PR #569, opened 2023-09-05
Add arm64-v8a and armeabi-v7a build target

atrifat:
This PR will add arm64-v8a and armeabi-v7a as additional build target. I think it can be useful for users to download smaller apk build compared to large universal apk build. Zeus for example provided them as the option in their  [Releases](https://github.com/ZeusLN/zeus/releases).

Thank you.


vitorpamplona:
Thank you @atrifat

PR #574, opened 2023-09-06
Updated translations for cs/de

davotoula:
Resubmitting translations. Previous edits were reverted by Crowdin workflow.


davotoula:
ooo... that's a neat feature but it will make things more difficult going forward. Personally I identify missing translations in the localised files and translate them. If the translation is same as English so be it but next time I will not see them as missing translation.

Personally I would switch that feature off if possible. Is it on for space saving only?

vitorpamplona:
hum... let me investigate to see if there is a way to do it.

davotoula:
I really like crowdin now. It will be my preferred way to update translations. Also will not pollute commit history and I don't have to mess with ide/git 👍

davotoula:
I really like crowdin now. It will be my preferred way to update translations. Also will not pollute commit history and I don't have to mess with ide/git 👍

PR #604, opened 2023-09-20
Change nav bar color according to the theme

greenart7c3:


vitorpamplona:
Nice! Thank you @greenart7c3

PR #605, opened 2023-09-20
immersive scrolling on home feed

greenart7c3:
https://github.com/vitorpamplona/amethyst/assets/115044884/d2fbc6b7-a615-4b26-91ca-107e0bbdd472


vitorpamplona:
ohhh I wanted to do this for so long! nice!

Issue #598, opened 2023-09-16
[BUG] long note not folded if low character count


libertas-primordium:
**Describe the bug**
A note with a lot of newlines but low character count is not automatically folded in the feed view.

**To Reproduce**
See this example:

nostr:nevent1qqs0lk0a0qg8ulha6lgp4yk8t09mgesy272qqyqpytpgwk8snr0m4tspz3mhxue69uhhyetvv9ujuerpd46hxtnfdupzqrgxfq9scm3mu0y6rfja0e4uyzgjyl24hax80m4kqda6wamvxq8vqvzqqqqqqy08fv2t

**Expected behavior**
This note is very long. It should be folded. 

**Device (please complete the following information):**
 - Phone Pixel 7 Pro
 - Android Version 13
 - App Version 0.76.0-PLAY

davotoula:
I'm looking into this.

davotoula:
It's a pretty straight forward issue.

The "Show more" code counts line break (\n) as a single character while `\n` in fact takes a whole row.

I'm thinking a fix that will count empty lines as a full row of characters.

OrigiSize:
@libertas-primordium Can you add a more descriptive title?

libertas-primordium:
Whoops! Yes I'll update the title.

davotoula:
https://github.com/vitorpamplona/amethyst/pull/608

PR #608, opened 2023-09-21
"show more" text folding enhancement

davotoula:
Enhancement to prevent a note with mainly line breaks to render as an extremely long note:
Introduce a line limit (default 10) in addition to the current character limit
Min of characters / lines is used for note folding

Fix for: https://github.com/vitorpamplona/amethyst/issues/598


davotoula:
@vitorpamplona the commit history got mangled a bit (doubled).

Let me know if you want me to resubmit a pull request with cleaner history.

vitorpamplona:
Nice! thank you @davotoula

PR #620, opened 2023-09-25
Fix contact link in relay information dialog

jiftechnify:
- Added specific handling for the `https:` schema
- Stop making the text clickable if it is neither of a `https:` URL nor an email address
- Strip `mailto:` schema from email addresses


vitorpamplona:
very nice, thanks @jiftechnify

PR #621, opened 2023-09-25
add notifications for users of amber

greenart7c3:
This just works if the user checked the remember my choice for decrypting nip44
I havent found a way to open amber when amethyst is in background. Seems to be a limitation since android 10


vitorpamplona:
This account-switching business in a singleton is probably not a good idea. 

I am a little worried the cachedDecryptedContent will leak between accounts. Maybe we should break the signer into the singleton/service part and the cached part, which should be per account.

PR #634, opened 2023-10-03
Updated translations for cs/de/sv

davotoula:
This was a large amount. Easier to do in IDE rather than in Crowdin!


vitorpamplona:
Thank you @davotoula

PR #636, opened 2023-10-06
fix floatingActionButton not hiding when scrolling

greenart7c3:


greenart7c3:
It's a bug introduced by the new material3 changes.
Vitor removed the animation because when opening the app it was not showing the button

image

PR #637, opened 2023-10-06
change defaults ConnectivityType to wifi only

greenart7c3:


vitorpamplona:
Is this a good idea? The default experience without images kind a suck :(

greenart7c3:
Don't know if it's a good idea. I think a lot of people will complain
Maybe we do it after we have a better ui for images and videos

KotlinGeekDev:
Personally, I am not sure it is a good idea. Maybe for this issue, we should run a poll by users?

davotoula:
Personally I love the "media over wifi only" option but I think for new users it would be a poor first user experience.

"wifi only" is for more experienced users.

VictorieeMan:
> Personally I love the "media over wifi only" option but I think for new users it would be a poor first user experience.
> 
> "wifi only" is for more experienced users.

Yeah, unless it looks good for new users it probably isn't a good idea. That's a very good point. The first impression on new users is where they decide to stay or not. That just better look good.

greenart7c3:
I'll close this for now. Maybe when we have a better ui we revisit this.

Issue #626, opened 2023-09-27
[BUG] regression: selecting "mute" list doesn't show any notes 


davotoula:
**Describe the bug**
In previous versions I was able to select the "mute" list and see notes that are normally hidden in my main feed. This was a good feature to find something interesting that was otherwise hidden in my main feed. 

In recent Version mute list is empty. 

**To Reproduce**
1. Go to home
2. Click on lists in middle top of screen 
3. Scroll down to "mute" list 
4. No notes are shown 

**Expected behavior**
Notes normally hidden should be shown 

**Video and Screenshots**
![Screenshot_20230927-211443.png](https://github.com/vitorpamplona/amethyst/assets/1747287/ec466cc6-3c58-431f-9881-7e23e31f0e60)

**Device (please complete the following information):**
 - Phone Brand/Model Pixel 7 Pro
 - Android Version [e.g. 33]: 33
 - App Version [e.g. v0.20.3]: 0.78.2

**Bounty (in Bitcoin sats) offered for a solution**
1k sats for showing all muted notes (mute list, Spammer Filter, muted words)

davotoula:
resolved

davotoula:
resolved in [v0.79.4](https://github.com/vitorpamplona/amethyst/commit/202af650de0793978c33b5878e153d3879a41e97)

PR #639, opened 2023-10-09
Bugfix mute feed is always empty

davotoula:
In noticed that Home feed new threads and conversations when mute list is selected at some point stopped showing.

It seems that the mute list changed at some point from simple "mute" to a combination of event code, user hex and mute.

This is a simple fix to bring back the mute list.

I noticed similar issue in chat feed, community feed, live feed and video feed but I haven't fixed it there.

See:
https://github.com/vitorpamplona/amethyst/issues/626


davotoula:
Please let me know if this should also be corrected in any of:

- chat feed
- community feed
- live feed
- video feed

PR #643, opened 2023-10-12
feat: enable 100sats as default zap

emidev98:
This pull request enables a default zap of 100 Sats


davotoula:
What is the reasoning behind this change please?

emidev98:
Hey @davotoula, I just feel like I want to spam people with 100sats and at the moment that option is not available for the app. The first approach I thought about was to code a new screen to allow the user to pick any quantity they want but then I realized that I had not enough time to do that.

davotoula:
If you long hold the zap button you can add or remove zap amounts

emidev98:
LoL! Did not know that! Thanks for that it's being very helpful!

PR #653, opened 2023-10-20
fix double encryption when sending dms, when login with external signer set privKey to null to be sure we have no private keys

greenart7c3:


vitorpamplona:
hum... We need to figure out a better way to converge the two signing procedures between Local and Amber.

PR #656, opened 2023-10-23
fixes switching from nsec to npub

### LINKED ISSUE #655: [BUG] When logging in and out with nsec and then npub it's not updating the UI with the right account
**Describe the bug**
When logging in with a nsec and then with npub it's not updating the current account with the right permissions 

**To Reproduce**
1. Create a new account
2. Copy it's npub
3. Logout
4. Login with npub
5. React to any post and you will see that it reacted even if you don't have a nsec

**Expected behavior**
When logout it's not updatng the UI with the correct account. In my tests I had to close the app and open it again to work

**Video and Screenshots**

greenart7c3:
fixes #655 

Looks like there's something not clearing when logging out and always using the old account variable.
For now i'm recreating the activity to fix this


vitorpamplona:
ohhh I found the issue. It's the ViewModel cache of AccountViewModel because the key is just the pubkey of the user.

vitorpamplona:
https://github.com/vitorpamplona/amethyst/commit/f072b98fd76dba9bae5f97f9c793280a4161bb1d should have fixed this.

PR #669, opened 2023-10-28
changing externa sharing service for text notes to njump.me

VictorieeMan:
Changed "https://snort.social/e/" to "https://njump.me/"

Benefits of this change are the link previews that njump.me generates for shared content. The snort social links are just the url root and a long nevent, making different shares difficult to tell apart, and a bit scary to click on for some.

More benefits of njump.me listed on https://njump.me/, hosted by fiatjaf. The included nostr onboarding paired with client agnosticism is also beneficial.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteQuickActionMenu.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteQuickActionMenu.kt
@@ -99,7 +99,7 @@ val externalLinkForNote = { note: Note ->
         if (note.event is FileHeaderEvent) {
             "https://filestr.vercel.app/e/${note.toNEvent()}"
         } else {
-            "https://snort.social/e/${note.toNEvent()}"
+            "https://njump.me/${note.toNEvent()}"
         }
     }
 }


vitorpamplona:
Does jump.me now supports most events we do? I don't know if they have a running list somewhere. 

Last time I looked they only supported 3-4 events kinds or so. Which means most of the links will be broken when shared. We use Snort because it's closer to Amethyst in terms of feature completeness :(

VictorieeMan:
Good point. This is what I thought the if statements in [externalLinkForNote()](https://github.com/vitorpamplona/amethyst/blob/1f28d3f32ff7e7624b2dc32079368684387e57f0/app/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteQuickActionMenu.kt#L87C5-L87C25) were dealing with.

Is there any complete list of what events that Amethyst support, or is some cross-checking detective work needed with the supported NIP list and what events they contain? We could recommend extended event support in njump.me to @fiatjaf. If I familiarize myself with that codebase, I could probably suggest a patch.

Ideally we develop njump.me to support all event kinds, Since it's onboarding, preview making, search indexing and client agnosticism makes it a very good note sharing tool. Being able to tell shared links apart is very useful, here's how I noticed the difference in the sharing services: 
image

Five snort links next to each other, and you don't know what you have sent someone x)

fiatjaf:
It only supports the basic text event types, for all others it displays a generic unfriendly page with event data.
But we want to support more kinds. What would be the next ones?

vitorpamplona:
It would be nice to have at least kind 1063, kind 40, kind 42, 30311, 34550, and 31922/31923/31924.

vitorpamplona:
But the goal of njump should be to render every single event. Even if the nevent is a share of a `kind:7` reaction, njump should find a way to render a view of the reaction itself and the original post.

fiatjaf:
Yes, that is the long-term goal, but it is moving slowly. We were fixing some bad architectural decisions first, we should be able to support more events easily from now on.

By the way, is there a way to have a "open in Amethyst" button that isn't just "open in native client"?

vitorpamplona:
> By the way, is there a way to have a "open in Amethyst" button that isn't just "open in native client"?

We could create a URI just for Amethyst. A copycat of `nostr:` but with `amethyst:`. Then as long as we are the only ones registering that URI with Android, the OS should pick it up.

greenart7c3:
You can do this way in your web app to open amethyst:

'''
href = `intent:nevent1qqsywt7cp095mpqvluljr3ue9f8ukscz45y5c4t0cx3e9gn6ukvg3wqppemhxue69uhkummn9ekx7mp0qgsr26r4lltjnvrwadxp67ns58m4qpzaqemhf5sup7hlujhjh7t296qrqsqqqqqpkefgt5#Intent;scheme=nostr;package=com.vitorpamplona.amethyst;end`;
'''

dtonon:
> You can do this way in your web app to open amethyst:
> 
> ```
> href = `intent:nevent1qqsywt7cp095mpqvluljr3ue9f8ukscz45y5c4t0cx3e9gn6ukvg3wqppemhxue69uhkummn9ekx7mp0qgsr26r4lltjnvrwadxp67ns58m4qpzaqemhf5sup7hlujhjh7t296qrqsqqqqqpkefgt5#Intent;scheme=nostr;package=com.vitorpamplona.amethyst;end`;
> ```

Interesting, thanks, I will do some tests on that!

dtonon:
@vitorpamplona 

> Does jump.me now supports most events we do?

As @fiatjaf said we (he!) just completed a complete rewrite, now we are going to integrate more kinds.
This week I will work on kind 1063 (file metadata), 1311 (live message) and 30311 (live event).

Do you have any stats about the sharing volume, split by kind? This way we can prioritize the work.

vitorpamplona:
Nice! Let's get this going! People are asking us to move to Primal as well. But I would prefer something more agnostic to client.

VictorieeMan:
> Nice! Let's get this going! People are asking us to move to Primal as well. But I would prefer something more agnostic to client.

Yes, it was the client agnosticism approach of njump.me that I thought was one of its strongest selling points.

VictorieeMan:
> Do you have any stats about the sharing volume, split by kind? This way we can prioritize the work.
@dtonon 

When I glanced through this code for creating this pull request, I didn't notice any telemetry for collecting such data. It looked like a clean add nostr note to an url function. The way I prefer it really, but yes that probably implies there isn't any data.

But ofc @vitorpamplona knows best and may chime in if my observation missed anything about this.

PR #686, opened 2023-11-10
Display amounts up to 4 digits without abbreviation

vicariousdrama:
When viewing reactions, it would be helpful to see up to 4 characters/digits of zap amounts. This changes the rounding to start at 10000 sats vs 1000, as most zap amounts are under that threshold.

See sample result from this change.  On the left is before the change. On the right is after. The value 1337 is shown instead of 1000, and 2100 instead of 2k among others.

2023-11-10-amethyst-before-and-after-4digit

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/note/ReactionsRow.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/note/ReactionsRow.kt
@@ -1438,13 +1438,14 @@ fun showCount(count: Int?): String {
     return when {
         count >= 1000000000 -> "${(count / 1000000000f).roundToInt()}G"
         count >= 1000000 -> "${(count / 1000000f).roundToInt()}M"
-        count >= 1000 -> "${(count / 1000f).roundToInt()}k"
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vitorpamplona/amethyst/ui/note/ReactionsRow.kt
```diff
@@ -1438,13 +1438,14 @@ fun showCount(count: Int?): String {
     return when {
         count >= 1000000000 -> "${(count / 1000000000f).roundToInt()}G"
         count >= 1000000 -> "${(count / 1000000f).roundToInt()}M"
-        count >= 1000 -> "${(count / 1000f).roundToInt()}k"
+        count >= 10000 -> "${(count / 1000f).roundToInt()}k"
```
vitorpamplona, line 1441: I think the Divisor here must be updated as well.

vicariousdrama: Can you elaborate?
Changing the divisor from a standard metric unit (1000) to match the size at which its being used (10000) would be erroneous and I'm not understanding what you mean otherwise.

PR #691, opened 2023-11-20
Fixes/improvements to external signer

greenart7c3:
- [x] Fix sending zaps
- [x] Send some default permissions on login (need a new version of Amber not released yet)
- [x] Save the external signer package name on login (need a new version of Amber not released yet)
- [x] Test all events
- [ ] Test in graphene os
- [x] Test old version of Amber and Amethyst
- [ ] Support for signing multiple events simultaneous
- [ ] Support for sending the account metadata

Some screenshots:

Screenshot_20231120_150154

Screenshot_20231120_150212


greenart7c3:
Everything seems to be working as expected.
The only problem I found is the toast messages, if i logout and login again the toast messages never shows again.
I think we can merge this and do Support for signing multiple events simultaneous and sending the account metadata in another pull request so this pull request doesnt get too big like the others

cc @vitorpamplona

vitorpamplona:
Makes sense. I will merge now. Do we need to way a new release of Amber before we release this one? Or is this code working with both versions?

greenart7c3:
From my tests it's working on both versions but I made a pre-release in case something goes wrong

PR #710, opened 2023-12-03
add a button to cashu preview to redeem on external wallet

believethehype:
Cashu preview would allow to 1) Melt to user's Ln address 2) (new) Deeplink to Cashu Wallet App on phone 3) copy cashu token

cashu

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/ui/components/CashuRedeem.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/ui/components/CashuRedeem.kt
@@ -175,58 +175,47 @@ fun CashuPreview(token: CashuToken, accountViewModel: AccountViewModel) {
                         LoadingAnimation()
                     }
 
-                    Spacer(modifier = StdHorzSpacer)
-
                     Text(
-                        "⚡",
+                        "⚡ Send to Zap Wallet",
                         color = Color.White,
                         fontSize = 16.sp
                     )
                 }
-
-                Spacer(modifier = StdHorzSpacer)
-                Button(
-                    onClick = {
-                        try {
-                            val intent = Intent(Intent.ACTION_VIEW, Uri.parse("cashu://$token"))
-                            intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK
-
-                            startActivity(context, intent, null)
-                        } catch (e: Exception) {
-                            accountViewModel.toast("Cashu", context.getString(R.string.cashu_no_wallet_found))
-                        }
-                    },
-                    shape = QuoteBorder,
-                    colors = ButtonDefaults.buttonColors(
-                        containerColor = MaterialTheme.colorScheme.primary
-                    )
-                ) {
-                    Spacer(modifier = StdHorzSpacer)
-                    Text("\uD83E\uDD5C", color = Color.White, fontSize = 16.sp)
-                }
-                Spacer(modifier = StdHorzSpacer)
-
-                Button(
-                    onClick = {
-                        if (useWebService) {
-                            // In case we want to use the cashu.me webservice
-                            val url = "https://redeem.cashu.me?token=$token&lightning=$lud16&autopay=false"
-                            val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-                            startActivity(context, intent, null)
-                        } else {
-                            // Copying the token to clipboard for now
-                            clipboardManager.setText(AnnotatedString(token.token))
-                        }
-                    },
-                    shape = QuoteBorder,
-                    colors = ButtonDefaults.buttonColors(
-                        containerColor = MaterialTheme.colorScheme.primary
-                    )
-                ) {
-                    Text("⎘", color = Color.White, fontSize = 18.sp)
-                }
-                Spacer(modifier = StdHorzSpacer)
             }
+
+            Spacer(modifier = StdHorzSpacer)
+            Button(
+                onClick = {
+                    try {
+                        val intent = Intent(Intent.ACTION_VIEW, Uri.parse("cashu://$token"))
+                        intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK
+
+                        startActivity(context, intent, null)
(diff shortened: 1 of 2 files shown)


vitorpamplona:
We can have the buttons in vertical, with text:
1. Redeem to Lightning
2. Redeem to Cashu Wallet
3. Copy token 

Or we can keep the original Redeem button, but it then opens a list popup with options. That would also help you describe what the option is doing. 

You can use the same popup component for the upload server chooser.

believethehype:
New Layout suggestion:
Screenshot 2023-12-04 at 16 18 15

swedishfrenchpress:
Hey guys,

I'm Erik, a UX designer working with Argon on the eNuts project. Argon mentioned you're tackling a refactor and I wanted to jump in with some design thoughts, hear what you guys think, and also float an idea about a new feature – displaying the token as QR code?

I'm really digging @believethehype design – Making the buttons clearer with text is a huge improvement. One small suggestion: What if we made all the buttons the same width? I think it would would tie the box together a bit more.

I've been mulling over this QR code display feature. Seems like it could be a useful add-on for users who juggle their ecash wallet across devices. I’ve sketched out a rough concept – keen to get your take on it.
https://github.com/vitorpamplona/amethyst/assets/78821053/fb55730a-1dfd-4aa5-8714-ed4c37e36178

I also want to share some feedback from Saunter (Alby UX Designer).
image

Lastly I have couple of questions and an idea I wanted to share.

1. Zap wallet – What exactly is a zap wallet? Is it something the user sets up in the Amethyst app? Or is it just the lightning address already connected to their nostr account?
2. Ran into an error when I tried redeeming a token that's already claimed. Is the error message that Amethyst displaying when trying to redeem an already spent token? Or this is related to the fact that I don't have a zap wallet setup.
3. Lastly a thought: What if we streamline this just one 'Redeem' button? Users tap it and then pick from options like 'Send to Zap Wallet', 'Open in Cashu Wallet', 'Copy', or 'Display QR'. In settings, we let them pick a default redemption method? 

Looking forward to hearing what you guys think.

vitorpamplona:
I like it. 

> On the QR, 

What's the goal here? I am not sure what they will do with the QR. This is already on their phone, there isn't another phone to scan it. I assume most people will never use the QR.  

> What exactly is a zap wallet?

This one is sort of ambiguous. There are generally two setups users mix: 
- The receiving lightning wallet (most likely Alby, WoS or Mutiny)
- The sending lightning wallet (connected with Nostr Wallet Connect etc). 

Even though they are different setups, users tend to mix them into a "Zap wallet". The button sends it into the user's receiving wallet. 

> Ran into an error when I tried redeeming a token that's already claimed.

I am not sure if there is a way to check if the redemption status before redeeming itself. Maybe it's not even safe. I am not sure. 

> What if we streamline this with just one 'Redeem' button?

We can do this as well.

swedishfrenchpress:
> I like it.
> 
> > On the QR,
> 
> What's the goal here? I am not sure what they will do with the QR. This is already on their phone, there isn't another phone to scan it. I assume most people will never use the QR.
> 
> > What exactly is a zap wallet?
> 
> This one is sort of ambiguous. There are generally two setups users mix:
> 
> * The receiving lightning wallet (most likely Alby, WoS or Mutiny)
> * The sending lightning wallet (connected with Nostr Wallet Connect etc).
> 
> Even though they are different setups, users tend to mix them into a "Zap wallet". The button sends it into the user's receiving wallet.
> 
> > Ran into an error when I tried redeeming a token that's already claimed.
> 
> I am not sure if there is a way to check if the redemption status before redeeming itself. Maybe it's not even safe. I am not sure.
> 
> > What if we streamline this with just one 'Redeem' button?
> 
> We can do this as well.

Regarding the QR code feature, it was conceived with the possibility that a user might want to redeem it on a different phone. However, let's consider pausing it for the time being, as this idea is currently speculative.

Regarding NWC/NIP-47 (which I just looked up). We could develop a solution that caters to two scenarios: 

1. the user has not set up a receiving wallet.
2. the user has already set up a receiving wallet. 

I came up with a new idea to introduce a single 'Redeem' button for simplicity, as having three options can be be a big cognitive load for the user. After tapping 'Redeem', users are presented with different wallet that they have currently installed on their phone. If they have already set up a NWC/NIP-47 wallet via Amethyst, we could display this as 'Amethyst' to minimize confusion (I'm open to feedback on this). Additionally, allowing users to select a default wallet could prevent them from interacting with the modal repeatedly.  

Prototype Example: 
https://github.com/vitorpamplona/amethyst/assets/78821053/a21fc53d-f192-4912-b444-04e5c492a4f2

vitorpamplona:
How about something like this? 

Screenshot 2023-12-06 at 4 40 48 PM

swedishfrenchpress:
I'm unclear what the middle button's function is - does it access the list of ecash / ln wallets? Assuming it does, here's my thought:

Adding a third button might not be ideal. For instance, if a user opts to redeem using the NWC wallet, the existing single redeem button suffices. However, if they prefer using a cashu wallet, they would have to repeatedly select a wallet by pressing the second button. This setup doesn't cater to users who wish to redeem with an external wallet in a single tap. Overall: a single redeem button is a better direction imo.

What do you think about preparing some A/B designs and asking the nostr community what they think?

vitorpamplona:
Now with light and dark modes for the chat and regular feed. 

|                              |                                      |                                 |                                          |
|-------------------------------------------|----------------------------------------------|-------------------------------------------------|--------------------------------------------------------|
| Screenshot_20231206_183446 | Screenshot_20231206_183421 | Screenshot_20231206_183314 | Screenshot_20231206_183058 |

vitorpamplona:
> I'm unclear what the middle button's function is

The middle button is supposed to be the default icon for taking you out of this app into something else. 

The issue with your idea for a single button is that I cannot add Amethyst in Android's popup list unless I get amethyst to respond to all `cashu:` links, which I don't think is ideal (it's not a cashu wallet).

swedishfrenchpress:
Got it - it makes sense. Do you think it would be possible to have a setting option where a user can select which wallet they want to redeem cashu tokens with by default?

vitorpamplona:
Between the ln and the cashu option or just among the popup to choose a wallet when you press the ↗️ button? 

Android does keep a default option over there.

PR #749, opened 2024-01-22
save a draft while you are typing the post

greenart7c3:
This only saves the text for now.
We can try to add the note options like pools later.
This also doesn't save the note you are replying to so if the app crashed while you are replying some note you have to go back to that note.
Is this good for now?

cc @vitorpamplona @vicariousdrama @alltheseas


vicariousdrama:
This seems like it should work fine for notes not done as a reply, though I do have a new concern.. What if something about the message text itself would cause a crash (e.g. rendering an image preview of a referenced url). Would a user be able to clear that by any means apart from wiping the data cache?

greenart7c3:
I can show a pop up before opening where you can edit the note or discard or add an option in the drawer menu. I'll try to see if I can get the reply too so you don't have to find the note and press reply again

greenart7c3:
I made a option in the drawer menu when you have a draft that only shows the text of the draft and im also saving the note you are replying to

vitorpamplona:
This seems quite limited. It just works for 1 new text and 1 reply and it doesn't save which reply that was for. It also doesn't save any of the tags or added components to the text. It won't work for chats. 

I kinda wanted to find a way to save the new post as an actual event (of any kind the user is creating) and save a list of those events locally only. Then in the interface, we could have a feed for drafts and drafted posts should show up as if they were real, but with a different color or sign and a button to edit or send right there. 

In chat, for instance, the saved draft could be an actual message bubble at the bottom of the chat. the user can send, edit or delete the draft in the chat screen. 

Same for replies. It shows in the thread with action buttons to delete, edit or send. 

For new posts, they could show as a new list you can select on the top bar. 

All of these require saving the actual event locally. Maybe we can save them in the usual Preferences file, but I think this might need the full DB.

greenart7c3:
Yeah I just did some basic draft to at least save something when the user is posting something to the main feed.
I think a DB will work better if we are going to keep multiple drafts and all the options

davotoula:
Sometimes starting small can be benefitial as long as expanding on the feature doesn't require a rewrite...

quartz/src/main/java/com/vitorpamplona/quartz/events/DraftEvent.kt
```diff
@@ -0,0 +1,58 @@
...
+        createdAt: Long = TimeUtils.now(),
+        onReady: (DraftEvent) -> Unit,
+    ) {
+        val tags =
+            arrayOf(
+                arrayOf("d", dTag),
+                arrayOf("k", "${originalNote.kind()}"),
+            )
+
+        signer.nip44Encrypt(originalNote.content(), signer.pubKey) {
```
vitorpamplona, line 54: I was thinking on creating the event and encrypting the entire JSON of the event as opposed to just the content. In that way, we don't lose zap splits or imeta tags that were previously arranged.

vitorpamplona:
Drafted a quick NIP to get people's reactions to this new event kind: https://github.com/nostr-protocol/nips/pull/1124

greenart7c3:
Here's is how the ui is for now.
Still missing the edit on dm's and adding some validations for the user not comment, reply react to drafts.
I also need to find a good way to send only the last edit because im getting rate limited by relays if i send every time i type.
cc @vitorpamplona 

Screenshot_20240318_163020
Screenshot_20240318_162957
Screenshot_20240318_162925
Screenshot_20240318_163113

vitorpamplona:
Nice! The SearchScreen.SearchBar compose has a Channel that uses a 300ms debounce to search only 300ms after the user stops typing. You can actually do a 1 or 2-second debounce to avoid singing and sending all the time.

vitorpamplona:
PS, do you like the new simplified mode?

greenart7c3:
Yes I'm liking the simplified mode

vitorpamplona:
How is this going? We are ready to merge and release? I will release a version this week

PR #750, opened 2024-01-22
remove the need for application package name in the androidManifest for the external signer

greenart7c3:


vitorpamplona:
This is awesome! Thanks for pushing this @greenart7c3

PR #796, opened 2024-03-04
Accessibility updates

davotoula:
Adding a few missing content descriptions for a better accessibility experience.


vitorpamplona:
Nice! Thank you @davotoula

PR #803, opened 2024-03-08
Use nostr.com instead of njump.me when copying link

greensheeps:


vitorpamplona:
Humm... Isn't it the same thing?

vitorpamplona:
There are some rendering variations between the njump and nostr.com. I will keep njump for now, but we can change it later if nostr.com gets better rendering.

PR #812, opened 2024-03-20
add a lifeCycleOwner to register external signer on resume

### LINKED ISSUE #811: [BUG] Amber stop working when device is idle for a long time
**Describe the bug**
Amber stops opening after a while

**To Reproduce**
When using amethyst with amber it stops opening amber after a while. I think it's when the device enters deep sleep.

**Expected behavior**

**Video and Screenshots**

greenart7c3:
fixes #811


vitorpamplona:
Nice!

PR #814, opened 2024-03-23
Fix NIP-44 encryptDecryptTest to decrypt with swapped private and public keys to follow NIP-44 documentation

tyiu:
See https://github.com/nostr-protocol/nips/blob/4b79bc67c471f77061d62704538e5fdd6ac28ae8/44.md?plain=1#L291

Also, I made the equivalent change here:
https://github.com/paulmillr/nip44/pull/10


vitorpamplona:
Nice! good catch!

PR #818, opened 2024-03-26
Fix garbled URL preview for non UTF-8 HTML, with optimization

jiftechnify:
## Problem
`UrlPreviewUtils` doesn't take character encodings into account when getting OGP metadata from HTMLs. According to [WHATWG encoding standard](https://encoding.spec.whatwg.org/#preface), authors of HTML are required to use UTF-8 as its encoding, so Amethyst's decision today is very reasonable. Though, unfortunately, considerable amounts of websites in Japanese (also very famous ones) still use legacy charsets. For example:

- https://www.itmedia.co.jp/ ... uses SHIFT_JIS
- https://www.4gamer.net/ ... uses EUC-JP

Today if you share such websites on Amethyst, URL previews are sadly garbled:

Screenshot_20240327-011602

## Fix
This PR fixes this by detecting character encodings of HTMLs from following information:

- `charset` parameter in `Content-Type` header 
- Byte order mark in UTF-8/UTF-16/UTF-32 texts
- `<meta charset="...">`
- `<meta http-equiv="content-type" content="text/html; charset=...">`

The issue is fixed as below:

Screenshot_20240324-195319

## Optimization
This fix requires parsing a HTML twice most of time (when neither `Content-Type` header nor BOM give us any information).

- 1st pass: tentatively read the HTML as UTF-8 to detect actual encoding
- 2nd pass: read the HTML using the encoding detected in 1st pass to extract metadata

Moreover, it seems that jsoup donsn't support partial HTML parsing, so we were forced to fully parse a HTML, despite all the data required to show URL previews are only in `<meta>` tags, typically sit in the very beginning of HTMLs.

To resolve the unefficiency, I decided to implement custom parser for `<meta>` tags that read a HTML as little as possible. According to a micro-benchmark, metadata extraction based on the custom parser is 12x faster than jsoup-based method (take it with a grain of salt since I ran it on my development PC, not on an Android device...).

```
 main: org.example.ParseOGPBenchmark.customParser

Warm-up 1: 1992.640 ops/s
Warm-up 2: 2109.442 ops/s
Warm-up 3: 2127.839 ops/s
Warm-up 4: 2092.212 ops/s
Warm-up 5: 2104.461 ops/s
Iteration 1: 2116.675 ops/s
Iteration 2: 2114.388 ops/s
Iteration 3: 2131.505 ops/s
Iteration 4: 2127.793 ops/s
Iteration 5: 2119.129 ops/s

2121.898 ±(99.9%) 28.452 ops/s [Average]
  (min, avg, max) = (2114.388, 2121.898, 2131.505), stdev = 7.389
  CI (99.9%): [2093.446, 2150.350] (assumes normal distribution)

main: org.example.ParseOGPBenchmark.jsoup

Warm-up 1: 163.284 ops/s
Warm-up 2: 172.806 ops/s
Warm-up 3: 172.232 ops/s
Warm-up 4: 172.863 ops/s
Warm-up 5: 174.018 ops/s
Iteration 1: 174.190 ops/s
Iteration 2: 171.965 ops/s
Iteration 3: 171.827 ops/s
Iteration 4: 170.970 ops/s
Iteration 5: 172.285 ops/s

172.247 ±(99.9%) 4.583 ops/s [Average]
  (min, avg, max) = (170.970, 172.247, 174.190), stdev = 1.190
  CI (99.9%): [167.665, 176.830] (assumes normal distribution)


main summary:
Benchmark                        Mode  Cnt     Score    Error  Units
ParseOGPBenchmark.customParser  thrpt    5  2121.898 ± 28.452  ops/s
ParseOGPBenchmark.jsoup         thrpt    5   172.247 ±  4.583  ops/s
```


vitorpamplona:
This is super nice!  Thank you @jiftechnify

vitorpamplona:
Any chance you can add your benchmark to the project as well? We can move the parser to `commons` to give access to the benchmark module

jiftechnify:
@vitorpamplona Sure I'll make it up, on another PR:+1:

Issue #843, opened 2024-04-20
[BUG] URL previews mishandle HTML entities


strout:
**Describe the bug**
URL previews (image/title/description boxes that replace some links) show HTML entities such as "\&amp;" instead of the characters they represent.

**To Reproduce**
1. View a note with a preview, such as nostr:nevent1qqsf4ysx2ha8uux4hqylt2l2lgg0lkhq8r9g6zqfswug2f7r4yze26spz3mhxue69uhhyetvv9ujuerpd46hxtnfdupzpqt3frpkjq24gqd5j3vqsu0mq4j2t740h9z5syl0990jwp4ujv6eqvzqqqqqqydhnpmu
2. See the HTML entities visible instead of characters like quotation marks.

**Expected behavior**
HTML entities should be replaced by the characters they represent.

**Video and Screenshots**
![Screenshot_20240419_211603_Amethyst](https://github.com/vitorpamplona/amethyst/assets/110026/c65cef0d-d3dc-464d-a07b-1e459a772828)

davotoula:
I have provided a manual solution to decode most common html entities. It doesn't cover all html entities.

Please try again and feel free to close this issue if it is sufficent.

thank you for reporting it!

vitorpamplona:
Fixed: the URL-preview meta-tag parser now resolves HTML character references (`&amp;amp;`, `&amp;#39;`, `&amp;#8217;`, `&amp;hellip;`, …) in title/description attributes, so previews show the real characters. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #845, opened 2024-04-23
843 url previews mishandle html entities

davotoula:
added test to regular expression for html entity numbers
added a few common numerical html entities

#843 
@jiftechnify

### DIFF
--- a/commons/src/main/java/com/vitorpamplona/amethyst/commons/preview/MetaTagsParser.kt
+++ b/commons/src/main/java/com/vitorpamplona/amethyst/commons/preview/MetaTagsParser.kt
@@ -130,7 +130,7 @@ object MetaTagsParser {
     // - commonly used character references in attribute values are resolved
     private class Attrs {
         companion object {
-            val RE_CHAR_REF = Regex("""&(\w+)(;?)""")
+            val RE_CHAR_REF = Regex("""&(#?)(\w+)(;?)""")
             val BASE_CHAR_REFS =
                 mapOf(
                     "amp" to "&",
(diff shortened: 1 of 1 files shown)


vitorpamplona:
hum... we might need to figure out a way to include all codes in this.. 

&nbsp; can also be written as &#160; (decimal) or &#xa0; (hex)

davotoula:
@vitorpamplona Would you consider using a 3rd party library?

davotoula:
Actually 2 questions..

Thorough and complete html parsing is a project on it's own. Could we leverage an external open source library? 

It seems as if the whole external html document is parsed. We are just showing title and few lines from the document. Maybe it's possible to only parse part of the external html document?

vitorpamplona:
> Could we leverage an external open source library?

Sure, frankly most of these HTML decoders should be inside Java/Kotlin libraries already, I just don't know where to find them or how to use the internal library together with @jiftechnify's fast parser. 

> Maybe it's possible to only parse part of the external html document?

I think these opengraph tags could be anywhere in the head of the document, so we might discard the `body`. But I am frankly not sure about it.

jiftechnify:
I made a misassumption about HTML entity references... I didn't think they are used so wildly in the world. As a consequence, the `MetaTagParser` I introduced in #818 only resolves few basic HTML entity refs:

https://github.com/vitorpamplona/amethyst/blob/main/commons/src/main/java/com/vitorpamplona/amethyst/commons/preview/MetaTagsParser.kt#L133-L166

If there is a library that do the one thing(resolving HTML entity refs), we could leverage it in my parser. I'll look for it.


> It seems as if the whole external html document is parsed. We are just showing title and few lines from the document. Maybe it's possible to only parse part of the external html document?

The `MetaTagsParser` already do this; it only parses `<meta>` tags inside `<head>` tags.

davotoula:
@jiftechnify 
That sounds like the right approach... leveraging an existing library!

Maybe this one would work?

https://github.com/MohamedRejeb/Ksoup

vitorpamplona:
> https://github.com/MohamedRejeb/Ksoup

I am not sure if we should use the entire library, since it might be slower than what @jiftechnify has done. But Ksoup seems to have an entity map: https://github.com/MohamedRejeb/Ksoup/blob/main/ksoup-entities/src/commonMain/kotlin/com/mohamedrejeb/ksoup/entities/EntityMaps.kt

davotoula:
I don't see mapping from hex entities...

vitorpamplona:
> I don't see mapping from hex entities...

Their Numberic Decoder does that before maps are applied: https://github.com/MohamedRejeb/Ksoup/blob/0ac02fe52d500605f9e068da70de582cbe1fa3e5/ksoup-entities/src/commonMain/kotlin/com/mohamedrejeb/ksoup/entities/text/translate/NumericEntityDecoder.kt#L70

Issue #836, opened 2024-04-13
[BUG] App crashes while adding words to mute filter


tohrxyz:
**Describe the bug**
When adding new words to mute filter, and submitting each one, the app crashes.

**To Reproduce**
1. Go to Security Filters
2. Go to Hidden words
3. type a new word
4. click 'add'
5. decent chance it crashes

**Expected behavior**
shouldn't crash

**Device (please complete the following information):**
 - Phone Brand/Model: Pixel 7a
 - Android Version 14 
 - App Version: v0.86.3-FDROID

PR #851, opened 2024-05-04
show toast error if unable to hide words

VASHvic:
I was investigating this issue : https://github.com/vitorpamplona/amethyst/issues/836
I was unable to make the app crash like the user describes, but i realized i could't add hide words if i was using a public key and i wasn't getting notified either. 
Anyway i'm happy if i can contribute and learn some Kotlin in the process, i hope the PR has the desired behaviour.


vitorpamplona:
Nice! Thank you, @VASHvic

Issue #852, opened 2024-05-09
[FEATURE] Split Zap Progress Into 2 Phases


djschnei21:
**Is your feature request related to a problem? Please describe.**
When using a NWC enabled wallet which is not always online (e.g. Mutiny Wallet on Android), the one tap zap experience is sub-optimal.  Because the zap progress circle tracks the full request in 1 "stage" (request zap and zap receipt), the progress circle ends up hanging (since Mutiny isn't open in the foreground).  When you scroll away from the (attempted) zapped note, that half-filled circle disappears, making it very hard to track the things I have already attempted to zap.  This result in frequent accidental double zaps and an unclear UX.  

**Describe the solution you'd like**
A solution to this would be to replace the zap progress circle with zap icons of differing colors to denote the various stages of a zap request.  One color could mean "you've requested a zap" and another could mean "zap receipt received/zap complete".  This would allow the user to "zap" a bunch of notes in one session, open Mutiny so that it processes the requests, then continue back in amethyst without ever losing track of which notes they've attempted to zap.  

**Bounty (in Bitcoin sats) offered for the implementation**
20,000 sat bounty

PR #853, opened 2024-05-10
replace progress circle for zaps with outbox icon

djschnei21:
taking a stab at #852


vitorpamplona:
Keep in mind that zap splits are becoming more common and because there could be so many receivers (like 50+ people) the zap progress can become really slow. It might take a whole minute to get all invoices ready for Mutiny. So, just shifting the icon immediately as it starts would leave users without any knowledge of what's happening. 

If we want to use a different icon for each stage, I think that icon should also have its own progress bar. Because each stage might take a moment.

PR #854, opened 2024-05-10
allow relay selection dialog to pick any relays

jeremyd:
Currently, the relay selection dialog only shows "write" relays.  This change allows you to select from any of the relays and override the relay posting engine to allow the messages through.

### DIFF
--- a/app/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/app/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -2435,6 +2435,10 @@ class Account(
         return (activeRelays() ?: convertLocalRelays()).filter { it.write }
     }
 
+    fun activeAllRelays(): List<Relay> {
+        return ((activeRelays() ?: convertLocalRelays()).toList())
+    }
+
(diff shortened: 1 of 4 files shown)


jeremyd:
currently still doing some testing, I think this is working, but auth seems broken in main branch?

Also trying to double-check it will not post to nostr.band

jeremyd:
update:  looks like auth works fine, just the relay icons on the note didn't show up (but it did post and another account shows the icon etc).

still need to confirm there is no leakage to nostr.band (tough to tell because of all these relay scrapers these days)..

more soon.

jeremyd:
confirmed, there does not appear to be any leakage.  Just was getting confused from search scrapers like nostr.band grabbing messages via nostr.watch relay lists.

So yeah, this is working!!  Very cool.  Any feedback @vitorpamplona ?

vitorpamplona:
Sorry, I was not able to check this yesterday, but I skimmed over it and it looks great. I will try to merge by Monday.

vitorpamplona:
@jeremyd what's your npub?

PR #916, opened 2024-06-14
Fix hidden notes when hidden words is empty

greenart7c3:
Don't know if we should check this here or change the containsAny function


vitorpamplona:
good catch!

Issue #915, opened 2024-06-14
[BUG] NIP05 verification follows redirects against specification


davotoula:
**Describe the bug**
Currently nip05 verification will follow a redirect for nip05 verification.

From the spec [NIP05 specification](https://github.com/nostr-protocol/nips/blob/master/05.md):

> Security Constraints
> The /.well-known/nostr.json endpoint MUST NOT return any HTTP redirects.
> 
> Fetchers MUST ignore any HTTP redirects given by the /.well-known/nostr.json endpoint.
> 

**Expected behaviour**
NIP05 verification should fail if it encounters a 3xx redirect

**Note**
HTTP to HTTPS redirect should be considered to be allowed but looking at the code HTTPS is used by default. FollowRedirects OFF will likely turn off protocol redirects as well

davotoula:
NIP05 verification should use a httpclient with followRedirects OFF.

Looking at the code there are currently 2 httpclient singletons: with and without proxy.

Am I correct assuming we would need 2 more so in total there would be 4 singletons: httpclient, httpclientwithoutproxy, httpclientNoRedirect, httpclientwithoutproxyNoRedirect?

Or is this overcomplicating it?

vitorpamplona:
You can use newBuilder() directly on the NIP-05 verification fetch 

```kt
            HttpClientManager.getHttpClient()
                .newBuilder()
                .followRedirects(false)
                .followSslRedirects(false)
                .build()
```

davotoula:
Thats nice... I'm worried about the extensive usage of nip05 verification. This will create a lot of new httpclients?

davotoula:
or possibly will update the singleton to not use redirects which would be bad when fetching regular nostr content?

davotoula:
fixed

PR #920, opened 2024-06-15
Set redirects to false for httpclient used during nip05

davotoula:
Change OkHttpClient used for NIP05 verification to not follow redirects.

I verified that this change fails NIP05 verifications with a redirect.

After reading the OkHttpClient docs I'm comfortable this will not create unnecessary resources.

> Customize Your Client With newBuilder()
> You can customize a shared OkHttpClient instance with newBuilder. This builds a client that shares the same connection pool, thread pools, and configuration. Use the builder methods to add configuration to the derived client for a specific purpose.

Fix for https://github.com/vitorpamplona/amethyst/issues/915


vitorpamplona:
Thank you @davotoula

PR #939, opened 2024-06-24
Moves relay classes to a new package

greenart7c3:


vitorpamplona:
Nice!

vitorpamplona:
Data Sources as well? Aren't they tightly coupled with Amethyst's UI? I never thought anyone else could use them.

vitorpamplona:
ohh just the base class. That makes sense.

app/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
```diff
@@ -75,7 +75,7 @@ private object PrefKeys {
     const val SAVED_ACCOUNTS = "all_saved_accounts"
     const val NOSTR_PRIVKEY = "nostr_privkey"
     const val NOSTR_PUBKEY = "nostr_pubkey"
-    const val RELAYS = "relays"
+    const val RELAYS = "com/vitorpamplona/ammolite/relays/relays"
```
vitorpamplona, line 78: check this

greenart7c3:
> ohh just the base class. That makes sense.

Yes, i also added a few functions to the listeners classes that I use in Amber
And fixed a bug when using the sendToSelectedRelays function and the relay is not ready

vitorpamplona:
Very cool!

davotoula:
Is this related to a feature or an issue?

What's the context for this change?

Thanks!

vitorpamplona:
It was a long requested feature to break the base layers of the app into multiple libraries. We can keep pushing things into libraries to help others develop. 

In this case, I think @greenart7c3 needed some of the relay classes for Amber/Citrine.

PR #941, opened 2024-06-24
changes app directory to amethyst

vitorpamplona:
@greenart7c3 @davotoula 

this just renames the "app" directory to "amethyst". Since we have so many modules now, I think it is better to be more specific about which "app" it is. Thoughts?


greenart7c3:
Nice, I think amethyst is better than app

PR #957, opened 2024-06-29
Profile Gallery (List based version)

believethehype:
Reasoning:

Nostr replaces many different platforms, but hasn't been focusing a lot on visual social media. By adding user galleries, users might add media to a grid so they can showcase some of their content/work to others. This has especially artists, photographers and users of platforms like instagram in mind. 

Current Status: 

User's can create a Kind 10011 List. (This is not added in nip51 yet, but will be suggested once we are happy with a format)

The current format of tags in that list is
 ["gallery", "eventid", "urlofmediatoshowingallery"]
(we might need something to differentiate between regular event ids and addressable ones, open for discussion)

The list contains both the id and the url as a user might want to pin the 2nd or 3rd image in a note, and we need the id so we can send users to the original note when clicking the gallery for comments, descriptions etc)

Users in Amethyst have a new Tab in Profile where they can decocrate a "gallery".
When opening media in fullscreen users can use the share button to add media to their gallery
When long pressing media in the gallery they can remove it. (this might take a couple of seconds)

You can of course visit other peoples profiles to see what they share in their Gallery.

Remaining issues:

- When going back from a note to the gallery, it doesn't remember scroll position and users have to scroll down to the gallery again
- Is the current 10011 format sufficient or do we need to distinguish tags for event ids and addressable events?

Update:
Videos now also should work in the Gallery


believethehype:
"When going back from a note to the gallery, it doesn't remember scroll position and users have to scroll down to the gallery again"

Just realised this happens on all feeds in the profile, also the notes and replies. Opened an issue  #963

believethehype:
This is working right now, but I'm converting it to a draft again until there's agreement in https://github.com/nostr-protocol/nips/pull/1340 on how the tags should be named, before it gets merged.

davotoula:
Trying out your pull request. Is there an entry point from the app ui into your new functionality?

believethehype:
https://v.nostr.build/7G2me.mp4

davotoula:
Omg so obvious, I can't believe I missed it! 😂Thanks

believethehype:
I think #973 is the better way to move forward, given the limitations of lists, feel free to close

PR #973, opened 2024-07-05
Profile Gallery

believethehype:
Here is an alternative attempt at profile galleries. This is an alternative to #957 and only one of both PRs should be used.

The main difference:

This uses kind 1163 (for now) Events per Gallery Entry (Similar to 1063 but with e tag to an event with fileheader info only optional) instead of kind 10011 lists. Here the events represent gallery entries and an infinite amount can exist, where for lists  there might be limited by size and have a potential chance to get wiped. 

The current state of the PR works in general (add, delete, show) ~~but it seems it's sometimes having issues to load single events (especially older ones). Not sure if thats a limitation of the filter or if there's some easy way to fix this (Didn't seem to happen for lists)~~

One remaining current limitation is that when a note gets deleted and a gallery entry references to that note, Amethyst will not know and show a loading icon with currently no possiblity to remove the entry


vitorpamplona:
So, is this one ready to go? Or are we waiting for the NIP repo to move merge?

believethehype:
I had withdrawn my previous PR on the NIPs repo as this was the approach with lists (NIP51) .
I think this one is the better approach. 

I haven't made a new PR since. Basically what we need is to reserve the 1163 (or whatever we decide) Kind with an url tag and an e tag.

I don't know if this should be part of an existing NIP or a new NIP (it's a pretty simple addition, yet it should be written somewhere sooner or later). 

In terms of Code, I believe this is ready to go, except you have any suggestions :)

vitorpamplona:
Nice! 

> Basically what we need is to reserve the 1163 (or whatever we decide) Kind with an url tag and an e tag.

That would be a small nip just to define the kind and tags

believethehype:
I made a new PR on the nips repo.

https://github.com/nostr-protocol/nips/pull/1356

PR #974, opened 2024-07-06
Fix poll options behavior

### LINKED ISSUE #532: [BUG] Deleting "not last" poll options in note breaks creation of poll
**Describe the bug**
Deleting a poll option that is not last in the list causes issues.

**To Reproduce**
1. Click to create a new note
2. Add two or more poll options
3. Delete poll options in the middle of the list

**Expected behavior**
Poll list can still be edited by adding new items or removing items.

**Actual behaviour**
1. Unable to add additional poll options
2. Unable to delete additional poll options
3. Bit random in terms what can be added / removed aftwards

 - Phone Brand/Model: Pixel 7 Pro
 - Android Version: 33  
 - App Version: v0.71.0

KotlinGeekDev:
Fixes #532 .
As observed by both @davotoula and I, the poll state becomes inconsistent when an item, which is not the last item, is deleted, and persists even when adding items back. This PR aims to solve that issue.


KotlinGeekDev:
cc @davotoula

PR #975, opened 2024-07-07
Some longform notes (or longform DVM feeds) crashed Amethyst

believethehype:


vitorpamplona:
Hum... this is weird. Why was it crashing? was the `LocalConfiguration.current.locales.get(0)` null?

believethehype:
It gave me java.util.Formatter: UnknownFormatConversionException: Conversion = 'F in said line when loading lastest longform notes DVM yeterday. This also happend a while ago, until it didn't.  I'm not sure but I have the impression it has to do with *some* longform notes.

vitorpamplona:
Usually, that means you are loading a string from strings.xml with a `%1$s` that requires a parameter, and the parameter is not being passed or being passed in a weird type.

believethehype:
Yes, i think that is what was happening. An empty or weird type. The explicit .toString() cast avoided the crash when it happened yesterday.

vitorpamplona:
Yeah, but it isn't better to solve which string was used without a param? It could also be that one of the translations removed or misspelled the `%1$s`, removing the param that otherwise would be there.

vitorpamplona:
Fixed on https://github.com/vitorpamplona/amethyst/commit/5a60c3a595869fc9b85ccc6655e3222b30cdfd19

PR #992, opened 2024-07-22
Remove focus from status text field when closing the drawer

greenart7c3:
[Screen_recording_20240722_145158.webm](https://github.com/user-attachments/assets/d8ae5422-3eba-4443-8d54-1b2c88a76537)


vitorpamplona:
nice find!

Issue #996, opened 2024-07-26
[BUG] Copy text on an edited post copies the original text to clipboard 


davotoula:
**Describe the bug**
Copy text on an edited post copies the original text to clipboard instead of the (new) edited text

**To Reproduce**
1. Create a new post
2. Edit the post and add new text
3. Copy text
4. Copied text is the original note text and not the new edited text

**Expected behavior**
Copied text should be the new edited content

**Device (please complete the following information):**
 - Phone Brand/Model Pixel 7 Pro
 - Android Version 34
 - App Version 89.7 google play

davotoula:
Works. Closing

PR #997, opened 2024-07-27
Fix copying to clipboard edited notes

VASHvic:
Hello! found an easy issue i could contribute again, hope the fix its alright
https://github.com/vitorpamplona/amethyst/issues/996


vitorpamplona:
Thank you @VASHvic!

PR #999, opened 2024-07-29
Show mint in cashu token rendering

believethehype:
Adds the mint to the rendering of the token so recipent knows where to receive the token from.
asd


believethehype:
also includes a minor fix to order of Share ImageAction

PR #1001, opened 2024-07-31
Gallery: use FileHeader Info for rendering / faster loading

believethehype:
Instead of AsyncImage use GalleryContentView that considers fileheader info (if available) such as blurhash


believethehype:
I also reworked the code for how preview images are loaded by loading the associated note way later during the process. This leads to the gallery rendering way faster than before.

Issue #993, opened 2024-07-23
[BUG] Unable to add media to gallery when "Show more" has been clicked


davotoula:
**Describe the bug**
For notes that are large enough to produce a "Show more" icon, when the "Show more" is clicked it is not possible to add media to gallery.

**To Reproduce**
1. Go to:  nostr:nevent1qqsrec6u7pxf7a80f8h374m2xmnlr5up3edw5k2jh0w4fk70xfcer2qppemhxue69uhkummn9ekx7mp0qgsv4zwtz8cuwh2mvc3zdrl5853g365t9j6mn25edlul7uz0eyzt0zcrqsqqqqqp0ghja3
2. Click "Show more"
3. Click on any image in the post
4. Only option available is "Copy URL"

**Expected behavior**
"Add media to Gallery" to be present. Note: If you click an image before clicking "Show more" you are able to add it to the gallery

**Video and Screenshots**
<img src="https://github.com/user-attachments/assets/d63ae5b8-7b5b-4035-b62b-1b449184d017" height="800"/>

**Device (please complete the following information):**
 - Phone Brand/Model Pixel 7 Pro
 - Android Version 34 
 - App Version 89.7

@believethehype

believethehype:
Interessting, Copy node id to clipboard is also missing in this Case.

git-3000:
You're right. I tested it and the save to gallery is missing.
I also realized from your post that I have been downloading images instead of saving it to my gallery.

The download button says saved to your gallery which I thought meant saved to my Nostr profile gallery

I guess they mean saved to my phone gallery.

![Uploading Screenshot_20240723-231316_Amethyst.png…]()

PR #1002, opened 2024-07-31
Bugfix:  add to gallery and copy note id to clipboard missing after clicking "show more"  #993

believethehype:
As pointed out in issue #993. when clicking the "show more" button on longer notes, media could not be added to the gallery (nor could the note id be copied). This fix adds the uri when ImageSegment is fetched from the Richtextviewer renderer


davotoula:
Nice one! I spent 3 days trying to locate where the uri is lost 😂

vitorpamplona:
ohhhh there is a bug on the caching of the RichTextParser! I am taking a look at this. This affects other places as well.

vitorpamplona:
https://github.com/vitorpamplona/amethyst/commit/231af1d3d84cc2fc8c2445d7974b9fc1da5ba8db fixes #993 and this PR.

PR #1012, opened 2024-08-12
show only 3 users in the recommended relays section

greenart7c3:
Removing the damus relay for example it would show 100+ users there and freeze the ui when scrolling

image


vitorpamplona:
Thank you @greenart7c3

PR #1018, opened 2024-08-16
Change tap to fullscreen video behavior for livestreams.

### LINKED ISSUE #539: [FEATURE] Fullscreen video on stream view portrait to landscape mode
When opening a live stream in portrait and then turning phone to landscape mode, I would like the video to be automatically in fullscreen as does NewPipe or other video apps.

Thank you :)

KotlinGeekDev:
The aim of the PR is to fix #539 , while taking into account other devices like foldables.


KotlinGeekDev:
Please @vitorpamplona I don't know if it's a good idea to include Composable functions in the Quartz module.

vitorpamplona:
Yeah, we should probably keep the utility function in the same package as the videoview

KotlinGeekDev:
> Yeah, we should probably keep the utility function in the same package as the videoview

Got it.

KotlinGeekDev:
@vitorpamplona The PR is all set and ready.

vitorpamplona:
Thank you @KotlinGeekDev

vitorpamplona:
@KotlinGeekDev we need a fix for portrait videos. The device should not go to landscape on fullscreen if the video is portrait. Many videos in the video tab are like that.

KotlinGeekDev:
@vitorpamplona does it mean we should just go to fullscreen only when the user willingly changes orientation?

vitorpamplona:
It should go full screen all the time, but only rotate to landscape when the video is in landscape. Otherwise, the portrait video shows in landscape an only uses a tiny space in the middle of the screen

KotlinGeekDev:
> It should go full screen all the time, but only rotate to landscape when the video is in landscape. Otherwise, the portrait video shows in landscape an only uses a tiny space in the middle of the screen

I see. So basically the same behsviour as before?

KotlinGeekDev:
> If the video is portrait, then yes, same behavior as before. If the video is in landscape, then it should change orientation

Got it.

KotlinGeekDev:
Please @vitorpamplona it therefore means disabling it as you did will enough to fix the issue you brought up, right?

vitorpamplona:
Disabling like that disables the entire PR, right? It won't rotate for landscape videos, when it should.

KotlinGeekDev:
> Disabling like that disables the entire PR, right? It won't rotate for landscape videos, when it should.

Interesting. From my experience testing the effects of disabling, it works even in landscape mode.

KotlinGeekDev:
My understanding of the original issue is that the author wanted the NewPipe app experience, whereby, on portrait devices, tapping the fullscreen button changes the orientation to landscape, then going fullscreen.
Please @anthony-robin correct me if I am wrong.

vitorpamplona:
> Interesting. From my experience testing the effects of disabling, it works even in landscape mode.

But how would it work if it is not calling any function to change orientation? In my device, it now doesn't change any orientation at all.

KotlinGeekDev:
> > Interesting. From my experience testing the effects of disabling, it works even in landscape mode.
> 
> But how would it work if it is not calling any function to change orientation? In my device, it now doesn't change any orientation at all.

I think that the way the fullscreen experience currently works(the way it worked before my PR) is that the orientation was manually set by the user in the system settings, which affected the app.

anthony-robin:
> @vitorpamplona My understanding of the original issue is that the author wanted the NewPipe app experience, whereby, on portrait devices, tapping the fullscreen button changes the orientation to landscape, then going fullscreen. Please @anthony-robin correct me if I am wrong.

Hi @KotlinGeekDev, the original issue was about a NewPipe experience (that might not be the only app that does this but the one I had in mind when opening the issue) by having an auto fullscreen video mode when I turn my phone in a landscape position on a "Live stream" video page. 

Also note that when the landscape fullscreen is turned back to portrait, fullscreen should be disabled back to the original rendering (if it is possible)

I didn't really thought about the portrait mode so I think it is better to keep it in fullscreen portrait with a good video rendering and not a tiny content with lot of black on horizontal sides.

Let me know if you have more questions or if things are not clear enough.
Anyway, thank you for working on the subject :)

KotlinGeekDev:
> 
> Hi @KotlinGeekDev, the original issue was about a NewPipe experience (that might not be the only app that does this but the one I had in mind when opening the issue) by having an auto fullscreen video mode when I turn my phone in a landscape position on a "Live stream" video page.
> 
> Also note that when the landscape fullscreen is turned back to portrait, fullscreen should be disabled back to the original rendering (if it is possible)
> 
> I didn't really thought about video that are recorded in portrait mode directly so I think it is better to keep it in fullscreen portrait with a good video rendering and not a tiny content with lot of black on horizontal sides.
> 
> Let me know if you have more questions or if things are not clear enough. Anyway, thank you for working on the subject :)

Oh @anthony-robin got it! Thanks for making it clearer.

Issue #1038, opened 2024-08-31
[BUG] Profile Lightning payment is 100x too large


markwaters:
**Describe the bug**
Profile Lightning payment is 100x too large.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open a users profile in Amethyst.
2. Click on their lightning address.
3. A box pops up with a default zap amount of 1000 sats
4. Not using NWC , so wallet opens with a payment page ready to send 1,000,000 sats (!) , if I try to send 47 sats the wallet page shows 47,000 sats.

**Expected behaviour**
Sats amount to be correct.

**Device (please complete the following information):**
 - Pixel 7.
 - GrapheneOS.
 - 0.90.4-fdroid.

davotoula:
Reproduced in Google play version

davotoula:
Found the issue, fixing

davotoula:
What wallet supplier did the user use?

Does it happen every time for you or just certain users?

markwaters:
It happens for me using minibits and phoenix wallets.

I only tried two different user accounts and it happened identically for both.

davotoula:
Seems like a minor fix only. Hopefully @vitorpamplona will review and find no other side issues 👍

markwaters:
Thanks for your work on Amethyst @davotoula

davotoula:
All the credit goes to @vitorpamplona. Amazing project he runs.

davotoula:
Please close if resolved

markwaters:
Confirmed. Fixed. Thank you.

PR #1039, opened 2024-08-31
millisats were incorrectly multiplied by 1000

davotoula:


davotoula:
#1038

davotoula:
Hey @vitorpamplona, do you have any concerns about this fix?

- From looking at the code, the multiplication by 1000 should be done by ui and then milliSats are used throughout execution. This class accidentally did a second multiplication
- I've tested with the mentioned bug: lightning tip from user profile and it has fixed the issue of 1000x amounts.
- I've gone through the code and no other zap/cashu/lightning functionality is using the changed code path:`AccountViewModel.sendSats`.
- Minor rename to a field to make it clear milliSats should be used to avoid duplicate multiplication by 1000

vitorpamplona:
Thank you @davotoula! Releasing a version soon

PR #1042, opened 2024-09-02
Long press to copy relay url

greenart7c3:


davotoula:
Tested, it works.

Initially I was long pressing the relay icon hoping it would copy the url to clipboard, maybe add the functionality to the icon as well?

Maybe adding the functionality to relay icons under user (when Complete UI mode is used)?

PR #1051, opened 2024-09-04
Configurable media compression

davotoula:
As a media creator on Nostr it would be nice to have options of how much the client compresses media (photo and/or video) before it uploads it.

Some media might be high quality photo while others might be low quality memes. It would be nice to be able to change media compression for different use cases.

I experimented with high quality SLR photos, lower quality phone photos and 1080p/4k videos. I compared visually the degradation and difference from original.

Below are the new suggested compression levels for photos and videos:

Photos - With a high quality SLR photo of 16.7MB

- LOW = 40 resulting in 107KB
- MEDIUM = 60 resulting in 146KB
- HIGH = 90 resulting in 341KB
- Old default was 80

Videos - With original 1080p phone video size: 16.7MB

- LOW = VERY_LOW resulting in 1.1MB
- MEDIUM = MEDIUM resulting in 2.8 MB
- HIGH = HIGH resulting in 3.8 MB
- Old default was MEDIUM

My first bigger change involving UI so feel free to comment on everything!
screen_Shot_media

- [x]  Make MediaCompressor configurable
- [x]  Add Compression Level slider to UI
- [x]  Propagate Compression level setting from UI to MediaCompressor
- [x] Uploading from VideoScreen
- [x]  Localise all added text
- [x]  Refactor for cleaner code (use integer instead of float, use Enum instead of integer ASAP in the code)
- [x]  Investigate other media upload areas for regression (eg Nip96Uploader)
- [x]  Manual testing


amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewPostView.kt
```diff
@@ -1845,6 +1847,38 @@ fun ImageVideoDescription(
...
+                            when (sliderPosition.toInt()) {
+                                0 -> CompressorQuality.LOW.toString()
+                                1 -> CompressorQuality.MEDIUM.toString()
+                                2 -> CompressorQuality.HIGH.toString()
+                                else -> CompressorQuality.MEDIUM.toString()
+                            },
+                    )
+                }
+            }
+
```
KotlinGeekDev, line 1881: I wonder if in addition to the UX above, we should have a global media compression setting for the user, but which can be overridden here. cc @vitorpamplona

vitorpamplona:
This is super nice! It would be nice if we could know the limit of the server and then automatically default for the mode that will make it fit. But I am not sure if the lib gives us the expected size after compression

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewPostView.kt
davotoula: I'm considering that for version 2 but it's always a balance of not cluttering settings screen versus convenience.

Currently the default is set to medium in UI and most users will probably just click upload using medium.

A global setting would be useful for users that always want to upload low or high.

davotoula:
It seems the lib returns size after compression. We would have to keep compressing the **original** multiple times until result is below certain size and only then posting.

I feel like that's a different functionality than this "selecting compression level" as compression level would be ignored if user wants to squeeze media until it fits upload server.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewPostView.kt
KotlinGeekDev: > I'm considering that for version 2 but it's always a balance of not cluttering settings screen versus convenience.
> 
> Currently the default is set to medium in UI and most users will probably just click upload using medium.
> 
> A global setting would be useful for users that always want to upload low or high.

Exactly.

KotlinGeekDev:
> It seems the lib returns size after compression. We would have to keep compressing the **original** multiple times until result is below certain size and only then posting.
> 
> I feel like that's a different functionality than this "selecting compression level" as compression level would be ignored if user wants to squeeze media until it fits upload server.

Interesting. It's making me think of probably using the compression level as part of some UploadServerProperties state(or any other method), where the server and the (finally accepted) compression level would be stored in and tracked, so as to default to it when next the user chooses that server. But that complicates things.

On another note, now I am starting to think if it would be a nice idea for upload servers to advertise size limits/tiers to clients. It may potentially help a lot with this.

vitorpamplona:
Also, @realiefan has been doing some experiments with https://github.com/iefanx/compresspro in Flutter, which is using https://github.com/jonataslaw/VideoCompress flutter lib, which is a wrapper for https://github.com/deepmedia/Transcoder

Here is the part of the code that sets up the Transcoder for compression
https://github.com/jonataslaw/VideoCompress/blob/c8820bde3ecce4041f33bfcf24cc89730023188b/android/src/main/kotlin/com/example/video_compress/VideoCompressPlugin.kt#L78

He seemed to be able to get better compressions with that lib.

davotoula:
> Here is the part of the code that sets up the Transcoder for compression https://github.com/jonataslaw/VideoCompress/blob/c8820bde3ecce4041f33bfcf24cc89730023188b/android/src/main/kotlin/com/example/video_compress/VideoCompressPlugin.kt#L78
> 
> There are a lot more options than the current lib we use

This would be really fun to do as a follow-up. I still have my original photos and videos that were used as a reference.. and I still have the many converted versions (in various qualities).

davotoula:
Ok, I think that's it for version 1.

This was really challenging. First time that I was changing Views, View Models and adding UI elements.

davotoula:
Edit message
edit
New post dark
Screenshot_20240905_171929
New post from Video screen dark
Screenshot_20240905_172033
New post from Video screen LIGHT
Screenshot_20240905_172107
New post Light
Screenshot_20240905_172133

KotlinGeekDev (approved):
The v1 looks good to me. Nothing jumping at me(unless @vitorpamplona has some feedback). Nice PR @davotoula !

vitorpamplona:
Super nice!

vitorpamplona:
@davotoula do you want to add an option for no compression?

PR #1052, opened 2024-09-04
Fix token sanitization when using gcompatup

greenart7c3:
@vitorpamplona 
This is one of the reasons some people are not receiving notifications
You need to increase the size of the token field in the notification server too if you are still using the 255
The token was 257 here
This happens when using [gcompat-up](https://play.google.com/store/apps/details?id=org.unifiedpush.distributor.fcm) for notifications


vitorpamplona:
Very interesting. The new version of the server does not parse the token as before, so I think the server part is ready to go.

Issue #1084, opened 2024-09-16
[BUG] Unable to url preview certain links, eg substack.com


davotoula:
**Describe the bug**
Unable to url preview certain links, eg substack.com. The preview is empty of all content

**To Reproduce**
1. Create a new post
2. Paste in "https://ronaldrobson.substack.com/p/uma-conversa-com-antonio-vargas"
3. There is no URL preview of the link

**Expected behavior**
A URL preview with image, title and description

PR #1085, opened 2024-09-16
Cover URL preview case with property name with no value

davotoula:
Cover case with property name with no value, eg:

`<meta data-preact-helmet property="og:title" content="Some fancy title">`

Result:
Screenshot_20240916_155223

#1084


vitorpamplona:
Thanks david! I just modified a bit to add the value-less attribute to the list as well here: https://github.com/vitorpamplona/amethyst/commit/28da06ef010782a402c52f51b1ffd1603e64a821

davotoula:
That's a good idea. Might be something useful in the value less property 👍

PR #1090, opened 2024-09-20
Fix for notifications when using amber

greenart7c3:
We still have to change this later to work when using multiple accounts and the events are not allowed in amber but for now this fixes the notifications when amethyst is allowed to sign events in background


vitorpamplona:
Yeah, I didn't know how/if to call the launcher if the app is on the background. It will be weird to just be using the phone and then Amber pops up.

PR #1118, opened 2024-10-03
Refactor media compressor as prep for replacing 3rd party compression library

davotoula:
Added unit tests
Split out File handling to a separate utility class (also helps with testing)
Refactor MediaCompressor for clarity (see commit log for details)
Manual regression testing


davotoula:
Was great getting into testing in kotlin and especially mockk... all those lambdas 😆

vitorpamplona:
This is awesome! Thanks @davotoula

PR #1122, opened 2024-10-05
Re-enable tap to fullscreen behaviour with stricter orientation checks.

KotlinGeekDev:
This should fix the issue introduced by #1018 .


vitorpamplona:
This actually didn't fix the issue, right? A vertical video should stay vertical in portrait when the player goes to full screen. You can test with the my latest owl video on the Video feed. 

See the video: 

[Screen_recording_20241008_131006.webm](https://github.com/user-attachments/assets/f48b6dcf-bf6a-4786-9f39-d0f6bc7b0076)

vitorpamplona:
There is also a bug that keeps rotating the phone for images: 

[Screen_recording_20241008_145733.webm](https://github.com/user-attachments/assets/f28a790d-5c91-46b6-bfef-5618bb122f18)

KotlinGeekDev:
@vitorpamplona that is strange. Let me check it again. As for the images, the rotation behaviour should not affect them, since the PR only targeted livestreams.

PR #1131, opened 2024-10-10
Prioritise search results with user name starting with search term

davotoula:
It was a bit frustrating that user profiles with the search term anywhere in the description (eg lightning address) were mixed with users with user name same as search term.

Added comparison between search term and display name to comparator to prioritise results with search term in name.

Before:
before

After:
after

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
@@ -74,8 +74,13 @@ class SearchBarViewModel(
         _searchResultsUsers.emit(
             LocalCache
                 .findUsersStartingWith(searchValue)
-                .sortedWith(compareBy({ account.isFollowing(it) }, { it.toBestDisplayName() }))
-                .reversed(),
+                .sortedWith(
+                    compareBy(
+                        { it.toBestDisplayName().startsWith(searchValue, true) },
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Hum... are the screenshots reflective of the final solution? Why is "I want alby" the first if it doesn't start with "alby"?

davotoula:
Good catch. I think the screenshot was taken when contains was used rather than startsWith. Let me update screenshot.

I wanted to reset the result list to first item every time an update happens because the most relevant results will be at top but could figure it out.

vitorpamplona:
That makes sense. 

My only question now is if we want the startWith to be first or second, after the isFollowing. 

If it is after, Alby would still be your first record, but it would be followed by all the people you follow that have alby in the name and then all the random stuff. I think the Alby search is an outlier because so many people have alby in their wallet address, but if you are searching for somebody's second name, then placing it in second might make more sense.

davotoula:
Searching for somebody's second name looks pretty good as well?

after2

I tried a few other scenarios with Strike, Alby, Pamplona, Bitcoin. They all look good with `startsWith` being placed first.

davotoula:
In either case I don't have an attachment to either and am happy to move it around 😄

Just the fact of adding the startsWith has helped already!

vitorpamplona:
Very cool, let's see how people react to it.

davotoula:
I think it comes down to whether you are discovering new users (startsWith first) or searching among your friends/followers (is Following first).

Maybe in the future we can have a friends/family icon (same as relay screen 
) next to the search text that when clicked will limit search to followers only

Issue #1119, opened 2024-10-05
[BUG] Unable to expand original note when replying


dtonon:
**Describe the bug**
As per subject, the "Show more" button does nothing or is absent.

 - App Version 0.92.1

bitcoinuser:
I was about to open an issue about this, haha, Thanks.

geovnn:
For anyone willing to look at this: the problem is in ui/actions/NewPostView.kt line 361. The row which contains the original note has a height constrain of 200 dp, solutions might be:
- Add a verticalScroll to this row
- Remove the height constrain
- Add a paramether to NoteCompose to specify the note height when collapsed (when the "show more" button is visible)

vitorpamplona:
The new post screen has had so many bugs with dynamic content that we had to fix it to 200.dp to solve some of them. 

The video below demonstrates some of this. The massive black space between the buttons and the keyboard randomly appears and the randomness seems to get worse when the replying post's content is bigger. 

I have looked at this code for so long (months) and I can't find the culprit. 

The black box is also not always the same size, which is very triggering. 

[Screen_recording_20241010_132222.webm](https://github.com/user-attachments/assets/735a3381-124c-45d8-b6df-90e614abecf9)

vitorpamplona:
We could try to convert it to a route to see if it improves.

vitorpamplona:
I am starting to hate these popup-based designs. (old stuff)

geovnn:
Sure it would be more appropriate, it's a bit too crowded for a popup. I will update here if I implement something.

dtonon:
@vitorpamplona if you have problems with the current components, a good and simple alternative is to have first the compose area, that expands itself as the user add more lines, and _below_ put the full original note. This way with a simple scroll up the user can review the original note, and then scrolling down go back to composing. And this happens only if needed, without always wasting top screen valuable space.
I would also move the submit as floating button to the bottom-right; having the original note at the bottom it never covers the user's text.

geovnn:
I made the changes to make NewPostScreen route based, seems to solve the issue and I haven't found any bug. Had to change other files to fix the imports, as other components used some in the NewPostView (which I moved to screen>loggedIn, maybe common components should be put in another file?). I will make a pull request now but it should be tested some more.

bitcoinuser:
This is already fixed and can be closed, right?

davotoula:
Not released yet. You may want to wait closing it until you have verified the fix?

[...]

PR #1137, opened 2024-10-14
Change NewPostView to NewPostScreen to make it route based

geovnn:
As discussed in #1119

Changes:
- Moved NewPostView to NewPostScreen and replaced Dialog with DisappearingScaffold
- Previous conditional call to NewPostView now start the navigation to the route passing the hexid of the baseReplyTo, quote, fork, version and draft parameters
- accountViewModel.getNoteIfExists(hex) is called to retrieve the parameters
- Fixed imports which used stuff from the NewPostView file


vitorpamplona:
Yeah, but we can switch to the new Route system that passes full objects as jsons, like this: https://www.youtube.com/watch?v=AIC_OFQ1r3k

geovnn:
Cool, didn't know about this!

PR #1152, opened 2024-10-23
Add bottom bar in DVM feed view

believethehype:


vitorpamplona:
Do we need the bar? None of the Threaad views have bars anymore (similar to how twitter does, you have to slide back to get the bar)

believethehype:
I'm a bit neutral on this one, it depends on your point of view. Point 1) An interaction with a DVM is like a thread and you follow the "thread" and go back to leave the "thread". Point 2) A DVM feed is like the main feed and only if you click a note you enter a thread. Design decision.

Issue #684, opened 2023-11-09
[Question] How to setting/config auto-translation ?


ttpro1995:
- I want to turn on/off auto translate to some language
- I accident turn disable auto translate into some language, and cannot find a way to turn it on. 
- ... 

Summary: where is the setting for auto translate ? I can't find it.

davotoula:
I'm also interested in this. I suspect by default it doesn't translate system languages you have added in android settings.

MoonDusk1996:
I'm also interested in this, I had automatic translation but somehow it was disabled or it's not working for me anymore

davotoula:
Translation settings have had a major update

PR #1168, opened 2024-11-05
Re-enable translations for all languages button in application settings

davotoula:
Several users have reported (#684) that once you set a language as "never translate from" it's not possible to revert it.

This is a simple solution by adding a button in the application preferences to reset ALL languages to be translated again.

Changes:

- Add a button to applications settings for Play version only
- Reset dontTranslateFrom field by setting it to an empty set
- Added translations

Testing:

- emulator (play/fdroid)
- Pixel 7 (play/fdroid)

It's not pretty but UI is not my forte:
button_in_settings

A different approach would be to allow a reset of a specific language via the note drop-down. I didn't see a straightforward way to pass the note language code to the dropdown so didn't choose this approach.

### DIFF
--- a/amethyst/src/fdroid/java/com/vitorpamplona/amethyst/ui/components/SelectNotificationProvider.kt
+++ b/amethyst/src/fdroid/java/com/vitorpamplona/amethyst/ui/components/SelectNotificationProvider.kt
@@ -194,7 +194,7 @@ fun PushNotificationSettingsRow(sharedPreferencesViewModel: SharedPreferencesVie
         SettingsRow(
             R.string.push_server_title,
             R.string.push_server_explainer,
-            selectedItens = readableListWithExplainer,
+            selectedItems = readableListWithExplainer,
             selectedIndex = list.indexOf(currentDistributor),
         ) { index ->
             if (list[index] == "None") {
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -1068,6 +1068,11 @@ class Account(
         sendNewAppSpecificData()
     }
 
+    fun resetDontTranslateFrom() {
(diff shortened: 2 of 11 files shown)


KotlinGeekDev (commented):
Looks good to me.

KotlinGeekDev:
I was thinking that the UI for this could be an entry, to open language-specific settings for the app. The position of this button would remain the same there.
Please, what do you think @davotoula @vitorpamplona ?

davotoula:
I'm not a fan of nested settings but maybe that's where amethyst has to go since it is getting more complex.

Saying that, this enhancement would get the feature out quickly to users.

I'm not good at UI asi maybe somebody else should build a new settings UI. I could however collaborate with somebody that provides pretty UI screens...

There is a second option I would consider if @vitorpamplona insists... Trying to make the reset of translation for a specific language in the note drop down. And skipping the new button in settings...

vitorpamplona:
Yeah, we will need to build a settings menu for all Account-based settings. The only issue here is that the top items will affect all users, while the button is unique to the active account. :(

KotlinGeekDev:
Maybe we could do a header-content style approach to the Settings page(maybe implementing a section as a drop-down(in line with @davotoula 's idea?) ).

davotoula:
Marked as draft because I'm looking into a solution in the note dropdown. Allowing to reset single language at a time and avoiding changing the settings page.

vitorpamplona:
@davotoula the next version of Amethyst will save these local settings encrypted in the relay as well. So, it will be more important to have a way to reset them. :)

davotoula:
I've been distracted by other amethyst work 😆

There is a cleaner way to do the language reset in the note drop down. I implemented a way to detect the disabled language in the note view, just need a way to communicate it to the drop down... just like private bookmark and edit draft gets updated in the drop-down WatchBookmarksFollowsAndAccount.

work in progress!

davotoula:
Deleting this pull request as a "per note" solution will be cleaner.

Issue #815, opened 2024-03-23
[FEATURE] improved list/set management


vicariousdrama:
**Is your feature request related to a problem? Please describe.**

Lists are a key method by which users organize content. The current support for lists in amethyst requires users to leave and use another client that offers the ability to work with lists beyond uncategorised bookmarks and mute lists

**Describe the solution you'd like**

As a user, I want to have full control of my lists within the mobile application.

- All of my lists that are not marked as deleted should appear in the drop-down to use for filtering.
- A user should be able to **create** a new named list/set (name/title)
- A user should be able to **rename** a list/set (to correct for spelling errors)
- A user should be able to **copy** a list as a new list/set
- A user should be able to **add a pubkey** (p tag) to a list from the user's profile, or from an event they published (e.g., menu action to `Add author to list`, then present a submenu or modal allowing selection of an existing list or creating a new list)
- A user should be able to **add an event** (e tag) id to a list from the event when viewed (e.g. menu action to `Add event to list`, then present a submenu or modal allowing selection of an existing list or creating a new list)
- A user should be able to **see all items** on their list, and an indicator of the kind
- A user should be able to **add a URL** (r tag) to a list from where a list is being managed.
- A user should be able to **delete** an item from the list where its being managed.
- A user should be able to **reorder** items in their list (can be implemented as move up/move down or drag and drop or other suitable approach)

Nice to haves, but not required
- ability to set and update a picture (image) for a list, and its about (description) tag
- listing lists of followers and being able to make a fork of them

In a nutshell, a user should have full CRUD over their personal lists within Amethyst

**Bounty (in Bitcoin sats) offered for the implementation**

I (npub1yx6pjypd4r7qh2gysjhvjd9l2km6hnm4amdnjyjw3467fy05rf0qfp7kza) am offering 1_000_000 sats for this feature to be added for kinds 10003, 30000, 30003 within constraints of allowed tags

Please advise if this should be broken down into multiple issues and bounties.

Inconsistencies and shortcomings of NIP51 may need to be addressed prior to implementation so that clients can get on the same page

vicariousdrama:
I (npub1yx6pjypd4r7qh2gysjhvjd9l2km6hnm4amdnjyjw3467fy05rf0qfp7kza) am increasing my bounty offer to 2_000_000 sats for this feature to be added

KotlinGeekDev:
[...]

PR #1169, opened 2024-11-05
List/Set management: Part 1, Profiles

KotlinGeekDev:
Implements list management support for profiles. 
This is the first part of the solution for #815 .


KotlinGeekDev:
OK. The PR is ready now. Please have a look @vitorpamplona @davotoula @greenart7c3 .

davotoula:
Super well done! I will have a thorough look!

KotlinGeekDev:
Here is a demo of how things work: 

https://github.com/user-attachments/assets/367e70eb-3591-4872-af11-90ca7116c0e1

davotoula:
This is a game changer BTW. It was too fiddly managing lists via various obscure websites that needed your login and could corrupt your lists.

KotlinGeekDev:
> This is a game changer BTW. It was too fiddly managing lists via various obscure websites that needed your login and could corrupt your lists.

I was planning to make a dedicated app for lists as well. :sweat_smile:

davotoula:
> > This is a game changer BTW. It was too fiddly managing lists via various obscure websites that needed your login and could corrupt your lists.
> 
> I was planning to make a dedicated app for lists as well. :sweat_smile: 


Call it Yanla - yet another nostr lists app

davotoula:
Merged already?! Nice!

vitorpamplona:
Yeah, these larger PRs are never perfect. So it is better to merge them and keep working with smaller PRs. 

In this particular one, I think the UI can be better. But we can keep pushing improvements over time

KotlinGeekDev:
> Yeah, these larger PRs are never perfect. So it is better to merge them and keep working with smaller PRs.
> 
> In this particular one, I think the UI can be better. But we can keep pushing improvements over time

Please @vitorpamplona what aspects of the UI can be made better?

vitorpamplona:
On light mode, the rounded border of the list entries is kind ugly. Maybe use just a line between them as with the list of people in the Messages screen. There is also no padding from the border to the start-end of the screen. 

Inside each list the 3 dots up top doesn't have border. The spacingb between the follow and trash buttons is big. The rouded corners between them are also different. And the trash button is probably too big. In this screen there is too much start-end border.

KotlinGeekDev:
> On light mode, the rounded border of the list entries is kind ugly. Maybe use just a line between them as with the list of people in the Messages screen. There is also no padding from the border to the start-end of the screen.
> 
> Inside each list the 3 dots up top doesn't have border. The spacingb between the follow and trash buttons is big. The rouded corners between them are also different. And the trash button is probably too big. In this screen there is too much start-end border.

I see. I had not tested the app in light mode. :sweat_smile:  
Thanks for the feedback.

KotlinGeekDev:
> Inside each list the 3 dots up top doesn't have border. The spacingb between the follow and trash buttons is big. The rouded corners between them are also different. And the trash button is probably too big. In this screen there is too much start-end border.

I kept the rounded corners different because the 'Follow/Unfollow' button rounded corners do not look good on the trash button.

vitorpamplona:
Also, I don't know how to add items to a list. 

Maybe the Follow button is not that relevant in these entries. Or is it? It's kind a confusing being in a "Follow Set" and seen a button that you are "not following" that user. I know it's because of the kind 3 status, but users don't know...

KotlinGeekDev:
> Also, I don't know how to add items to a list.
> 
> Maybe the Follow button is not that relevant in these entries. Or is it? It's kind a confusing being in a "Follow Set" and seen a button that you are "not following" that user. I know it's because of the kind 3 status, but users don't know...

From the specs and from what I understand, 'following' a user adds them to the kind 3 list, which is different from the follow set(kind 30,000). One thing I do not know is if that distinction matters for following a user. If it doesn't matter, then a check could be added to Amethyst for that.

vitorpamplona:
The Spec is largely irrelevant here. If we call "Follow Sets", it is confusing for users to see a "Follow" button on people who are inside the Set.

I don't know how to solve this, but it is our own issue.

KotlinGeekDev:
> Also, I don't know how to add items to a list.
> 

For adding a profile, there is the list button next to the follow button when the profile page is opened. A user can also be added from a note they posted, using the 'Add author to set' option.

KotlinGeekDev:
> The Spec is largely irrelevant here. If we call "Follow Sets", it is confusing for users to see a "Follow" button on people who are inside the Set.
> 
> I don't know how to solve this, but it is our own issue.

Maybe we could make the follow button open the set management dialog, and present the current kind3 follows as the default follow list, along with the other follow sets, and he adds the user to any number of them.

Issue #1166, opened 2024-11-03
[BUG] Onscreen keyboard pops up all over the place


yojimble:
When browsing nostr the phone keyboard has started popping up randomly, renders the app unusable as you never know when it will show up to get in the way

https://github.com/user-attachments/assets/eef34eab-91f9-4ed9-8975-f5997386d516

**To Reproduce**
use the app

**Expected behavior**
keyboard pops up in response to text boxes only

amethyst version v0.92.7

davotoula:
What phone, amethyst Version (google or fdroid) and keyboard (built in gboard or third party) are you using?

yojimble:
it's a samsung galaxy A5, android 8, amethyst from play store, samsung built in keyboard although I have tried others and they all do it

davotoula:
Oooo.. Android 8! I'll see if I can recreate it!

yojimble:
> Oooo.. Android 8!!

🤪 tryin to keep it going for as long as possible!

davotoula:
I have seen several mentions online about a "popping up Samsung onscreen keyboard". Can you try the suggested fixes?

- Gboard is reported as not having the issue, please try it
- https://r1.community.samsung.com/t5/galaxy-a/keyboard-list-and-default-bug/td-p/24895295
- https://www.youtube.com/watch?v=_4JnrBUUINk

There is also a minor code change we can try out. Do you know how to manually install an app from a download from Github so that we can try  a "debug" version that will not replace your main Amethyst app?

yojimble:
Super thanks, I have installed gboard and possibly it is popping up less - although still when clicking my own profile (viewing not editing)

I know how to install an .apk

davotoula:
Try this debug apk please.

https://github.com/vitorpamplona/amethyst/actions/runs/11693860560/artifacts/2149519494

yojimble:
Hi, this seems to have worked for the inbuilt samsung keyboard thankyou! where can I send a zap????

davotoula:
@yojimble cool it worked. I need to discuss with Vitor whether we can merge this safely to main amethyst.

In the meantime, can you also try removing the main amethyst from memory management (not sure android 8 had it) and try it again?

yojimble:
I don't know if it has anything called that, I could force close it and clear the cache... then reset my emojis

davotoula:
You will find Battery Optimization in your main android settings:

[...]

PR #1170, opened 2024-11-05
hide keyboard on resume

davotoula:
workaround for #1166

### DIFF
--- a/amethyst/src/main/AndroidManifest.xml
+++ b/amethyst/src/main/AndroidManifest.xml
@@ -73,7 +73,7 @@
             android:name=".ui.MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="adjustResize"
+            android:windowSoftInputMode="stateHidden"
             android:configChanges="orientation|screenSize|screenLayout"
             android:theme="@style/Theme.Amethyst">
 


davotoula:
This seems to have resolved this issue. 

@vitorpamplona my only guess is that low memory on the device stops and resumes the app occasionally and the `windowSoftInputMode="adjustResize"` pops up the keyboard on each resume.

### Old phones (tried Oreo):
The adjustResize causes the keyboard to always be up when app is resumed. From google docs: _"Android gives focus to the first text field in your layout when the activity starts"._ (https://developer.android.com/develop/ui/views/touch-and-input/keyboard-input/visibility)

It looks like the "Status update" field is put in focus and the keyboard is shown.

Exiting a text field does NOT close the keyboard on older phones. 

Using stateHidden will not show keyboard on resume... Even if a text field is in focus (eg. create a not or login screen nsec). One has to click on the text field for keyboard to show up

### Newer phones (tried API 34):
The keyboard is not shown on resume, the "Status update" is not in focus.

If you exit a text field the keyboard is closed automatically.

Only drawback with this change I noticed on newer phones: If you exit app while in a text field and resume the keyboard will not show until you click the text field again.

KotlinGeekDev:
> 
> Only drawback with this change I noticed on newer phones: If you exit app while in a text field and resume the keyboard will not show until you click the text field again.

Interesting. IMHO I don't know if it's much of a drawback though. :thinking:

davotoula:
Just investitigating one more on device workaround before requesting this pull request.

KotlinGeekDev:
> Just investitigating one more on device workaround with the user before requesting this pull request.

Got it.

davotoula:
Marking as ready for merge as the user reported this fix is the only thing that resolves their "popping up-onscreen keyboard" issue on their Samsung A5 Android 8 device.

vitorpamplona:
Hum... isn't this an issue with the status update in the drawer being in focus? That element is not visible so the code for the drawer shouldn't even be running. 

We might need to keep resize there, like on `stateHidden|adjustResize`, but the issue with not showing the keyboard on newer phones if you leave the app on the new post screen is quite annoying.

davotoula:
I agree. There should be a higher percentage of newer phones that benefit from keyboard being closed/open promptly than this fix.

Maybe there is a solution to avoid focus in elements that are not visible. Perhaps using `FocusRequester()` in addition to the `focusManager`.

I'm happy to close this!

PR #1171, opened 2024-11-07
Crashfix: test for empty string in mentions

davotoula:
Observing a crash when trying to comment on a specific note. Root cause looks a missing hex in the mention tag of the post:

`{
  "id": "2be8bcd50a46628a08523c4b638406e04d1d7ab75c8f9ede5aaaf568e4bde081",
  "pubkey": "20986fb83e775d96d188ca5c9df10ce6d613e0eb7e5768a0f0b12b37cdac21b3",
  "created_at": 1730957039,
  "kind": 1,
  "tags": [
    [
      "p",
      "",
      "",
      "mention"
    ]
  ],`

Event ID: `nevent1qqszh69u659yvc52ppfrcjmrssrwqnga02m4eru7med24atguj77pqgpzamhxue69uhhyetvv9ujumn0wd68ytnzv9hxgtczyqsfsmac8em4m9k33r99e803pnndvylqadl9w69q7zcjkd7d4ssmxqcyqqqqqqgx0u6l0`

The  `isValidHex` function specifically fails empty hexes so I filter out the empty hex mention in the calling `NewPostViewModel` class.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewPostViewModel.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewPostViewModel.kt
@@ -222,7 +222,10 @@ open class NewPostViewModel : ViewModel() {
                 if (replyNote.event !is CommunityDefinitionEvent) {
                     replyNote.author?.let { replyUser ->
                         val currentMentions =
-                            (replyNote.event as? TextNoteEvent)?.mentions()?.map { LocalCache.getOrCreateUser(it) }
+                            (replyNote.event as? TextNoteEvent)
+                                ?.mentions()
+                                ?.filter { it.isNotEmpty() }
+                                ?.map { LocalCache.getOrCreateUser(it) }
                                 ?: emptyList()
 
                         if (currentMentions.contains(replyUser)) {
--- a/amethyst/src/test/java/com/vitorpamplona/amethyst/ui/actions/NewPostViewModelTest.kt
+++ b/amethyst/src/test/java/com/vitorpamplona/amethyst/ui/actions/NewPostViewModelTest.kt
@@ -1,0 +1,125 @@
+/**
+ * Copyright (c) 2024 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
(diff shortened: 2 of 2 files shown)


KotlinGeekDev (commented):
Nice, though the root issue I suspect, is the client producing this event. It will probably need to be reported to them.

davotoula:
Yes.. I need to check nostr docs whether empty mentions like that are valid.. Or maybe just not defined.

In any case amethyst should expect malformed messages and not crash. Display a toast to user.. Or just ignore silently?

KotlinGeekDev:
@davotoula you seem not to the only one to have noticed(based on the replies to that note), and from opening it on many clients, most silently ignore it, and render it as is ("@npub1..."). Only Coracle looks for the referenced profile.

KotlinGeekDev:
IMG_20241107_132212
@davotoula seems to work on Amethyst?

davotoula:
That's with the debug build from pull request

KotlinGeekDev:
> That's with the debug build from pull request

This is actually with the latest release build.

davotoula:
@KotlinGeekDev can you please double check? I really don't see how it could work with release build.

Note id: nostr:nevent1qqszh69u659yvc52ppfrcjmrssrwqnga02m4eru7med24atguj77pqgpzamhxue69uhhyetvv9ujumn0wd68ytnzv9hxgtczyqsfsmac8em4m9k33r99e803pnndvylqadl9w69q7zcjkd7d4ssmxqcyqqqqqqgx0u6l0

KotlinGeekDev:
> @KotlinGeekDev can you please double check? I really don't see how it could work with release build.
> 
> Note id: nostr:nevent1qqszh69u659yvc52ppfrcjmrssrwqnga02m4eru7med24atguj77pqgpzamhxue69uhhyetvv9ujumn0wd68ytnzv9hxgtczyqsfsmac8em4m9k33r99e803pnndvylqadl9w69q7zcjkd7d4ssmxqcyqqqqqqgx0u6l0

Got it.

davotoula:
Which reply from above do you think shows it works with latest amethyst release?

davotoula:
By the way, viewing works.

It crashes if you click the create a comment button.

KotlinGeekDev:
> By the way, viewing works.
> 
> It crashes if you click the create a comment button.

Yeah, you're right about that.

vitorpamplona:
It looks good, I just need to get home to do some testing and then merge.

PR #1185, opened 2024-11-21
Enable unfollow in note dropdown

davotoula:
Just noticed this when working on something else.

Any reason why "unfollow" wasn't in the dropdown like the "follow" is?


vitorpamplona:
Yeah, i think at the time we didn't feel the need to.

Issue #1186, opened 2024-11-22
[BUG] Exception when opening configuration of "Media Servers"


GitKalle:
**Describe the bug**
When I try to open the configuration of the "Media Servers" the application crashes. But this just happens on one of my three accounts.

**To Reproduce**
1. Open Amethyst
2. Use the relevant account
3. Select "Media Servers" in the sidebar

**Expected behavior**
The configuration of Media Servers opens.

**Output of logcat**
```
11-22 08:56:12.299 16484 16484 E AndroidRuntime: FATAL EXCEPTION: main
11-22 08:56:12.299 16484 16484 E AndroidRuntime: Process: com.vitorpamplona.amethyst, PID: 16484
11-22 08:56:12.299 16484 16484 E AndroidRuntime: java.lang.IllegalArgumentException: Key "https://nostr.build" was already used. If you are using LazyColumn/Row please make sure you provide a unique key for each item.
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.ui.layout.LayoutNodeSubcompositionsState.subcompose(SourceFile:23)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.ui.layout.LayoutNodeSubcompositionsState$Scope.subcompose(Unknown Source:2)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.foundation.lazy.layout.LazyLayoutMeasureScopeImpl.measure-0kLqBqw(Unknown Source:35)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.foundation.lazy.LazyListMeasuredItemProvider.getAndMeasure-0kLqBqw(Unknown Source:14)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.foundation.lazy.LazyListMeasuredItemProvider.getAndMeasure-0kLqBqw$default(Unknown Source:8)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.foundation.lazy.LazyListMeasureKt.measureLazyList-x0Ok8Vo(Unknown Source:331)
11-22 08:56:12.299 16484 16484 E AndroidRuntime:        at androidx.compose.foundation.lazy.LazyListKt$rememberLazyListMeasurePolicy$1$1.invoke-0kLqBqw(Unknown Source:516)
[...]

PR #1187, opened 2024-11-22
Append server type to lazy column index to make it unique

davotoula:
Fix for #1186

Currently the app allows entering same media server as nip96 and blossom.

The server name is used as key for LazyColumn and this breaks LazyColumn as it doesn't allow duplicate key.

This fix prepends server type before serve name when used as key which makes it unique. This will prevent crash for any users that have already entered duplicate media servers.

Going forward do we want allow same media server as both nip96 and blossom or should we prevent it?

If we allow a media server as both nip96 and blossom should we make that clear on upload (not part of this pull request):
Screenshot_20241122_164540


vitorpamplona:
this won't work. I am taking a look right now

davotoula:
It resolved the issue but maybe there is a better more complete fix.

PR #1214, opened 2024-12-07
Move relay dialog to a route

greenart7c3:
Fixes the bottom of the screen rendering
The object RelayToAdd was added because I didn't want to mess with encoding in the routes
Another way to fix this is changing Dialog for a ModalBottomSheet


vitorpamplona:
This is good. We are likely to migrate all Dialogs to their own routes. 

I tried fixing this issue on dialogs, but for some reason there are padding bugs everywhere. 

So, routes it is.

PR #1229, opened 2024-12-18
Account backup qr codes

davotoula:
Amethyst has a great QR code scanning feature in login screen but I didn't see a way to export keys as QR code.

New fucntionality:
- Show QR code for private key. The QR code can be scanned by a different device to allow for quick account import.
- Show QR code for private key or for encrypted private key. You can print the QR code of encrypted private key and keep as backup and scan at a later date and provide password to allow import.

https://github.com/user-attachments/assets/d9c31eb5-b482-4073-bd62-853ebb740352

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/DrawerContent.kt
@@ -539,7 +539,7 @@ fun ListContent(
         MediaServersListView({ editMediaServers = false }, accountViewModel = accountViewModel, nav = nav)
     }
     if (backupDialogOpen) {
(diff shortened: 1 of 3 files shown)


vitorpamplona:
Super nice!

Shall we add the Profile picture and the user's name to the QR page just like when showing your npub just to make the QR page look nicer?

Maybe with a warning that this is an NSEC/Ncryptsec and the user should not show it to anyone else. 

At some point, we should create a PDF with the QR with the ncryptsec to help users print and store with their other documents at home. 

Maybe we can use the 1Password account backup sheet as a template. We wouldn't need this many fields. But putting the ncryptsec in the QR and allowing the user to manually write down their password would be an awesome way to share the file with a printer (which logs things down) without exposing the nsec directly.

davotoula:
I like the idea of making it nicer and more account specific. Didn't think of that!

I'll see what I can do!

Issue #1202, opened 2024-11-30
[BUG] Expand a post then click on a profile mentioned in the post and then android back button, the post sometimes is automatically collapsed but shouldn't


bitcoinuser:
Hi,

Expand a post then click on a profile mentioned in the post and then android back button, the post sometimes is automatically collapsed but shouldn't. See the video to understand:

https://github.com/user-attachments/assets/626ac9ab-13cf-4ba5-bf0f-9af437cc5a7e

davotoula:
This fix has been released. Can you please verify if it resolves your issue?

davotoula:
@bitcoinuser

vitorpamplona:
Closing as stale — waiting on more info since 2024, and the feed/navigation state handling has been reworked several times since. If posts still collapse after back-navigation on a current release, please file a new report.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1250, opened 2025-01-05
Keep note state across LazyColumns recompositions

### LINKED ISSUE #1203: [BUG] Expand reactions, scroll the feed, the reactions automatically collapse but shouldn't
Hi,

Expand reactions, scroll the feed, the reactions automatically collapse but shouldn't.

https://github.com/user-attachments/assets/0095e6b3-2195-4041-bc1a-847334b5ac4e

geovnn:
Fixes #1203 
The same problem happens with the "show more" state when the note contains enough text.
This is because the LazyColumn in RenderThreadFeed removes items from memory when they are no longer visible. Changing the variables that maintain the state in ReactionsRow and ExpandableRichTextViewer to use rememberSaveable instead of remember seems to fix this problem.


geovnn:
also fixes #1202

Issue #1258, opened 2025-01-21
Crash while typing string in search [BUG]


alltheseas:
**Describe the bug**
Search crash occurs before I can type in a full query

**To Reproduce**
1. Open amethyst
2. Top right search icon
3. Start typing query (e.g. Sebastix)
4. Observe crash

**Expected behavior**
Do not crash while I am typing search term. 

**Video and Screenshots**

![Image](https://github.com/user-attachments/assets/ab4f5314-04b5-450a-ae67-a469de3f5b4b)

**Device (please complete the following information):
DC-01

 - Android Version 13 
20241220.1638.user
 - App Version
0.94.3
 - Amber Version (if using it to sign): n/a

**Bounty (in Bitcoin sats) offered for a solution**
The size of the bounty is proportional to how much this matters to you. If no bounty is offered,
not even a small one, this bug will not be worked on because it doesn't matter to you. We prioritize
bug fixing to issues that have bounties, even small ones.

alltheseas:
I can search for "Vitor" successfully

greenart7c3:
I could not reproduce this. Are you in the latest version 0.94.3? Can you provide a crash report?

alltheseas:
android 13 . I'll check shortly

alltheseas:
Of course now that I try to recreate the bug on a new day, I cannot recreate.

ksedgwic:
@alltheseas:
Pixel Tablet w/
android version: `Android 15`
amethyst version: `v0.94.3-PLAY`
search string: `sebastix` (I only get to "seb" and it crashes, something evil found?)

Other seaches seem ok, "sebastix" is evil ...

davotoula:
I can confirm the crash on searching "seb".

Google play, pixel 7, 94.3

Having a look now.

davotoula:
Reproducible only after a "cold start" of amethyst (after using force close) and quickly navigating to search bar.

Other search queries than "seb" do not reproduce the issue. Must be something in the search results of "seb"?

`06:00:51.904  E  FATAL EXCEPTION: main 
                 Process: com.vitorpamplona.amethyst, PID: 15380
                 java.lang.IllegalArgumentException: Can't represent a width of 5196264 and height of 0 in Constraints
                 	at androidx.compose.ui.unit.ConstraintsKt.invalidConstraint(Unknown Source:29)
                 	at androidx.compose.ui.unit.ConstraintsKt.createConstraints(Unknown Source:26)
[...]

PR #1265, opened 2025-01-28
Exclude file header events from search results

davotoula:
Exclude FileHeaderEvent from note search results as they crash the app due to large content size. 

In Direct note id search and in free text search

#1258

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -1845,6 +1845,21 @@ object LocalCache {
         )
     }
 
+    /**
+     * Will return true if supplied note is one of events to be excluded from
+     * search results.
+     */
+    private fun excludeNoteEventFromSearchResults(note: Note): Boolean =
+        (
+            note.event is GenericRepostEvent ||
+                note.event is RepostEvent ||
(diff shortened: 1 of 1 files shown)


amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
```diff
@@ -1845,6 +1845,21 @@ object LocalCache {
...
     }
 
+    /**
+     * Will return true if supplied note is one of events to be excluded from
+     * search results.
+     */
+    private fun excludeNoteEventFromSearchResults(note: Note): Boolean =
+        (
+            note.event is GenericRepostEvent ||
+                note.event is RepostEvent ||
```
KotlinGeekDev, line 1855: Hmm, not sure why these two need to be excluded. Maybe we could add a check for if the repost event is also one of the others we are excluding?(like FileHeaderEvent)

KotlinGeekDev (commented):
Nice!

amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
KotlinGeekDev: Got it.

vitorpamplona:
Looks good! Later on we need to figure out how to render these without crashing.

Issue #1294, opened 2025-03-15
[BUG] Failed to save the image when MIME type is application/octet-stream


davotoula:
**Describe the bug**

Failed to save the image
MIME type application/octet-stream cannot be inserted into content://media/external/video/media; expected MIME type under video/*"

**To Reproduce**
1. Go to nostr:nevent1qqstngh6em9dlth5mqfhgd94nfq88wamc73auw45v7vngvrklnmw6nqpz4mhxue69uhkummnw3ezummcw3ezuer9wchsygr3aj4a3d4nx4ywqa0lqxe3268lmgvaptp83qr8mxfj33k7fzzew5psgqqqqqqsqmcp3j
2. Click on Download button
3. See the error

**Expected behavior**
Images are downloaded to image folder, videos are downloaded to video folder

**Video and Screenshots**
![Image](https://github.com/user-attachments/assets/4bbc450d-76ef-41e2-9dca-af17a6547c8a)

**Device (please complete the following information):**
 - Phone Brand/Model e.g. Pixel 7 Pro
 - Android Version e.g. 33
 - App Version v0.94.3 play

vitorpamplona:
I think this has been fixed on the main branch.

davotoula:
I'm just submitting a pull request.

davotoula:
Verified

PR #1295, opened 2025-03-15
Failed to save the image

davotoula:
A web server can return media with contentType application/octet-stream and missing mime type.

Eg: https://smartflowsocial.s3.us-east-1.amazonaws.com/clients/cm7kdrwdk0000qyu6fwtd96ui/cc408b84-689f-4403-875b-ff91ba077a28.jpg

Currently the app will then set mime type to application/octet-stream. This breaks the download as application/octet-stream is not a valid mime type to be saved.

This fix uses the file extension to try and guess the mime type.

Another solution would have been to use magic number in file header but hopefully this fix is sufficient.

Testing 

On >=Q and < Q
Image with null mimetype and content type application/octet-stream
Video with null mimetype and content type application/octet-stream

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
@@ -43,6 +43,7 @@ import okio.IOException
 import okio.buffer
 import okio.sink
 import okio.source
+import okio.use
 import java.io.File
 import java.util.UUID
 
@@ -124,8 +125,8 @@ object MediaSaverToDisk {
                                 checkNotNull(contentType) { "Can't find out the content type" }
 
                                 val realType =
-                                    if (mimeType != null && contentType == "application/octet-stream") {
-                                        mimeType
+                                    if (contentType == "application/octet-stream") {
+                                        mimeType ?: getMimeTypeFromExtension(url)
                                     } else {
                                         contentType
                                     }
@@ -153,6 +154,11 @@ object MediaSaverToDisk {
                 },
             )
     }
+
+    private fun getMimeTypeFromExtension(fileName: String): String =
+        fileName.substringAfterLast('.', "").lowercase().let {
+            MimeTypeMap.getSingleton().getMimeTypeFromExtension(it).orEmpty()
+        }
 
     fun save(
         localFile: File,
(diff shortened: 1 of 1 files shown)


davotoula:
Fix for #1294

vitorpamplona:
oh interesting... Can you test if you can save an image that was sent on a NIP-17 DM to yourself? 

`application/octet-stream` is required when uploading and downloading DMs because the image is encrypted an needs to be converted into the mimetype of described in the kind:15 event. 

I just want to make sure this PR doesn't break the DM download and save locally.

davotoula:
Will test that and mark as "ready to merge" when tested

davotoula:
Hmm... DM image uploading isn't working for me on current release version... neither normal nor nip17.

What media server do you suggest for normal DM and nip17 dm?

vitorpamplona:
Yeah, most media servers don't accept encrypted files. :(

I use https://satellite.earth/ for DMs

davotoula:
Sent you amethyst DM. Please send encrypted DM image

davotoula:
Media download in Encrypted DM unaffected works by this fix (still works).

While testing I noticed that:
- on 0.94.3 media download in encrypted DMs is not working. Media in DM comes through as application/octet-stream which fails due to the bug
- on main this has been resolved, mime type is detected before attempting download so it works
- I'm getting failed downloads when url from media server contains dots (.). I suspect amethyst upload could be added garbage to url but needs more investigation. Example url which fails to download... Media was uploaded from Amethyst `https://cdn.nostrcheck.me/ca89cb11f1c75d5b6622268ff43d2288ea8b2cb5b9aa996ff9ff704fc904b78b/dfb500ba6f435f7f6bb04f69195e3349a37a38988a0ac9d4b0cd0e51272abbed.webp#x=1f461f3e70013974d9bd30ada326ebcd7c6a55ab943f457ad29dd82801abb856&size=5632&m=image/png&dim=320x218&blurhash=r47A_N?10J9p0cD~x0$.xJ.ONWIV$.x1aiR~NXR$ImR#xbxcxJoNW,W,WT#e$otMEGEHX3xJ$.xJ53I+%3-Z-ZxcNDI,R#ohjcj?bEa_j[WBagay&ox=dfb500ba6f435f7f6bb04f69195e3349a37a38988a0ac9d4b0cd0e51272abbed&alt=`

vitorpamplona:
That's the old nip04 with imeta tags on the url. You have to remove things after the # on the url to process it by extension.

davotoula:
This is in the code, MediaSaverToDisk:133

`saveContentQ(
displayName = File(url).nameWithoutExtension,
`
might need enhancing with substringBeforeFirst("#") or something

Issue #1297, opened 2025-03-16
[BUG] Failed to save DM file when URL contains inline meta-data and has dot(s) towards the end of the long URL


davotoula:
**Describe the bug**
Image URLs in DMs often contain fragments / inline meta data.

Trying to download one can fail if there are dots towards the end of the long URL

**To Reproduce**
1. Send clear text image via DM to yourself
2. Click on the image
3. Click to download the image

**Expected behavior**
The image (or video) should download to phone

**Video and Screenshots**

![Image](https://github.com/user-attachments/assets/317a0a06-5e68-4ca1-8b58-dddc1eba642c)

**Device (please complete the following information):**
 - Phone Brand/Model Pixel 7 Pro]
 - Android Version: 34 and 28
 - App Version: 0.94.3 and main

davotoula:
Verified

PR #1298, opened 2025-03-16
Trim meta data from file name before saving file to disk

davotoula:
Dots can appear in URL fragments such as inline meta-data.

A dot towards the end of a long url will break the usage of nameWithoutExtension which simply does substringBeforeLast(".").

Attempting to save a file using the long URL as file name fails (both on <Q and =>Q) 

This fix removes # and anything after it before saving local file.

Fix for #1297

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
@@ -131,14 +131,14 @@ object MediaSaverToDisk {
                                     }
 
                                 saveContentQ(
-                                    displayName = File(url).nameWithoutExtension,
+                                    displayName = File(trimInlineMetaData(url)).nameWithoutExtension,
                                     contentType = realType,
                                     contentSource = response.body.source(),
                                     contentResolver = context.contentResolver,
                                 )
                             } else {
                                 saveContentDefault(
-                                    fileName = File(url).name,
+                                    fileName = File(trimInlineMetaData(url)).name,
                                     contentSource = response.body.source(),
                                     context = context,
                                 )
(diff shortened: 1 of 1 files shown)


amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
```diff
@@ -255,5 +255,7 @@ object MediaSaverToDisk {
         MediaScannerConnection.scanFile(context, arrayOf(outputFile.toString()), null, null)
     }
 
+    private fun trimInlineMetaData(url: String): String = url.substringBefore("#")
```
KotlinGeekDev, line 258: Please could we not just extract the whole file creation and name trimming process? Something such as `sanitizedFile(name: String) = File(name.substringBefore("#"))`

KotlinGeekDev (commented):
Interesting!

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/MediaSaverToDisk.kt
davotoula: Good comment but I think I would like to keep "inline metadata" in the method name because it's a niche type of sanitization.

I'm not fully sure why this inline meta data is added when uploaded as dm but not when uploaded in feed.

Maybe @vitorpamplona knows... More recent functionality that hasn't been added to feed upload?

vitorpamplona: The inline metadata exists because in NIP-04 we can't use imeta tags. Otherwise, they would be public / not encrypted and anyone would be able to see the links and download the files inside DMs.

davotoula:
For what it's worth... Blurhash can contain a dot. URLEncoder.encode doesn't encode the dot. The dot becomes a problem when in a file name.

Issue #1292, opened 2025-03-12
[BUG] Weird concatenation when sharing a text using android share api


gzuuus:
**Describe the bug**
When a user selects some text, clicks Share, and then selects Amethyst, it creates a concatenation of words using '+' symbols. For example `Enim+do+mollit+adipisicing+veniam+commodo`

**To Reproduce**
1. Select some text in any app
2. Click on share
3. Pick amethyst
4. See the weird concatenation

**Expected behavior**
The text as  it is respecting whitespaces

vitorpamplona:
Hum.. maybe we need to Percent Decode the incoming text ...

davotoula:
Verified

PR #1305, opened 2025-03-25
use existing helper function that URL decodes message text

davotoula:
Message from intent is put into Route as URL encoded and should be URL decoded before used in UI

Fix for #1292


davotoula:
Before:
before

After:
after

Issue #1301, opened 2025-03-18
[BUG] Confusing to remove video from private bookmarks


Dawn11111111:
**Describe the bug**
Video saved to Private Bookmarks from the Media tab have no option available to remove them. Searching a video by noteID also does not surface any action options. Currently, a user must scroll back through the media feed and find the original note to access any action options.

**Expected behavior**
Have action options menu surface on Media tab videos that have been bookmarked or searched by noteID.

davotoula:
Do you have a note id example please?

davotoula:
TL;DR; Should we prevent bookmarking VideoEvent, FileHeaderEvent and possibly others or should we change code to render these events as a note so that menus are available?

Morning @vitorpamplona, the bookmark view is missing the three dot menu because in NoteCompose certain event types (BadgeDefinitionEvent, FileHeaderEvent,FileStorageHeaderEvent, VideoEvent..) are not using the render note method.

Instead they use event specific type renderers like JustVideoDisplay and FileHeaderDisplay.

In the below video I'm just using the note renderer for VideoEvent and FileHeaderEvent and I didn't see any obvious display issues. Where in the app should I check for side effects of using note renderer to display VideoEvent and FileHeaderEvent?

https://github.com/user-attachments/assets/deba9b74-90cb-4bcc-8d10-4806d374aae1

vitorpamplona:
So, those differences exist because of Quoted posts. If the video event is quoted in a kind 1, we made it load as if it was an inline video instead of making a padded quote with the user name and reaction buttons in it. I am not sure if this was the best decision, but it is why it is made like that.

davotoula:
> So, those differences exist because of Quoted posts. If the video event is quoted in a kind 1, we made it load as if it was an inline video instead of making a padded quote with the user name and reaction buttons in it. I am not sure if this was the best decision, but it is why it is made like that.

I see no difference in quoted / boosted posts containing a video when NOT using the JustVideoDisplay.

Creating pull request

PR #1307, opened 2025-03-26
Use CheckNewAndRenderNote method for FileHeaderEvent, FileStorageHeaderEvent and VideoEvent(s)

davotoula:
Skip event specific viewers for the above events and instead use the default CheckNewAndRenderNote.

Bookmarks now have the note wrapper that allows to remove it from bookmarks.

I observe no differences in other areas like Quoted / Boosted video events, media stream, drafts, DMs and etc.

Fix for: #1301


vitorpamplona:
Awesome! Thank you @davotoula

PR #1319, opened 2025-04-12
corrected profile picture border colour

davotoula:
regression: circle around profile pic was not visible because new modifier in theme was accidentally set to `background` rather than `onBackground`.

@vitorpamplona Should we also add a circle around logo in
1. AppDrawer?
2. TopBar?
3. AccountSwitch?


vitorpamplona:
Nice catch! 

> Should we also add a circle around logo inAppDrawer? TopBar? AccountSwitch?

Only if there is a reason to be or if it looks better.

davotoula:
In app drawer makes sense.. Possibly app top bar.

I'll try it out and see

PR #1322, opened 2025-04-17
Prepend clickable text with @ or link emoji

davotoula:
bugfix: fix typo preventing opening of hashtags from markdown notes
prepend clickable user with @ sign
prepend clickable link with link emoji

Before:
links-emoji-before

After:
links-emoji

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ClickableRoute.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ClickableRoute.kt
@@ -282,7 +282,7 @@ private fun DisplayAddress(
 }
 
 @Composable
-public fun DisplayUser(
+fun DisplayUser(
     userHex: HexKey,
     originalNip19: String,
     additionalChars: String?,
@@ -319,7 +319,7 @@ public fun DisplayUser(
 }
 
 @Composable
-public fun RenderUserAsClickableText(
+fun RenderUserAsClickableText(
     baseUser: User,
     additionalChars: String?,
     nav: INav,
@@ -327,7 +327,7 @@ public fun RenderUserAsClickableText(
     val userState by baseUser.live().userMetadataInfo.observeAsState()
 
     CreateClickableTextWithEmoji(
-        clickablePart = userState?.bestName() ?: ("@" + baseUser.pubkeyDisplayHex()),
+        clickablePart = "@" + (userState?.bestName() ?: baseUser.pubkeyDisplayHex()),
         suffix = additionalChars?.ifBlank { null },
         maxLines = 1,
         route = remember(baseUser) { routeFor(baseUser) },
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/markdown/MarkdownMediaRenderer.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/markdown/MarkdownMediaRenderer.kt
@@ -75,11 +75,7 @@ class MarkdownMediaRenderer(
         uri: String,
     ): Boolean =
         if (canPreview && uri.startsWith("http")) {
-            if (title.isNullOrBlank() || title == uri) {
-                true
-            } else {
-                false
-            }
+            title.isNullOrBlank() || title == uri
         } else {
             false
         }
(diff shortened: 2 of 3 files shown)


davotoula:
Maybe a cleaner solution would have been add the change as an option to your [compose-richtext fork](https://github.com/vitorpamplona/compose-richtext)

vitorpamplona:
What if people just put the #, @, and link emoji in the markdown URL title itself? 

I am not sure if there is a good solution for this :(

davotoula:
True.. But I think we are mainly trying to prevent hinding links (to malicious website) pretending to be hashtag/profile. This change will always prepend a link when a link is used.

Hiding a profile click as a link (by adding link emoji) is less severe I think? 

I'm happy to remove the link emoji bit from the pull request. It's a bit ugly anyway 😂

I'll leave the Bugfix in and perhaps the @ for profile (it looks like it might have a been a bug as well).

Let me know

KotlinGeekDev:
> 
> I'm happy to remove the link emoji bit from the pull request. It's a bit ugly anyway 😂
>
IMHO the link emoji is not necessary. :sweat_smile: 
 
> I'll leave the Bugfix in and perhaps the @ for profile (it looks like it might have a been a bug as well).
> 
Yeah, good idea.

KotlinGeekDev:
> What if people just put the #, @, and link emoji in the markdown URL title itself?
> 
> I am not sure if there is a good solution for this :(

Maybe we could check for duplication, some kind of pattern matching. But how feasible it is, is the issue. :sweat_smile:

davotoula:
Fixed bug when clicking hashtag in markdown content
Added @ to user profile name for consistency with non markdown content
No attempt to prevent spoofing URL link as hashtag/profile/

Ready to merge.

KotlinGeekDev (approved):
Looks good to me.

davotoula:
Maybe we can strip any @ and/or # from link descriptions?

vitorpamplona:
Another way would be to keep everything as is, and simply force underline links in markdown. In that way, you will see the https// prefix OR an underlined word when there is a markdown link. 

Underline is just an idea, we could use other formats as well. Whatever works best.

davotoula:
Does the markdown library support any customisation of links? I couldn't find anything.

KotlinGeekDev:
> Does the markdown library support any customisation of links? I couldn't find anything.

I don't think it's necessary to go that far. :sweat_smile: . IMHO you can just try to check  `renderAsCompleteLink` and try to use Compose to show them as underlined. Or try looking at the `RichTextString.Builder` to see if something can be done with it.

davotoula:
I tried underlining but it didn't look as sharp as the current look. Also, hashtags were underlined as well.

I like this approach, strip any # or @ from link labels to prevent spoofing.

Before: 
before

After:
after

Also:

- fixed bug in clicking on hashtag link
- always have @ before profile links

davotoula:
@vitorpamplona ready to be merged...

vitorpamplona:
I will see if I can do something similar to this, but not in the main thread. Running that regex over the entire text of the post while the user is scrolling might create even more hiccups in the UI.

davotoula:
Fair point. Another option would be to extend the markdown Plug-in (haliboo?) to sanitize just link labels.

I think you have a for already of the project.

I will create a new pull request with just the "bug fix" for missing @ for profile links

davotoula:
Closing for a more performant solution. Porting bug fixes to another PR

davotoula:
@vitorpamplona I've modified your fork of com.halilibo.richtext.markdown. How can I publish it (locally) so that Amethyst uses the new version?

vitorpamplona:
You will need to send to your github user and then change the gradle files to load from your github instead of mine. Just make sure to wait until https://jitpack.io updates itself.

vitorpamplona:
/gradle/libs.versions.toml

```
markdown-commonmark = { group = "com.github.vitorpamplona.compose-richtext", name = "richtext-commonmark", version.ref = "markdown" }
markdown-ui = { group = "com.github.vitorpamplona.compose-richtext", name = "richtext-ui", version.ref = "markdown" }
markdown-ui-material3 = { group = "com.github.vitorpamplona.compose-richtext", name = "richtext-ui-material3", version.ref = "markdown" }
```

davotoula:
ooo.. jitpack.io is really handy to include dependencies from forks. Thank you for teaching me that!

PR #1332, opened 2025-04-25
Reduce compiler warnings: amethyst module

davotoula:
- Add @optins
- Change to auto mirrored icons
- Remove unnecessary inline
- Replace deprecated String.capitalize with replaceFirstChar(it.titlecase
- Safe checked casting (where straight forward)

Compiler warnings went down from 143 to 113 (including some of your fixes)

@vitorpamplona Turns out we were both looking at some of the fixes at the same time so it made for some fun merge conflicts 😆

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/DebugUtils.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/DebugUtils.kt
@@ -164,7 +164,7 @@ inline fun <T> logTime(
         block()
     }
 
-inline fun debug(
+fun debug(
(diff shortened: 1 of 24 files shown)


vitorpamplona:
Nice! The migration from the beautiful capitalize to the replaceFirstChar is so ugly, though. I wish they gave us a better option

davotoula:
> Nice! The migration from the beautiful capitalize to the replaceFirstChar is so ugly, though. I wish they gave us a better option 

I agree. The guard makes it even uglier but it's recommended for performance

PR #1371, opened 2025-06-18
Add a UnsignedEventDeserializer

greenart7c3:
Is this good or should I create another event mapper just for the bunker event deserialization?


vitorpamplona:
Hum.. Maybe Event is not the right class to be mapped in the NIP46 signer. What if we used the [EventTemplate](https://github.com/vitorpamplona/amethyst/blob/main/quartz/src/main/java/com/vitorpamplona/quartz/nip01Core/signers/EventTemplate.kt) class? We would just need to make a serializer and a deserializer for this class. It could be similar to the [Rumor](https://github.com/vitorpamplona/amethyst/tree/main/quartz/src/main/java/com/vitorpamplona/quartz/nip59Giftwrap/rumors) class with serializer and deserializer from the nip59 package. 

I assume it doesn't make much sense to use the global event mapper for the signer param, since it doesn't need to be mapped into an Event subclass, right? All you need to do is to fix the defaults in case they don't exist pass it to the signer.

PR #1380, opened 2025-06-29
Add auto translate config to new user specific settings

davotoula:
Well, this took 4 different implementations across 6 branches 😄 

- Added a new user specific settings screen (account specific)
- Drop down shows languages that have been excluded from auto translation (user's spoke languages are excluded). User can delete a language by selecting it and it will be auto translated again
- Use dual icons for User Preferences
- Minor code cleanup

I would appreciate a code review @vitorpamplona , @KotlinGeekDev 
- I'm not using observable state for the drop down, it just uses a copy of existing "don't translate" languages
- Settings are updated in IO scope

https://github.com/user-attachments/assets/d3d69c69-79b2-4fef-8c1f-913e57b8a9b0

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSyncedSettings.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSyncedSettings.kt
@@ -113,6 +113,10 @@ class AccountSyncedSettings(
             security.warnAboutPostsWithReports = syncedSettingsInternal.security.warnAboutPostsWithReports
         }
     }
+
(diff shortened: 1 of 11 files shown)


vitorpamplona:
That's cool, over time we should probably move all these other user settings to the user settings menu item

davotoula:
What are the other user specific settings?

vitorpamplona:
Relays, Media Servers, Security Filters, Privacy Options are user specific settings. But we can work on them later.

KotlinGeekDev (commented):
Cool!

PR #1381, opened 2025-06-30
Propagate Auth event to `sendAndWaitForResponse`

greenart7c3:
- Pass feed types to `sendSingle` when sending signed event


vitorpamplona:
FYI, I am drastically changing how the NostrClient work for outbox. We might need to reengineer this function in the new design later on.

PR #1396, opened 2025-08-03
Dynamic version number based on branch name

davotoula:
I have been testing several feature branches on device and I am always bit confused which feature branch I'm currently testing 😆 This PR should help with testing

- executes a Git command to get the current branch name.
- If the branch is "main" or "master", returns the base version unchanged
- For feature branches, appends the branch name to the version

### DIFF
--- a/amethyst/build.gradle
+++ b/amethyst/build.gradle
@@ -8,6 +8,34 @@ plugins {
     alias(libs.plugins.serialization)
 }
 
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Very cool, but we need to remove this from the release build. Otherwise people will see "main" everywhere.

davotoula:
Main and master are excluded

Issue #1407, opened 2025-08-13
[BUG] [outbox] paid dvm not loading 


davotoula:
**Describe the bug**
paid dvm not loading . Works in release 

**To Reproduce**
1. Go to feed algoritms
2. Click on a paid feed, eg "popular from pubs you don't follow" 
3. Wait
4. See error: not loading 

**Expected behavior**
Zap payment dialogue shows 

**Video and Screenshots**
![share_202466227236879529.jpg](https://github.com/user-attachments/assets/553bd7ab-b4f9-40f8-af3f-68474a859ba8)

**Device (please complete the following information):**
 - Phone Brand/Model [e.g. Pixel 9a
 - Android Version 36
 - App Version outbox 
 - App Flavour Google Play

davotoula:
Possibly related:

`1:50:44.299  W  OK: 086cc7b98c05f0c78fe4a51537006f5780aa12c3b6655e9c478dfa060dfa8ed2 true 
21:50:44.374  W  type=1400 audit(0.0:156726): avc:  denied  { ioctl } for  path="socket:[287292]" dev="sockfs" ino=287292 ioctlcmd=0x894b scontext=u:r:untrusted_app:s0:c201,c256,c512,c768 tcontext=u:r:untrusted_app:s0:c201,c256,c512,c768 tclass=tcp_socket permissive=0 app=com.vitorpamplona.amethyst.debug
21:50:44.508  W  OnFailure null null Broken pipe null
21:50:44.667  W  Relay Closed Subscription EventWatcherSubAssemblerhZu0XB 
21:50:44.701  W  Relay Closed Subscription UserWatcherSubAssembler5YNRaw 
21:50:44.756  W  Relay Closed Subscription EventWatcherSubAssemblerhZu0XB 
21:50:44.767  W  Relay Closed Subscription UserWatcherSubAssembler5YNRaw error: shutting down idle subscription
21:50:44.776  W  Relay Closed Subscription UserReportsSubAssemblerc5yfPB blocked: Request rejected
21:50:44.792  W  Relay Closed Subscription UserReportsSubAssemblerc5yfPB blocked: this is a private relay, only authorized users can read here
21:50:44.820  W  Relay Closed Subscription UserReportsSubAssemblerc5yfPB rate-limited: there is a bug in the client, no one should be making so many requests
21:50:44.837  W  Relay Closed Subscription UserReportsSubAssemblerc5yfPB auth-required: this query requires you to be authenticated
[...]

PR #1408, opened 2025-08-13
Deduplication of code: ZapTheDevsCard, DvmContentDiscoveryScreen

davotoula:
Created ReusableZapButton with common zap button logic
ZapDVMButton refactored to use ReusableZapButton with DVM-specific configuration

Testing:
Zap the devs works
Paid DVM is not loading to to another suspected bug (#1407)  so couldn't test

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ReusableZapButton.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ReusableZapButton.kt
@@ -1,0 +1,327 @@
+/**
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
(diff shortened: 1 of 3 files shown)


amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ReusableZapButton.kt
```diff
@@ -0,0 +1,327 @@
...
+    val onZapComplete: ((Boolean) -> Unit)? = null,
+)
+
+@Composable
+fun ReusableZapButton(
+    baseNote: Note,
+    accountViewModel: AccountViewModel,
+    nav: INav,
+    config: ZapButtonConfig = ZapButtonConfig(),
+    callbacks: ZapButtonCallbacks = ZapButtonCallbacks(),
```
KotlinGeekDev, line 100: Please, what does this replace in the old code? Or is it future insurance? :)

davotoula: button config because multiple places are calling it and need different styles

callback as parameter because I thought I'll be using two different callbacks but never actually used it.. but left it in  😲 

I've manually tested zapping devs and paid Dvm

KotlinGeekDev: I see!

PR #1411, opened 2025-08-16
Reduce duplications: large cache

davotoula:
- Only identical code was extracted to common utility classes. Some more de-duplication could be done but would require rewriting logic.
- Extracted common collectors to an utility class
- Extracted common logic to an interface
- Each cache class keeps reference to its internal data
- Use override for two functions that are used inside common caching logic
- Updated LargeCache and LargeSoftCache to use common collectors and common caching logic

Performance: identical performance since no caching logic was changed. Smaller byte code due to less duplicate code (458 loc removed). 

Testing: testing on device. Please let me know if you'd like to see tests added and run before and after changes.

No profiling done but the caching logic is identical.

### DIFF
--- a/quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
+++ b/quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
@@ -20,7 +20,307 @@
  */
 package com.vitorpamplona.quartz.utils
 
+import com.vitorpamplona.quartz.utils.CacheCollectors.BiFilter
+import com.vitorpamplona.quartz.utils.CacheCollectors.BiMapper
+import com.vitorpamplona.quartz.utils.CacheCollectors.BiMapperNotNull
+import com.vitorpamplona.quartz.utils.CacheCollectors.BiNotNullMapper
+import com.vitorpamplona.quartz.utils.CacheCollectors.BiSumOfLong
 import java.util.function.BiConsumer
+
+class BiFilterCollector<K, V>(
+    private val filter: BiFilter<K, V>,
+) : BiConsumer<K, V> {
+    val results: ArrayList<V> = ArrayList()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        if (filter.filter(k, v)) {
+            results.add(v)
+        }
+    }
+}
+
+class BiFilterUniqueCollector<K, V>(
+    private val filter: BiFilter<K, V>,
+) : BiConsumer<K, V> {
+    val results: HashSet<V> = HashSet()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        if (filter.filter(k, v)) {
+            results.add(v)
+        }
+    }
+}
+
+class BiMapCollector<K, V, R>(
+    private val mapper: BiMapper<K, V, R?>,
+) : BiConsumer<K, V> {
+    val results: ArrayList<R> = ArrayList()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        val result = mapper.map(k, v)
+        if (result != null) {
+            results.add(result)
+        }
+    }
+}
+
+class BiMapUniqueCollector<K, V, R>(
+    private val mapper: BiMapper<K, V, R?>,
+) : BiConsumer<K, V> {
+    val results: HashSet<R> = HashSet()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        val result = mapper.map(k, v)
+        if (result != null) {
+            results.add(result)
+        }
+    }
+}
+
+class BiMapFlattenCollector<K, V, R>(
+    private val mapper: BiMapper<K, V, Collection<R>?>,
+) : BiConsumer<K, V> {
+    val results: ArrayList<R> = ArrayList()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        val result = mapper.map(k, v)
+        if (result != null) {
+            results.addAll(result)
+        }
+    }
+}
+
+class BiMapFlattenUniqueCollector<K, V, R>(
+    private val mapper: BiMapper<K, V, Collection<R>?>,
+) : BiConsumer<K, V> {
+    val results: HashSet<R> = HashSet()
+
+    override fun accept(
+        k: K,
+        v: V,
+    ) {
+        val result = mapper.map(k, v)
+        if (result != null) {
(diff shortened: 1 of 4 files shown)


quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
```diff
@@ -0,0 +1,359 @@
...
+        fun filter(
+            k: K,
+            v: V,
+        ): Boolean
+    }
+
+    class BiFilterCollector<K, V>(
+        val filter: BiFilter<K, V>,
+    ) : BiConsumer<K, V> {
+        var results: ArrayList<V> = ArrayList()
```
KotlinGeekDev: This could be a `val`. Same applies to the `results` variables in this file, unless this causes issues.

KotlinGeekDev (changes requested):
Nice!

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
```diff
@@ -0,0 +1,359 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.vitorpamplona.quartz.utils
+
+import java.util.function.BiConsumer
+
+object CacheCollectors {
```
vitorpamplona: What's the goal of this Object here? Couldn't all the classes just be declared in the file itself?

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
```diff
@@ -0,0 +1,359 @@
...
+            k: K,
+            v: V,
+        ) {
+            if (filter.filter(k, v)) {
+                results.add(v)
+            }
+        }
+    }
+
+    fun interface BiMapper<K, V, R> {
```
vitorpamplona: We could break this in separate files by interface.

quartz/src/main/java/com/vitorpamplona/quartz/utils/LargeCache.kt
```diff
@@ -76,489 +76,7 @@ class LargeCache<K, V> {
...
-    }
-
-    fun <U> associateNotNullWith(transform: (K, V) -> U): Map<K, U> {
-        val runner = BiAssociateNotNullWithCollector(size(), transform)
-        innerForEach(runner)
-        return runner.results
-    }
-
-    private fun innerForEach(runner: BiConsumer<K, V>) {
-        cache.forEach(runner)
```
vitorpamplona, line 198: Just make sure this forEach is the one from ConcurrentSkipListMap and not from the other kotlin wrappers/extensions. The original is almost 2 times faster than the extension functions

quartz/src/main/java/com/vitorpamplona/quartz/utils/LargeCache.kt
```diff
@@ -67,7 +67,7 @@ class LargeCache<K, V> {
         key: K,
         builder: (key: K) -> V,
     ): Boolean {
-        val value = cache.get(key)
+        val value = cache[key]
```
vitorpamplona: Is this coming from an extension? If so, it might be slower.

davotoula: They are identical 

cache[key] is syntatic sugar and more kotlin style 

It gets compiled to .get by compiler

Lint complains about it but if code base is using mainly .get then I'll change it back

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
davotoula: Good catch, fixing it

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
davotoula: I thought it was cleaner to have all collectors inside the object. IDE completion CacheCollectors. would bring up all available collectors.

But I'll lift it out to be more in line with the rest of code base. Imports will be clearer as well.

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
davotoula: Sorry, not sure what you mean, each interface to have its own file? I was planning to just have all interfaces in the same file as well. Just a flat structure together with the classes?

quartz/src/main/java/com/vitorpamplona/quartz/utils/LargeCache.kt
davotoula: yes, looks like it! Using the ConcurrentSkipListMap directly. The signature of ConcurrentSkipListMap.forEach is different from Kotlin forEach. We're using ConcurrentSkipListMap.forEach signature

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
davotoula: Split it into separate files by functional purpose? Eg:

FilterOperations.kt
```
fun interface BiFilter<K, V> { ... }
class BiFilterCollector<K, V>(...) { ... }
class BiFilterUniqueCollector<K, V>(...) { ... }
```

Or by interface type?
BiFilter.kt
```
fun interface BiFilter<K, V> {
    fun filter(k: K, v: V): Boolean
}

class BiFilterCollector<K, V>(...) { ... }
class BiFilterUniqueCollector<K, V>(...) { ... }
```

quartz/src/main/java/com/vitorpamplona/quartz/utils/CacheCollectors.kt
davotoula: Also thank you for the review

PR #1412, opened 2025-08-17
Reduce duplications: hint providers

davotoula:
There is duplicated code in the various hint providers. I was able to identify four patterns that allowed me to extract the most common code into the existing HintProviders interface that the various hint providers can now implement.

514 lines of duplicate code removed. Smaller byte code and also now a single point of change should it be needed.

New hint providers (interfaces) with common logic:
- BasicHintProvider: For simple events extending Event directly
- StandardHintProvider: For threaded events using MarkedETag 
- ExtendedHintProvider: Adds ATag support to StandardHintProvider
- ETagHintProvider: For events using ETag 

Performance: should be identical performance since no logic was changed (expect additional casting to make the logic general). Smaller byte code due to less duplicate code (514 loc removed).

Testing: testing on device for a few days. Please let me know if you'd like to see tests added and run before and after changes.

No profiling done but the logic is identical.

### DIFF
--- a/quartz/src/main/java/com/vitorpamplona/quartz/experimental/zapPolls/PollNoteEvent.kt
+++ b/quartz/src/main/java/com/vitorpamplona/quartz/experimental/zapPolls/PollNoteEvent.kt
@@ -28,24 +28,9 @@ import com.vitorpamplona.quartz.experimental.zapPolls.tags.MinimumTag
 import com.vitorpamplona.quartz.experimental.zapPolls.tags.PollOptionTag
 import com.vitorpamplona.quartz.nip01Core.core.HexKey
 import com.vitorpamplona.quartz.nip01Core.core.TagArrayBuilder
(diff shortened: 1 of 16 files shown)


vitorpamplona:
Hum.. I don't think this one will work. The reason the code is duplicated is because the a, q and e tags change from nip to nip and evolve separately from each other. Unifying them in code was how this was before, but it forces devs to figure out the differences and similarities in order to remove all duplicates. Hints is similar, now they are mostly the same, but I already see the need to customize their parsing from each other. 

Moving forward, I want to remove the ETag-like generic classes and just do... say.. RootTag that is an e tag, but specilized to the needs of the nip, similar to how nip22 does. After all, tags are already re-defined in each nip, even if they have the same name.

davotoula:
I agree with you, I suspected that! 

I don't understand the code fully so I suspected this deduplication may look good "on paper" but convolutes and complicates the actual logic.

Issue #1413, opened 2025-08-17
[BUG] [outbox] Crash when selecting add new emoji for reaction


davotoula:
`Process: com.vitorpamplona.amethyst.debug, PID: 3866
                 java.lang.IllegalArgumentException: Type of the key Address(kind=30030, pubKeyHex=f839fb6714598a7233d09dbd42af82cc9781d0faa57474f1841af90b5c930939, dTag=Modern Family) is not supported. On Android you can only use types which can be stored inside the Bundle.
                 	at androidx.compose.runtime.saveable.SaveableStateHolderImpl.SaveableStateProvider(SaveableStateHolder.kt:74)
                 	at androidx.compose.foundation.lazy.layout.LazySaveableStateHolder.SaveableStateProvider(LazySaveableStateHolder.kt:76)
                 	at androidx.compose.foundation.lazy.layout.LazyLayoutItemContentFactoryKt.SkippableItem-JVlU9Rs(LazyLayoutItemContentFactory.kt:126)
                 	at androidx.compose.foundation.lazy.layout.LazyLayoutItemContentFactoryKt.access$SkippableItem-JVlU9Rs(LazyLayoutItemContentFactory.kt:1)
                 	at androidx.compose.foundation.lazy.layout.LazyLayoutItemContentFactory$CachedItemContent$createContentLambda$1.invoke(LazyLayoutItemContentFactory.kt:95)
                 	at androidx.compose.foundation.lazy.layout.LazyLayoutItemContentFactory$CachedItemContent$createContentLambda$1.invoke(LazyLayoutItemContentFactory.kt:85)
                 	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
                 	at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)`

davotoula:
Looks like a LazyColumn / LazyRow key issue. The `Address` object is not allowed as a key

PR #1414, opened 2025-08-17
Use pubkeyhex for lazy column key instead of Address

davotoula:
fix for #1413


davotoula:
That's an easy change. Let's see what @vitorpamplona suggests!

KotlinGeekDev:
> That's an easy change. Let's see what @vitorpamplona suggests!

Yeah.

PR #1415, opened 2025-08-18
Use explicit field annotation target for constructor params

davotoula:
Kotlin's behaviour for annotation targets is changing. This change adds @field as explicit annotation target.

Only constructor params have been changed. Regular properties are not affected by the behaviour change.

`See https://youtrack.jetbrains.com/issue/KT-73255 for more details.`
`w: file:///Users/david/StudioProjects/amethyst/quartz/src/main/java/com/vitorpamplona/quartz/nip96FileStorage/info/ServerInfo.kt:57:5 This annotation is currently applied to the value parameter only, but in the future it will also be applied to field.`
`- To opt in to applying to both value parameter and field, add '-Xannotation-default-target=param-property' to your compiler arguments.`
`- To keep applying to the value parameter only, use the '@param:' annotation target.`

### DIFF
--- a/quartz/src/androidTest/java/com/vitorpamplona/quartz/nip44Encryption/TestPackageClasses.kt
+++ b/quartz/src/androidTest/java/com/vitorpamplona/quartz/nip44Encryption/TestPackageClasses.kt
@@ -32,61 +32,59 @@ data class V2(
 )
 
 data class Valid(
-    @JsonProperty("get_conversation_key")
-    val getConversationKey: ArrayList<GetConversationKey> = arrayListOf(),
-    @JsonProperty("get_message_keys") val getMessageKeys: GetMessageKeys? = GetMessageKeys(),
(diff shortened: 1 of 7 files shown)


davotoula:
There is a option to add the field declaration to regular properties as well. For consistency and for very rare case a regular property would have multiple annotations.

I chose not to do it to minimise changes...

KotlinGeekDev:
> There is a option to add the field declaration to regular properties as well. For consistency and for very rare case a regular property would have multiple annotations.
> 
> I chose not to do it to minimise changes...

Makes sense; to only apply it where required.

vitorpamplona:
I just realized this might create more issues in the future. I turned Address into a parcelable instance to avoid having to choose.

davotoula:
You mean for the lazy column indexing? Yes, parcelable is a great choice for uniqueness... If the overhead to parcel is small

vitorpamplona:
Yeah, other parts might need to be indexed by address, and frankly that class should be indexable.

PR #1416, opened 2025-08-18
Reduce duplications: tag parsing utility

davotoula:
Extracted common tag parsing patters into separate utility


vitorpamplona:
Sorry, I also don't think this is a good idea. It hides which indexes where tested to exist and then later access random points in the array. It just becomes really hard to see if you have tested the right conditions before calling ...tag[2].

This is more or less how it was before. It was very confusing to understand what was going on with the array indexes. 

Code duplication isn't that bad. I would focus on the performance or ease of use of the code instead.

davotoula:
Makes sense. Working on the code actually gave me good insight!

PR #1417, opened 2025-08-18
Change BasicOkHttpWebSocket to be like the docs

greenart7c3:


vitorpamplona:
Do we want this or does this PR exist just to match the docs? 

I removed the function because I thought it would be easier to use. But we can def move the version this PR suggests.

greenart7c3:
The pr exists just to match the docs, if you want i can fix the docs instead

vitorpamplona:
Let's go for this. But also keep in mind that quartz is quite complicated and needs to expose some easier classes to use.

PR #1435, opened 2025-08-26
Capture and post video

davotoula:
I really like the instant post photo feature but always felt it missed an option to post a video instantly.

- This adds a dialog to select photo (as before) or video (new).
- The UI isn't as slick as I would want to but maybe somebody with better UI skills can improve on it
- I changed the video compression LOW setting to VERY_LOW. LOW results in 8mbit, VERY_LOW results in 4mbit which streams better
- Doesn't work in emulator due to emulator limitations around recording video

Code review? @KotlinGeekDev 

Screen recording:

https://github.com/user-attachments/assets/f7239718-cdfe-4305-b4dc-b01321d6cac5

Screen grab:

Screenshot 2025-08-26 at 12 36 40

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
@@ -84,14 +84,14 @@ class MediaCompressor {
         val videoQuality =
             when (mediaQuality) {
                 CompressorQuality.VERY_LOW -> VideoQuality.VERY_LOW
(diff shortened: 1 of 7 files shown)


KotlinGeekDev (commented):
Interesting! 
A suggestion I have is: why not add a new button for video capture? :) cc @vitorpamplona

vitorpamplona:
Yeah, I new button instead of the popup would be better I think

davotoula:
Ok... Then I can make it a video button. And also not touch previous code.

I was just worried about adding more buttons.

Please try the branch build if you can. Recording video was too easy to add. Worried it could be device dependant.

PR #1440, opened 2025-08-28
Upgrade agp and add JvmTarget.JVM_21 to Quartz

davotoula:
- Upgraded AGP to 8.12.2
- Added JvmTarget.JVM_21 to Quartz because I was getting several errors when trying to push:

`/amethyst/amethyst/src/main/java/com/vitorpamplona/amethyst/service/playback/pip/IntentExtras.kt:48:13 Cannot inline bytecode built with JVM target 23 into bytecode that is being built with JVM target 21. Specify proper '-jvm-target' option.`

I'm using JDK 23... maybe if JDK 21 is used then this issue wasn't noticed.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 accompanistAdaptive = "0.37.3"
 activityCompose = "1.10.1"
-agp = "8.12.1"
+agp = "8.12.2"
 android-compileSdk = "36"
 android-minSdk = "26"
 android-targetSdk = "36"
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Interesting... Should we try to keep it with 1_8, though? Maybe it works. 

As a library, we will need to be more forgiving of who can use it.

davotoula:
Let me try  1_8 on the quartz library. This should be able to inline into the components that use JVM target 21?

davotoula:
That seems to work.

But we have to avoid any java 8+ features in quartz.

PR #1443, opened 2025-08-29
Capture and post video

davotoula:
I really like the instant post photo feature but always felt it missed an option to post a video instantly.

This PR adds a video camera icon next to the photo camera icon in the bottom row actions. It has been added to the following "new post" screens:

- Short note post screen (aka new post)
- New media post
- New Group DM
- New public message (don't think this is active)
- New product
- GeoHashTag post and HashTag post via GenericCommentPostScreen

I have changed the video compression level "low" (20% of original) to "very low" (10% of original) to allow for faster media streaming. Video compression is a complex topic and it would be fun to make it even better at some point!

Code review? @KotlinGeekDev

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
@@ -84,14 +84,15 @@ class MediaCompressor {
         val videoQuality =
             when (mediaQuality) {
                 CompressorQuality.VERY_LOW -> VideoQuality.VERY_LOW
-                CompressorQuality.LOW -> VideoQuality.LOW
+                // Override user selection LOW to use VERY_LOW for better video streaming experience
+                CompressorQuality.LOW -> VideoQuality.VERY_LOW
                 CompressorQuality.MEDIUM -> VideoQuality.MEDIUM
                 CompressorQuality.HIGH -> VideoQuality.HIGH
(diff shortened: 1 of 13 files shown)


davotoula:
By adding the direct posting of video on the media screen, together with using the new vlog kinds... Amethyst is like a vlog client now.

One client to rule them all! 😂

KotlinGeekDev (approved):
Nice! It will be interesting to see what users will think of the video quality. :)

davotoula:
It depends so much on the source material! 

I only changed the low setting to be lower, the others remain the same. 

For me the new setting is better. A 40mbit 4k video now compresses to 4mbit 1080p. It's still a bit high in my opinion but better than the previous 8mbit. 

The compressor also has an option to set bitrate directly. Maybe a more complete solution would be to test for resolution size and set bitrate based on that. Much more manual control that could result in magically better streams. 

Still not ideal because other factors like amount of action also affects the quality. Some codecs have "picture quality" setting but I didn't see that avaliable on our compressor. 

Small steps forward! 


> Nice! It will be interesting to see what users will think of the video quality. :)

KotlinGeekDev:
> It depends so much on the source material!
> 
Exactly.

> I only changed the low setting to be lower, the others remain the same.
> 
> For me the new setting is better. A 40mbit 4k video now compresses to 4mbit 1080p. It's still a bit high in my opinion but better than the previous 8mbit.
>
Yeah, I agree. 

> Small steps forward!
> 
Yeah!

PR #1459, opened 2025-09-05
Setting JvmTarget to 21 for quartz

davotoula:
to fix build/push issue due to mixed Java versions when using Java 23 locally

Similar issue to before. I'm using Java 23 locally and since Quartz didn't have JVM target specified it compiles with 23 and fails to inline with the other 21 code.

### DIFF
--- a/quartz/build.gradle.kts
+++ b/quartz/build.gradle.kts
@@ -152,8 +152,8 @@ kotlin {
                 api(libs.secp256k1.kmp.jni.android)
 
                 // LibSodium for ChaCha encryption (NIP-44)
-                implementation (libs.lazysodium.android)
-                implementation (libs.jna)
+                implementation ("com.goterl:lazysodium-android:5.2.0@aar")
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Didn't 1.8 work?

davotoula:
It seems to have changed with multi platform?

davotoula:
Oh I see.. I think it was accidently reverted with another pull request.

Let me change it to 1.8 and retest

davotoula:
Ok, it works with 1.8 as before and fixes the local build/push issue (when on JVM>21).

@vitorpamplona Should I look into changing our build worker to JVM 23 to catch similar build issues next time with a failing build?

vitorpamplona:
> Should I look into changing our build worker to JVM 23 to catch similar build issues next time with a failing build?

Not sure. I think as we migrate Quartz to a more independent repo we will have to do more tests in the build scripts anyway.

PR #1460, opened 2025-09-07
update secp256k1KmpJniAndroid to 0.19.0

davotoula:
This resolves the last 16KB page size issue in amethyst.

Warning is gone from Android studio.

See https://github.com/ACINQ/secp256k1-kmp/pull/126


vitorpamplona:
Nice!

PR #1468, opened 2025-09-11
Gallery view for multiple images

davotoula:
Introduce an ImageGallery that has custom layouts for low amount of images but uses constrained LazyVerticalGrid with calculated height for >20 images (for performance reasons)

Added text parsing in RichTextViewer to separate text and batches of images that are sent to the new ImageGallery.

Profile with various images notes: npub1r23946z8nu3cjklcl26ujnvkr8e0u0t3e57eqv0y7j0ql2ru39ns2935gp

Test note: nostr:nevent1qqsrayqnhqm9wvyqdpp0ala873jpsxy3j5xap5heta0lg03a2fexx4sppemhxue69uhkummn9ekx7mp0qgsv4pk4t26zd4fdu9ejgy7knwd3kp360xqfdzutd9344zd9v2h3hmqrqsqqqqqpszwm8e

Test note: nostr:nevent1qqsgruz3cxzcdcwxap4adgmz575txq6snt39zet7v5l5tr8h3h2k4vcpzpmhxue69uhkummnw3ezumt0d5hsygx0nv4px6dyekgnycumzwhrjgqdwqcyx9u6s98gvj8fu56qjdlwcypsgqqqqqqselr27n

Test note: nostr:nevent1qqsvudc2fyslzdm8hd4n9g0p40573ks0r49cc5jpdfxafr8cprc89sqprdmhxue69uhhyetvv9ujumn0wd68yurvv438xtnrdakj7q3qf6ugxyxkknket3kkdgu4k0fu74vmshawermkj8d06sz6jts9t4ksxpqqqqqqzvz8ku2

5733148881c14c714f25c9b5a9fd8a4cfa83e5b4b1d9aebb6ec3ed57cda22de0
7b5778b9c735309b321bca80f00212f10f4e7690a54bd29e1caae4e99f7b839f

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/RichTextViewer.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/RichTextViewer.kt
@@ -73,8 +73,6 @@ import com.vitorpamplona.amethyst.commons.richtext.HashTagSegment
 import com.vitorpamplona.amethyst.commons.richtext.ImageSegment
 import com.vitorpamplona.amethyst.commons.richtext.InvoiceSegment
 import com.vitorpamplona.amethyst.commons.richtext.LinkSegment
-import com.vitorpamplona.amethyst.commons.richtext.MediaUrlImage
-import com.vitorpamplona.amethyst.commons.richtext.ParagraphState
 import com.vitorpamplona.amethyst.commons.richtext.PhoneSegment
 import com.vitorpamplona.amethyst.commons.richtext.RegularTextSegment
 import com.vitorpamplona.amethyst.commons.richtext.RichTextViewerState
@@ -113,15 +111,6 @@ import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.toImmutableList
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-
-data class RenderContext(
-    val state: RichTextViewerState,
-    val backgroundColor: MutableState<Color>,
-    val quotesLeft: Int,
-    val callbackUri: String?,
-    val accountViewModel: AccountViewModel,
-    val nav: INav,
-)
 
 fun isMarkdown(content: String): Boolean =
     content.startsWith("> ") ||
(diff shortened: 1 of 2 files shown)


amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
```diff
@@ -0,0 +1,264 @@
...
+        }
+    } else {
+        // Lazy for large sets — expands fully, no independent scroll
+        BoxWithConstraints(modifier = Modifier.fillMaxWidth()) {
+            val totalSpacing = Size5dp * (columns - 1)
+            val imageSize = (maxWidth - totalSpacing) / columns
+            val rows = (images.size + columns - 1) / columns
+            val gridHeight = (imageSize * rows) + (Size5dp * (rows - 1))
+
+            LazyVerticalGrid(
```
vitorpamplona: Did you test this one? my only concern is having a Lazy component inside another Lazy component (compose usually crashes)

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
```diff
@@ -0,0 +1,264 @@
...
+            3 -> ThreeImageGallery(images, accountViewModel, roundedCorner)
+            4 -> FourImageGallery(images, accountViewModel, roundedCorner)
+            else -> ManyImageGallery(images, accountViewModel, roundedCorner)
+        }
+    }
+}
+
+@Composable
+private fun SingleImageGallery(
+    images: ImmutableList<MediaUrlImage>,
```
KotlinGeekDev, line 87: SInce it's a SingleImageGallery, you could just pass in `MediaUrlImage` instead of a list, and move `images.first()` to the `when` statement. :)

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
```diff
@@ -0,0 +1,264 @@
...
+private fun TwoImageGallery(
+    images: ImmutableList<MediaUrlImage>,
+    accountViewModel: AccountViewModel,
+    roundedCorner: Boolean,
+) {
+    Row(
+        modifier = Modifier.aspectRatio(ASPECT_RATIO),
+        horizontalArrangement = Arrangement.spacedBy(IMAGE_SPACING),
+    ) {
+        images.take(2).forEach { image ->
```
KotlinGeekDev, line 111: You could move `images.take(2)` from here to the `when` statement as well.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/RichTextViewer.kt
```diff
@@ -391,41 +548,86 @@ private fun RenderWordWithoutPreview(
...
     }
 }
 
+@Composable
+private fun RenderWordsWithImageGallery(
+    words: ImmutableList<Segment>,
+    context: RenderContext,
+) {
+    var i = 0
+    while (i < words.size) {
```
KotlinGeekDev: Please what does the performance look like( since all these calculations are being done on the UI thread)?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
KotlinGeekDev: > Did you test this one? my only concern is having a Lazy component inside another Lazy component (compose usually crashes)

True, regular `Column` does not like lazy layouts inside it. :sweat_smile:

davotoula: This is used for a rare case of 20+ images. I have tested it in emulator and on device without issues.

The profile npub1r23946z8nu3cjklcl26ujnvkr8e0u0t3e57eqv0y7j0ql2ru39ns2935gp has various image combinations to test. Please test the branch on your device! @vitorpamplona @KotlinGeekDev

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
davotoula: The general `private fun GalleryImage` still needs a list of images (for swiping in zoomed in mode) so we don't save much by doing this change because we have to make other changes to still pass the list

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
davotoula: I think I prefer the consistency of always passing all images and each XImageGallery does the required element extraction. Unless you see a big performance improvement by passing images.take(2) / images.first() directly?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/RichTextViewer.kt
davotoula: I have a minor concern as well about this. The debug version is always a little bit slower than release so hard to see real impact. Let me see if I can optimise it... maybe skip logic if less than 2 images in post?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
davotoula: Actually I removed the special case for 20+ images with nested lazy components to avoid the risk you guys mentioned!

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
KotlinGeekDev: Got it.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
KotlinGeekDev: > I think I prefer the consistency of always passing all images and each XImageGallery does the required element extraction. Unless you see a big performance improvement by passing images.take(2) / images.first() directly?

I see, I see. Since SingleImageGallery is to stay that way, we can keep it then, for consistency.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/RichTextViewer.kt
KotlinGeekDev: > I have a minor concern as well about this. The debug version is always a little bit slower than release so hard to see real impact. Let me see if I can optimise it... maybe skip logic if less than 2 images in post?

Got it. I do wonder if the calculations can be offloaded...And the impact on rendering.

davotoula: I'm rewriting it 😂

KotlinGeekDev: > I'm rewriting it 😂

:joy: Got it.

davotoula:
I will now be measuring execution times before and after changes...

davotoula:
I'm actually seeing quicker rendering. Feel free to try the branch for any bugs or to time it.

Timing code:
```kotlin
Column(modifier = modifier) {
        if (remember(content) { isMarkdown(content) }) {
            RenderContentAsMarkdown(content, tags, canPreview, quotesLeft, backgroundColor, callbackUri, accountViewModel, nav)
        } else {
            val startTime = System.currentTimeMillis()
            RenderRegular(content, tags, canPreview, quotesLeft, backgroundColor, callbackUri, accountViewModel, nav)
            Log.d("RichTextViewer", "Content length: ${content.length}, Time: ${System.currentTimeMillis() - startTime}ms. ")
        }
    }
```

Methodology:
- Start app
- filter for tag:RichTextViewer
- delete log entries (clear entries from home feed)
- go to bookmarks with a few selected test scenarios
- go to test user with the various test scenarios
- save log entries

This was done with main code and then with final Image Gallery code

**Before changes**
Samples: 54
Range: 1 → 44 ms
Mean (avg): ~ 9.3 ms
Median: ~ 6 ms

Outliers:
2404 chars → 44 ms
2699 chars → 43 ms
1395 chars → 22 ms

Typical smaller loads (≤500 chars): 2–10 ms

**After changes**
Samples: 52
Range: 1 → 20 ms
Mean (avg): ~ 6.4 ms
Median: ~ 5 ms
Outliers:

2404 chars → 20 ms
2699 chars → 17 ms
1395 chars → 14 ms

Typical smaller loads (≤500 chars): 2–7 ms

before_after

davotoula:
This is now ready for merge. Personally I think it looks gorgeous with an image gallery rather than a large flow of large images

davotoula:
I guess this could made into an app/user preference to allow for old image rendering as well... Maybe v2 😂

vitorpamplona:
I don't think that is needed. We should just render things in the best way possible

PR #1476, opened 2025-09-16
resolve intermittent CI build failures - fix benchmark apk location

davotoula:
@vitorpamplona  is there a reason why benchmark apk was copied to debug folder?

I ran a few builds using this change and none failed...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,13 +61,13 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: Play Benchmark APK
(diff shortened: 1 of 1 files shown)


davotoula:
Also is there a reason why `run: ./gradlew test --no-daemon` is run twice?

vitorpamplona:
hum... probably just a copy-paste mistake in both cases. There was only debug before.

davotoula:
Then again, the apk has a different name so not sure why it would fail a build.. I'll keep trying

davotoula:
I think we should try it to see whether it reduces failed builds!

PR #1478, opened 2025-09-17
Reduce max requests when in emulator

davotoula:
The current 512 Dispatcher max requests keeps crashing my emulator. This PR is a simple test for whether we are in emulator and if yes set Dispatcher max requests to 32 (this keeps emulator happy).

1. I also noticed that the code is called 3 times on start-up so total max requests from the app are probably closer to 1,500.

```
2025-09-17 11:25:31.291  9165-9165  OkHttpClientFactory     com.vitorpamplona.amethyst.debug     I  Emulator detected, using reduced maxRequests: 32.
2025-09-17 11:25:31.311  9165-9165  OkHttpClientFactory     com.vitorpamplona.amethyst.debug     I  Emulator detected, using reduced maxRequests: 32.
2025-09-17 11:25:31.314  9165-9165  OkHttpClientFactory     com.vitorpamplona.amethyst.debug     I  Emulator detected, using reduced maxRequests: 32.
```

2. I also noticed that we are omitting maxRequestsPerHost so we are using the default 5. This could be low considering a relay could be serving subscribing to timeline + mentions + DMs + profile updates + notifications + other things? Should we consider a higher value @vitorpamplona ?

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/lnurl/LightningAddressResolver.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/lnurl/LightningAddressResolver.kt
@@ -172,7 +172,7 @@ class LightningAddressResolver {
                 val messageNode = tree.get("message")
                 val statusNode = tree.get("status")
 
-                if (tree.get("error").isBoolean && messageNode != null) {
+                if (tree.get("error") != null && tree.get("error").isBoolean && messageNode != null) {
                     if (errorNode.asBoolean()) {
                         return messageNode.asText()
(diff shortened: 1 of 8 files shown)


vitorpamplona:
Maybe just don't set anything for the emulator? I think the default is 64 BTW. 

> maxRequestsPerHost 

That is the number of concurrent connections for the same relay/server domain name. I think 5 is still an ok number.

davotoula:
Changed to use the default (currently 64) for emulator.

Good to hear the default 5 requests/host is sufficient.

Ready to merge!

PR #1480, opened 2025-09-17
Add feed indicator to live bubbles

davotoula:
I really like the new chat bubbles but I'm especially interested if the feed is actually live.

This PR adds a live/offline indicator in the chat bubble.

@KotlinGeekDev would appreciate a code review. Doing network calls from UI so using Dispatchers.IO for asynch.

Checks every 30s for feed status update.

Screenshot_20250916-215752
Screenshot_20250916-202525

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/live/LiveStatusIndicator.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/live/LiveStatusIndicator.kt
@@ -1,0 +1,100 @@
+/**
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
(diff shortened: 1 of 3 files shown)


vitorpamplona:
hum... doesn't the offline channel disappear? It was supposed to. But maybe I was checking only every 15 minutes or so

davotoula:
Ah yes, that could be the case. Maybe it's just a matter of waiting a bit.

I thought the chat lived independent of the live stream so wanted to have an indication whether a stream is live or not.

Let's hold off on this one until further testing?

davotoula:
You can see with the black offline indicator that there are chats with offline stream. I've also often joined the chat to see stream is offline.

vitorpamplona:
hum... Do you think it makes sense to keep the bubble if the stream is offline? My initial thinking was that once the live stream is over, things should just get out of the way... But maybe not...

davotoula:
I'm not sure what's supposed to close the bubble. In theory people could remain in chat talking about the stream. I thought it was something the creator of the chat would do, to close it down...

I've seen intermittent streams often happen when stream owners have technical issues. So any timeout should have a grace period I guess.

What is current logic to close the bubble, is there a period of offline stream? If the stream comes back online, will the chat appear again or must a new one be started?

vitorpamplona:
Yes, it is supposed to reappear. 

Also, we remove 24/7 streams like NoGood music because it is annoying to have that thing in the top at all times.

davotoula:
Ah, I understand it bit more now. 

What is the timeout from a stream is offline to chat disappears? I'll keep testing the current behaviour.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/live/LiveStatusIndicator.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+@Composable
+fun LiveStatusIndicatorForChannel(
+    channel: Channel,
+    accountViewModel: AccountViewModel,
+    modifier: Modifier = Modifier,
+) {
+    var isOnline by remember { mutableStateOf(false) }
```
KotlinGeekDev: You could replace this and the `LaunchedEffect{...}` block below with `produceState()`. 
So, something like:
```kotlin
val isOnline by produceState(false){
 //The code in LaunchedEffect...
value = checkChannelIsOnline(channel, accountViewModel)
delay(30_000)
}
```

KotlinGeekDev: I omitted the `while(true){..}` here because I think it's not needed. But if the behavior is not what is expected, we could add the `while` loop in the `produceState(){..}` block.

KotlinGeekDev (commented):
Nice, nice.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/live/LiveStatusIndicator.kt
davotoula: You are right, produceState is probably a better fit than LaunchedEffect

davotoula:
I have been using this feature for 5 days now. I have to say I like it. Please consider it for inclusion.

Normally it takes some time before a stream is online even though the bubble is there. The new indicator shows me when stream is live so I don't have to keep checking manually.

If stream is offline temporarily the indicator shows it right away and I can wait in main feed until indicator is red again.

I guess the checking could have been done inside view model UI could observe a state flow but I didn't want to touch the view model :-)

KotlinGeekDev (approved):

davotoula:
The bubble will dissappear as usual

PR #1486, opened 2025-09-19
Build optimisations

davotoula:
Done:
Enable parallel builds
CI-optimised memory

Todo:
Add build cache
assembleBenchmark rebuilds same code again?
Split jobs for parallel execution?

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,11 +6,13 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx3g -Dfile.encoding=UTF-8 -XX:MaxMetaspaceSize=1g
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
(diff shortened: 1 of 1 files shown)


vitorpamplona:
> assembleBenchmark rebuilds same code again?

I think it has to, right? the debug build has all the debug symbols. The benchmark build should get rid of all of them

vitorpamplona:
Also, we could add a release build per commit (maybe nightly?), fully signed by the same release keys.

davotoula:
> Also, we could add a release build per commit (maybe nightly?), fully signed by the same release keys.
that would be great, for people to run NIGHTLY but a release version.

davotoula:
> > assembleBenchmark rebuilds same code again?
> 
> I think it has to, right? the debug build has all the debug symbols. The benchmark build should get rid of all of them

I'll look into that.

Do all the flavours need building on each CI build? We don't run automated benchmarks do we?

Maybe debug build would be sufficient for CI and any other can be built locally as and when needed?

vitorpamplona:
> Do all the flavours need building on each CI build? 

I think that is a nice to have. 

> We don't run automated benchmarks do we?

Not yet. There must be a way for us to run benchmarks and enforce certain speeds in some methods. I just don't know how. 

> Maybe debug build would be sufficient for CI and any other can be built locally as and when needed?

Yeah, I think debug is fine. Though I rarely test in real phones on debug because it's hard to see the real performance hiccups that need ot be addressed.

PR #1489, opened 2025-09-23
Smarter video compression

davotoula:
Ready to review... @KotlinGeekDev @greenart7c3 @vitorpamplona 

The main idea is to control precisely what resolution and what bitrate we want during compression.

Added configuration for each different input resolution which will determine downscaling and appropriate bitrate for output:
- HIGH: Resolution unchanged and apply appropriate bitrate.
- MEDIUM: One resolution level lower (eg 4k->1080p) and apply appropriate bitrate.
- LOW: Two resolution levels lower (eg 4k->720p) and apply appropriate bitrate.

Added a workaround for a bug in library when using streamable=true. This will allow for videos that are optimised for streaming (meta data at front).

Added 1.5 multiplier for 60fps videos (industry standard).

Added a test to ignore compressed media if it's larger than original media.

❗The library has a major drawback. It uses Int for bitrate which limits lowest setting to 1 and allow increments in 1 Mbps steps only. Nothing more granular.

The native android MediaCodec supports bitrate in kbps so I'm considering a fork of the library to allow for bitrates in kbps. This will allow for sub 1Mbps compression for low resolution videos and also more granular bitrate control (eg 1.2Mbps).


davotoula:
> Nice changes. Good idea moving the video compressor to another file. Please are the changes to the compression behavior noticeable when the compressed video is played? (since that and other factors like file size will be visible to users)

Before we just let the compressor "auto compress" and it applied size and Bitrate it wanted. 

Now we can actually tell users what different compression levels will do. 

For example Low now produces smaller and faster videos. I've had some feedback that users appreciate small and fast when on mobile. 

I've forked the compressor library and am updating the api to support bps instead of mbps for Bitrate. This will allow videos with Bitrate smaller than 1mbs and also finer Bitrate control. 

Once I've tested it with a new android app we can move amethyst to it and have "world class" video compression 😂

KotlinGeekDev:
I see. Cool!

davotoula:
Personally I dislike streaming raw videos made by phone/dslr on nostr. A dedicated video website is so much better for high quality videos since it encodes the source into various sizes and serves the best for your current network speed (maybe a future Nostr enhancement NIP)

I've had a user help me test this feature on their device
- Uncompressed: 1080p, h265, 16Mbps, 19.2mb
- High: 1080p, h264, 1.9Mbps, 2.3mb (88% saving)
- Medium: 720p, h264, 1.2Mbps, 1.57mb
- Low: 480p, h264, 1.2Mbps, 1.55mb

I can't see much difference in quality on phone to be honest, even 480p looks  good!

For a 1080p input the HIGH setting is great, for 4k the MEDIUM would be better as it lowers the resolution to 1080p which is more than enough for video on Nostr.

I have resurrected the now archived library ([LightCompressor-enhanced](https://github.com/davotoula/LightCompressor-enhanced)). I've already added bitrate parameter in bps for more granular control. I'm hoping to add h265 if possible. We can then switch to 

@vitorpamplona Ready for merge!

vitorpamplona:
With Blossom (as opposed to NIP-96), this idea of the server offering multiple versions is pretty much dead. But we could upload 3 or 4 files to people's servers and tweak the iMeta tag to include multiple resolutions that can be selected by the client on the other side. 

But I think the default link should always be the minimum size that still looks good on new phones.... WhatsApp style.

PR #1491, opened 2025-09-24
Set management UI fixes

KotlinGeekDev:
This tries to address some issues found with the set management UI.
cc @vitorpamplona @davotoula


davotoula:
Nice!

davotoula (approved):

PR #1493, opened 2025-09-24
Show a dialog to select a signer when using multiple signers

greenart7c3:
This fixes a issue where the signer can´t get the client packagename if the user selects the "just once" option in the native android dialog


vitorpamplona:
Nice! thank you @greenart7c3

PR #1508, opened 2025-10-03
Fix sendAndWaitForResponse never receiving a response

greenart7c3:


davotoula:
I've seen ocasional loss of account in amethyst. Could this be a fix for it?

greenart7c3:
No, this is just a function I use in amber and citrine.
It was sending the event before being able to subscribe properly so the function always returned false

vitorpamplona:
Nice, I was going to fix this. My local branch has now new test cases for these functions and NostrClient in general. We need to keep adding tests :)

PR #1512, opened 2025-10-06
BUGFIX: Fix nwc string deleted on restart

davotoula:
The putOrRemove(key: String, event: Any?) signature loses type information. When JsonMapper.toJson() tries to serialise, it sees Any? and can't find the serialiser for Nip47URI, causing:

  `SerializationException: Serializer for class 'Any' is not found`

I've added an overloaded putOrRemove method for Nip47URI.

This fixes the issue for me.

@vitorpamplona, should JsonMapper or OptimizedJsonMapper be used?


vitorpamplona:
JsonMapper should be fine.

PR #1526, opened 2025-10-21
revert spotless plugin  version

davotoula:
(need to find a solution for @Composable funct…ion names)
Remove unused imports

Unable to commit anything due to 100s invalid @Composable function names.

I works putting SUPRESS in all the files but surely it must be possible with single config like 
ktlint_function_naming_ignore_when_annotated_with = Composable

See: https://pinterest.github.io/ktlint/latest/rules/standard/#function-naming


vitorpamplona:
It might not be this, I changed an option yesterday that was supposed to not be needed anymore. I will revert that one as well

PR #1528, opened 2025-10-21
Image gallery improve layout

davotoula:
Aspect ration of first image determines the layout: landscape as first will create landscape thumbnail, portrait first will create portrait thumbnail.

This means two landscape will be stacked on top of each other, two portrait will be next to each other.

3 images will either start with a portrait thumbnail and two images next to it or landscape with two images below depending on the aspect ration of the first image.

4 or more images gallery unchanged as they are square thumbnails only.

Before:

https://github.com/user-attachments/assets/98c5e994-bf39-496f-bb03-b022e503df7f

After:

https://github.com/user-attachments/assets/fe8fe056-44e9-4ff4-8d1f-ea383e27af31

Mean
~29.5 ms
~31.3 ms
⬆️ ~6.1% increase

Median
~23.1 ms
~26.0 ms
⬆️ ~12.5% increase

Screenshot 2025-10-21 at 20 13 05

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,8 @@ trim_trailing_whitespace = true
 max_line_length = 5000
 ij_kotlin_allow_trailing_comma=true
 ij_kotlin_allow_trailing_comma_on_call_site=true
-ktlint_function_naming_ignore_when_annotated_with=Composable, Test
+ktlint_function_naming_ignore_when_annotated_with=Composable
+ktlint_standard_no-unused-imports=enabled
 
 [*.gradle]
 indent_size = 4
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
@@ -86,7 +86,7 @@ class AppModules(
         }
 
     val applicationIOScope = CoroutineScope(Dispatchers.IO + SupervisorJob() + exceptionHandler)
-    val applicationDefaultScope = CoroutineScope(Dispatchers.Default + SupervisorJob() + exceptionHandler)
+    val applicationDefaultScope = CoroutineScope(Dispatchers.IO + SupervisorJob() + exceptionHandler)
 
     // Blocking load of UI Preferences to avoid theme/language blinking
     val uiPrefs by lazy {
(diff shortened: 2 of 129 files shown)


davotoula:
Code review please 🙏

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
```diff
@@ -22,29 +22,98 @@ package com.vitorpamplona.amethyst.ui.components
...
+}
+
+@Composable
+private fun rememberFirstImageOrientation(firstImage: MediaUrlImage?): FirstImageOrientation {
+    val initialOrientation =
+        remember(firstImage) {
+            firstImage.resolveOrientation(ASPECT_RATIO, PORTRAIT_ASPECT_RATIO)
+        }
+
+    var orientation by remember(firstImage) { mutableStateOf(initialOrientation) }
```
KotlinGeekDev: Maybe this could be changed to `val orientation = remember(...` and it's uses updated to `orientation.value = ..`
The behaviour of this change will need to be checked, though.

davotoula: I've done the change but don't fully understand it 😱 what's the reason?

KotlinGeekDev: It is essentially the same thing as before, but without the delegate. And without the delegate, you can use `val` on the property and still be able to mutate state, since it is a `MutableState`. :)

vitorpamplona:
Are these times just for the gallery or for the entire post?

davotoula:
Just for the gallery render. I wanted to measure timings compared to before changes to the gallery!

vitorpamplona:
So, not even including the post parser? 

I can merge this, but we need to figure out how to get these times down, ideally the entire post needs to render in < 16ms

davotoula:
I just wanted to measure difference compared to before the changes, it wasn't a thorough measurement of notes rendering times. 

It's a dev build, maybe a release build performs better?

Let me do some more profiling please!

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ImageGallery.kt
davotoula: i see, that's nice. change done

davotoula:
This PR added negligible execution time (when using dev build) to the already existing functionality. It's ready for merge.

I'll get back to timing the parsing (and image gallery) in a more reliable way. Would benchmark build be suitable for that? Or I can add logging to local release build?

vitorpamplona:
We don't really have a good way to measure the performance of the complete Note composable yet, but there is a RichTextParser in the benchmark module. 

This might be part of a larger refactoring to get caches ready for rendering outside of compose. At least that's how the markdown team upstream has been moving towards. We can also start adding more parsed variables to the Segments. The parsing of the segments themselves could also be faster to avoid spliting strings by `\n` and ` ` and just passing through it.

PR #1529, opened 2025-10-24
Performance Optimizations: Smoother Scrolling and Reduced UI Lag

TekkadanPlays:
Updated some stuff I'm sorry for the AI slop


vitorpamplona:
This is very weird. How much performance are we gaining here?

TekkadanPlays:
I may be retarded, sorry, I suck at Git. and coding.

Can you check the first commit and see if that makes any more sense? There aren't many changes here but this cured the aids on my device.

vitorpamplona:
Just the Prefetching stuff? That was my question because the list is already in memory. In theory, there is no difference between returning 25 or 1000 elements because everything is already loaded anyway. 

If you are just starting, you might try easier changes to the app. Anything related to performance will end up changing a lot and thus require some good analysis to see if the performance is actually better.

TekkadanPlays:
Okay yes I realize I missed some stuff when I made the PR. God I suck at git.. Give me a bit to resolve this

TekkadanPlays:
I tried 😭 

nostr

TekkadanPlays:
Sorry I am fucking retarded

Issue #1530, opened 2025-10-25
Out of memory when uploading a large file, uncompressed


davotoula:
"in large files, we might need to upload directly from disk." 

 I was trying to upload a large video file (250MB) to a blossom server with Amethyst but it looks like it's trying to allocate an array of the size of the video:

 AndroidRuntime: java.lang.OutOfMemoryError: Failed to allocate a 264633896 byte allocation with 100663296 free bytes and 182MB until OOM, target footprint 445729120, growth limit 536870912
10-25 13:42:11.929  4239  4779 E AndroidRuntime: 	at java.util.Arrays.copyOf(Arrays.java:3785)
10-25 13:42:11.929  4239  4779 E AndroidRuntime: 	at java.io.ByteArrayOutputStream.toByteArray(ByteArrayOutputStream.java:211)
10-25 13:42:11.929   	at com.vitorpamplona.amethyst.service.uploads.blossom.BlossomUploader.upload(r8-map-id-64479140e0b870e61e05c85a17a7ae4d7809fefdbe7efda10746b0385d3f1cc2:6)

davotoula:
The main issue is that there is hashing that attempts to the load whole file into memory. This happens before upload (BlossomUploader) and then again after upload when it downloads the file (ImageDownloader) from the server to verify hash is same.

The solution seems to be to calculate the hash in (8K) chunks while streaming the file rather than loading the whole file into memory.

PR #1531, opened 2025-10-25
 Fix OutOfMemoryError for Large File Uploads

davotoula:
Amethyst crashed with OutOfMemoryError when uploading large files (250MB+ videos) to Blossom/NIP96 servers. The app attempted to allocate byte arrays the size of the entire file for SHA256 hash calculation.

Root Causes

1. Pre-Upload Hash (BlossomUploader)

  // OLD: Loaded entire file into memory
  val payload = inputStream.readBytes()  // 250MB byte array
  val hash = sha256(payload).toHexKey()

2. Post-Upload Verification (ImageDownloader)

  // OLD: Downloaded entire file for verification
  Blob(
      huc.inputStream.use { it.readBytes() },  // 250MB byte array
      contentType
  )

Solution: Implemented streaming that processes files in 64KB chunks instead of loading them entirely into memory.

Changes Made

  1. New Utility: CountingInputStream in quartz/src/jvmAndroid
  2. Fixed BlossomUploader to use CountingInputStream instead of loading whole file into memory
  3. Added streaming hashing to ImageDownloader: Added waitAndVerifyStream(). Streams download and calculates hash. Left waitAndGetImage for backwards compatibility
  4. Updated UploadOrchestrator:
     - Uses streaming verification instead of FileHeader.prepare()
     - Manually constructs FileHeader with streamed hash
     - Reuses dimensions/blurhash from upload (no need to recalculate)

Testing:
Tested large file upload blossom to local haven relay. It passed upload hashing and download hashing.
Upload hashing: BufferSize: 8192, stream size: 279730097 bytes, Time to calculate sha256StreamWithCount: 1791ms
Download hashing happens during download so is internet speed dependant.
Tested small images and video which work as before
re-added blurhash to nip96 and blossom

Fixes:  #1530

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
@@ -50,64 +50,54 @@ class ImageDownloader {
      * Stream download and calculate hash for verification without loading entire file into memory.
      * This is memory-efficient for large files (videos, high-res images, etc.)
      */
-    suspend fun waitAndVerifyStream(
-        imageUrl: String,
-        okHttpClient: (url: String) -> OkHttpClient,
-    ): StreamVerification? =
+    private suspend fun <T> retryWithDelay(
+        maxAttempts: Int = 15,
+        delayMs: Long = 1000,
+        operation: suspend () -> T?,
+    ): T? =
         withContext(Dispatchers.IO) {
-            var verification: StreamVerification? = null
+            var result: T? = null
             var tentatives = 0
 
             // Servers are usually not ready, so tries to download it for 15 times/seconds.
-            while (verification == null && tentatives < 15) {
-                verification =
+            while (result == null && tentatives < maxAttempts) {
+                result =
                     try {
-                        tryStreamAndVerify(imageUrl, okHttpClient)
+                        operation()
                     } catch (e: Exception) {
                         if (e is CancellationException) throw e
                         null
                     }
 
-                if (verification == null) {
+                if (result == null) {
                     tentatives++
-                    delay(1000)
+                    delay(delayMs)
                 }
             }
 
-            return@withContext verification
+            return@withContext result
         }
+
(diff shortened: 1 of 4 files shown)


davotoula:
Strange, The 64kb buffer didn't work very well.. Back to 8kb buffer.

I will set up blossom server that accepts the 200mb+ uploads to verify the fix myself.

davotoula:
Ok, verifying the fix was a pain but using a local blossom server I managed. The fix works.

@vitorpamplona Should I look into making a smallest change possible, maybe without touching quartz?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
```diff
@@ -231,34 +231,32 @@ class UploadOrchestrator {
...
+        if (verification != null) {
             updateState(0.8, UploadingState.Hashing)
 
-            val result =
-                FileHeader.prepare(
-                    imageData.bytes,
-                    uploadResult.type ?: localContentType ?: imageData.contentType,
-                    uploadResult.dimension,
+            // Create FileHeader with hash from streaming verification
+            // Note: We skip blurhash/dimensions since we already have them from upload
```
vitorpamplona: Are you sure about this use of blurhash coming from other places? I was looking around and there seems to be places where we only put blurhash if the result of this verifyHeader is filled.

vitorpamplona:
No need to get the smallest thing possible. I think the changes are great. I am just not sure about the blurhash = null thing. It looks like some places still need this

davotoula:
I will double check the blur hash usage!

davotoula:
I see, yes. Lost the blur hash because of streaming hash calculation. Adding it now

davotoula:
Well, that's a lot of changes for the super rare 200mb+ uploads... but I guess it's future proof.

FileHeader still uses hashing in memory but the use cases don't require changing it to streaming hashing IMO.

@vitorpamplona ready for review/merge.

PR #1533, opened 2025-10-27
NIP-59: Remove expiration tags from kind-13 seals

alltheseas:
## Summary

NIP-59 explicitly requires **"Tags MUST always be empty in a kind:13"**. Previously, seals were being created with expiration tags, violating this specification.

## Changes

- Modified `SealedRumorEvent.kt:121-126` to ensure seal tags are always empty
- Removed conditional expiration tag logic from seal creation
- Expiration data remains on the gift wrap (kind 1059) where it belongs per spec

## Impact

Non-compliant seals can be rejected by stricter relays/clients and leak metadata meant to live only in the gift wrap. This fix ensures compliance with NIP-59.

## Testing

- ✅ Unit tests passing
- ✅ Spotless formatting checks
- ✅ Successfully built and installed to Android device

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip59Giftwrap/seals/SealedRumorEvent.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip59Giftwrap/seals/SealedRumorEvent.kt
@@ -24,7 +24,6 @@ import androidx.compose.runtime.Immutable
 import com.vitorpamplona.quartz.nip01Core.core.Event
 import com.vitorpamplona.quartz.nip01Core.core.HexKey
 import com.vitorpamplona.quartz.nip01Core.signers.NostrSigner
-import com.vitorpamplona.quartz.nip40Expiration.ExpirationTag
 import com.vitorpamplona.quartz.nip59Giftwrap.HostStub
 import com.vitorpamplona.quartz.nip59Giftwrap.WrappedEvent
 import com.vitorpamplona.quartz.nip59Giftwrap.rumors.Rumor
@@ -118,12 +117,9 @@ class SealedRumorEvent(
         ): SealedRumorEvent {
             val msg = Rumor.toJson(rumor)
 
-            val tags =
-                expirationDelta?.let {
-                    // minimum expiration is two days in the future due to the random created at
-                    // this will make sure the even arrives and is not deleted because of the 2 days.
-                    arrayOf(ExpirationTag.assemble(createdAt + it + TimeUtils.twoDays()))
-                } ?: emptyArray()
+            // NIP-59: "Tags MUST always be empty in a kind:13"
+            // Expiration belongs on the gift wrap (kind 1059), not the seal.
+            val tags = emptyArray<Array<String>>()
 
             return signer.sign(
                 createdAt = createdAt,


alltheseas:
ran into unrelated test error

```
Bash(./gradlew connectedAndroidTest) timeout: 10m 0s
  ⎿  Error: FAILURE: Build failed with an exception.

     * What went wrong:
     Execution failed for task 
     ':amethyst:connectedFdroidDebugAndroidTest'.
     > There were failing tests. See the report at: 
     file:///home/b/Desktop/develop/amethyst-DM/amethyst/amethyst/build/r
     eports/androidTests/connected/debug/flavors/fdroid/index.html

     * Try:
     > Run with --stacktrace option to get the stack trace.
     > Run with --info or --debug option to get more log output.
     > Run with --scan to get full insights.
     … +196 lines (ctrl+o to see all)

● The connected Android tests ran but there's a test failure in
  ImageUploadTesting > runTestOnDefaultServers. The failure shows:

  expected:<...fc5ee71b32fecdbeb260[]> but
  was:<...fc5ee71b32fecdbeb260[.webp]>

  This appears to be an existing issue with image upload testing
  where a server is returning URLs with .webp extensions when the
  test expects no extension. This failure is unrelated to the NIP-59
   seal tags fix - it's in the Blossom image upload code.

```

related to blossom test error see https://github.com/vitorpamplona/amethyst/pull/1532/commits/4c7ec081f14c55aa0ed9039c5bc5be82dbaa7632

vitorpamplona:
We probably need to change the NIP in this case. The expiration Tag is required on the Seal to avoid being broadcasted by itself to relays.

alltheseas:
>  The expiration Tag is required on the Seal to avoid being broadcasted by itself to relays.

In less technical terms

> The seal:
>  - Exists only in memory during creation
 > - Is immediately encrypted into the gift wrap's content
 > - Never touches the network as an independent event

 > The expiration tag on the GiftWrap (kind:1059) is sufficient because
  that's the only event relays ever see.


Referencing code lines:

```
When we send a NIP‑17 DM, we only ever hand the network layer the outer gift wrap (kind
  1059). See amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt:1463-1497—broadcastPrivately unwraps the seal locally for our own cache,
  but the only client.send call targets the GiftWrapEvent. On the receive side we decrypt the wrap, pull out the sealed rumor, and immediately drop the
  encrypted blob after indexing (amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/DecryptAndIndexProcessor.kt:317-324). There’s no
  path that pushes a bare SealedRumorEvent (kind 13) to relays, so an expiration tag inside the seal isn’t needed to prevent that scenario.

```

vitorpamplona:
In theory, yes. In practice, the Seal, since it is a fully signed event, can leak and just be broadcast to all relays. If that happens, the expiration tag should match the GiftWrap and to make the Seal disappear with the wrap

alltheseas:
@vitorpamplona is it accurate to say you are detecting that there is now unused code, which was previously used?

> SealedRumorEvent.create still advertises an expirationDelta argument, yet after the PR it never influences the tags
    (quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip59Giftwrap/seals/SealedRumorEvent.kt:95). Callers such as NIP17Factory.createWraps keep
    computing and passing a value under the assumption it matters (quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip17Dm/NIP17Factory.kt:70), so
    the silent no-op is confusing. I’d either drop the parameter (and update callers/tests) or document why it must stay for now.

Therefore y'all could consider removing this unused code in the context of seals, NIP-17.

The argument for keeping this code would be that it is used somewhere else in the Ametyhst code base (it is a part of quartz).

An annotation sharing this context without deleting would maybe help clarify things if y'all need the code for other purposes.

```
The only real reason to keep the expirationDelta argument around is API‐compatibility: SealedRumorEvent.create lives in the shared Quartz
  library, so external consumers that call it would need a source/binary update if we drop the parameter today. If you’re comfortable making that
  breaking change (or you haven’t published the signature yet), it’s cleaner to remove the unused argument—and to delete the comment that still hints at
  an expiration path. If you do need to preserve the signature for downstream projects, at least annotate or document that the value is currently ignored
  so callers don’t assume it still drives seal TTL.


```

vitorpamplona:
Sure, but what I am saying is that we should write Seals with expirations to make sure they also disappear even if it leaks (from Amethyst or any other client). I think we should change NIP-59 instead to accept expiration tags

alltheseas:
I see. Let me make some charts for the before and after flows with your proposed nip-59 change. Then we can bring in @staab for nip-59 proposed change review.

alltheseas:
### Flow of proposed PR
  ┌─────────────────────────────────────────────────────────────────────────────┐
  │                    NIP-17 Gift Wrap Flow & Expiration                       │
  └─────────────────────────────────────────────────────────────────────────────┘

  SENDER SIDE - Message Creation
  ═══════════════════════════════════════════════════════════════════════════════

  Step 1: Create Inner Rumor (Unsigned Event)
  ┌─────────────────────────────────────────┐
  │  ChatMessageEvent (kind:14)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: "Hello!"                      │
  │  tags: [["p", "recipient"], ...]        │
  │  sig: "" (UNSIGNED per NIP-17)          │
  │  expiration: 1735689600 (optional)      │
  └─────────────────────────────────────────┘
                      │
                      │ NIP17Factory.createWraps()
                      ↓

  Step 2: Wrap in Seal (kind:13) - NIP-59
  ┌─────────────────────────────────────────┐
  │  SealedRumorEvent (kind:13)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: nip44Encrypt(rumor JSON)      │
  │  tags: [] ← EMPTY per NIP-59 spec       │  ⚠️ NO EXPIRATION HERE
  │  createdAt: randomWithTwoDays()         │     (This was the fix)
  │  sig: schnorr signature                 │
  └─────────────────────────────────────────┘
                      │
                      │ GiftWrapEvent.create()
                      │ seal.toJson() → encrypt
                      ↓

  Step 3: Wrap in GiftWrap (kind:1059) - NIP-59
  ┌─────────────────────────────────────────┐
  │  GiftWrapEvent (kind:1059)              │
  │  ─────────────────────────────          │
  │  pubkey: random ephemeral key           │  🔑 Random key for each recipient
  │  content: nip44Encrypt(seal JSON)       │  📦 Seal is encrypted payload
  │  tags: [                                │
  │    ["p", "recipient_pubkey"],           │
  │    ["expiration", "1735689600"]         │  ✅ EXPIRATION HERE
  │  ]                                      │
  │  createdAt: randomWithTwoDays()         │
  │  sig: schnorr (with ephemeral key)      │
  └─────────────────────────────────────────┘
                      │
                      │ client.send(wrap, relays)
                      ↓

  ═══════════════════════════════════════════════════════════════════════════════
  RELAY - What Gets Broadcasted
  ═══════════════════════════════════════════════════════════════════════════════

  ┌─────────────────────────────────────────┐
  │  RELAY STORAGE                          │
  │                                         │
  │  ✅ Receives: kind:1059 GiftWrap        │
  │     - Has expiration tag                │
  │     - Will auto-delete after expiry     │
  │     - Encrypted content blob            │
  │                                         │
  │  ❌ Never sees: kind:13 Seal            │
  │     - Encrypted inside GiftWrap         │
  │     - Not a separate event              │
  │                                         │
  │  ❌ Never sees: kind:14 Rumor           │
  │     - Encrypted inside Seal             │
  │     - Not signed/broadcasted            │
  └─────────────────────────────────────────┘
                      │
                      │ Recipient queries relay
                      ↓

  ═══════════════════════════════════════════════════════════════════════════════
  RECIPIENT SIDE - Unwrapping
  ═══════════════════════════════════════════════════════════════════════════════

  Step 4: Unwrap GiftWrap
  ┌─────────────────────────────────────────┐
  │  giftWrap.unwrapThrowing(signer)        │
  │                                         │
  │  1. nip44Decrypt(content, ephemeral_pk) │
  │  2. Parse JSON → SealedRumorEvent       │
  └─────────────────────────────────────────┘
                      │
                      ↓

  Step 5: Unseal Rumor
  ┌─────────────────────────────────────────┐
  │  seal.unsealThrowing(signer)            │
  │                                         │
  │  1. nip44Decrypt(content, sender_pk)    │
  │  2. Parse JSON → Rumor                  │
  │  3. Merge → ChatMessageEvent            │
  └─────────────────────────────────────────┘
                      │
                      ↓

  Step 6: Store Without Encrypted Content
  ┌─────────────────────────────────────────┐
  │  LOCAL STORAGE                          │
  │                                         │
  │  GiftWrap: event.copyNoContent()        │  🗑️ Encrypted payload dropped
  │  Seal: event.copyNoContent()            │  🗑️ Encrypted payload dropped
  │  Rumor: Full ChatMessageEvent stored    │  💾 Plaintext message kept
  └─────────────────────────────────────────┘


  ═══════════════════════════════════════════════════════════════════════════════
  EXPIRATION TIMELINE
  ═══════════════════════════════════════════════════════════════════════════════

    Original Rumor                Seal Created           GiftWrap Created
    has expiration?               tags = []              gets expiration tag
          │                            │                         │
          ├─────────────────────────────────────────────────────┤
          │                                                      │
          ↓                                                      ↓
    If rumor has                                         expiration =
    expiration tag:                                      rumorExpiration +
      innerExpDelta = expiration - createdAt             createdAt +
                                                         2 days (safety buffer)
          │                                                      │
          └──────────────────────────────────────────────────────┘
                                (passed but ignored by Seal)
                                (applied to GiftWrap)

  Timeline:
  ─────────────────────────────────────────────────────────────────────►
    T=0              T=48h              T=expiration          Time
    │                 │                      │
    Create            Random createdAt       GiftWrap deleted by relay
    Message           window ends            (if relay honors expiration)
    

  ═══════════════════════════════════════════════════════════════════════════════
  KEY FINDINGS
  ═══════════════════════════════════════════════════════════════════════════════

  ✅ SAFE: Seals NEVER broadcasted independently
  ✅ SAFE: GiftWrap carries expiration (only event relays see)
  ✅ SAFE: Recipients drop encrypted payloads immediately

  ⚠️  CODE SMELL: expirationDelta parameter still passed to SealedRumorEvent.create()
                  but is silently ignored after the fix

  📋 RECOMMENDATION: Deprecate or document the unused parameter

  🎯 CONCLUSION: No NIP change needed - expiration works correctly via GiftWrap

### Before this PR (Amethyst code in production/master branch vs this PR)

  ═══════════════════════════════════════════════════════════════════════════════
                      BEFORE vs AFTER: The Fix Comparison
  ═══════════════════════════════════════════════════════════════════════════════

  BEFORE THE FIX (main branch, old code)
  ───────────────────────────────────────────────────────────────────────────────

  Step 2: Seal Creation (SealedRumorEvent.kt:118-126)
  ┌─────────────────────────────────────────┐
  │  SealedRumorEvent (kind:13)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: nip44Encrypt(rumor JSON)      │
  │  tags: [                                │
  │    ["expiration", "1735689600"]         │  ⚠️ HAD EXPIRATION TAG
  │  ]                                      │     (violates NIP-59)
  │  createdAt: randomWithTwoDays()         │
  │  sig: schnorr signature                 │
  └─────────────────────────────────────────┘
           │
           │ Code that created this:
           ↓
      val tags =
          expirationDelta?.let {
              arrayOf(ExpirationTag.assemble(createdAt + it + TimeUtils.twoDays()))
          } ?: emptyArray()


  AFTER THE FIX (fix-nip59-seal-tags branch, cfcb3f3d9)
  ───────────────────────────────────────────────────────────────────────────────

  Step 2: Seal Creation (SealedRumorEvent.kt:120-122)
  ┌─────────────────────────────────────────┐
  │  SealedRumorEvent (kind:13)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: nip44Encrypt(rumor JSON)      │
  │  tags: []                               │  ✅ EMPTY TAGS
  │                                         │     (complies with NIP-59)
  │  createdAt: randomWithTwoDays()         │
  │  sig: schnorr signature                 │
  └─────────────────────────────────────────┘
           │
           │ Code that creates this:
           ↓
      // NIP-59: "Tags MUST always be empty in a kind:13"
      // Expiration belongs on the gift wrap (kind 1059), not the seal.
      val tags = emptyArray<Array<String>>()


  ═══════════════════════════════════════════════════════════════════════════════
                      WHAT CHANGED: The Diff
  ═══════════════════════════════════════════════════════════════════════════════

  File: quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip59Giftwrap/seals/
        SealedRumorEvent.kt

  - import com.vitorpamplona.quartz.nip40Expiration.ExpirationTag  ❌ Removed

    suspend fun create(...) {
        val msg = Rumor.toJson(rumor)

  -     val tags =                                                  ❌ Removed
  -         expirationDelta?.let {                                  ❌ Removed
  -             arrayOf(ExpirationTag.assemble(...))                ❌ Removed
  -         } ?: emptyArray()                                       ❌ Removed

  +     // NIP-59: "Tags MUST always be empty in a kind:13"        ✅ Added
  +     // Expiration belongs on the gift wrap (kind 1059)...      ✅ Added
  +     val tags = emptyArray<Array<String>>()                     ✅ Added

        return signer.sign(...)
    }


  ═══════════════════════════════════════════════════════════════════════════════
                      IMPACT ANALYSIS
  ═══════════════════════════════════════════════════════════════════════════════

  ┌────────────────────────────┬──────────────────┬──────────────────────────┐
  │         Component          │     BEFORE       │         AFTER            │
  ├────────────────────────────┼──────────────────┼──────────────────────────┤
  │ Seal (kind:13) tags        │ [["expiration"]] │ []                       │
  │ GiftWrap (kind:1059) tags  │ [["p"],["exp"]]  │ [["p"],["exp"]]          │
  │ What relay receives        │ GiftWrap only    │ GiftWrap only            │
  │ Seal broadcasted?          │ NO               │ NO                       │
  │ NIP-59 compliant?          │ ❌ NO            │ ✅ YES                   │
  │ Metadata leak risk?        │ ⚠️ Seal had tags │ ✅ Seal tags empty       │
  └────────────────────────────┴──────────────────┴──────────────────────────┘


  ═══════════════════════════════════════════════════════════════════════════════
                      FUNCTIONAL BEHAVIOR (Unchanged)
  ═══════════════════════════════════════════════════════════════════════════════

  BEFORE and AFTER are functionally identical because:

  1. Relays only see GiftWrap (kind:1059) in both cases ✓
  2. GiftWrap has expiration tag in both cases ✓
  3. Seal is encrypted inside GiftWrap in both cases ✓
  4. Recipients immediately drop encrypted content in both cases ✓

  The fix is purely about SPEC COMPLIANCE, not behavioral change.


  ═══════════════════════════════════════════════════════════════════════════════
                      WHY THE FIX WAS NEEDED
  ═══════════════════════════════════════════════════════════════════════════════

  NIP-59 Specification says:
  ┌───────────────────────────────────────────────────────────────────────────┐
  │ "Tags MUST always be empty in a kind:13"                                  │
  └───────────────────────────────────────────────────────────────────────────┘

  Problems with the old code:
    ❌ Violated explicit NIP-59 requirement
    ❌ Stricter relays/clients might reject non-compliant seals
    ❌ Unnecessary metadata in seal (even though encrypted)
    ❌ Could confuse other implementations reading the spec

  Summary: Your flowchart question revealed an important clarification - I had shown the AFTER state. The key difference is:
  - BEFORE: Seal had [["expiration", "..."]] tags (violated spec)
  - AFTER: Seal has [] empty tags (complies with spec)

  Everything else in the flow remains identical!

### Hypothetical NIP-59 amendment to allow expiration tags on seals

═══════════════════════════════════════════════════════════════════════════════
            HYPOTHETICAL: If NIP-59 Allowed Expiration Tags on Seals
  ═══════════════════════════════════════════════════════════════════════════════

  PROPOSED NIP-59 AMENDMENT
  ───────────────────────────────────────────────────────────────────────────────
  ┌───────────────────────────────────────────────────────────────────────────┐
  │ OLD SPEC: "Tags MUST always be empty in a kind:13"                        │
  │                                                                            │
  │ NEW SPEC: "Tags SHOULD be empty in a kind:13, except for an optional      │
  │            expiration tag for defense-in-depth security"                   │
  │                                                                            │
  │ OR:       "Tags MAY contain only an expiration tag in a kind:13"          │
  └───────────────────────────────────────────────────────────────────────────┘


  SENDER SIDE - Message Creation (WITH AMENDED SPEC)
  ═══════════════════════════════════════════════════════════════════════════════

  Step 1: Create Inner Rumor (Unsigned Event)
  ┌─────────────────────────────────────────┐
  │  ChatMessageEvent (kind:14)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: "Hello!"                      │
  │  tags: [["p", "recipient"], ...]        │
  │  sig: "" (UNSIGNED per NIP-17)          │
  │  expiration: 1735689600 (optional)      │
  └─────────────────────────────────────────┘
                      │
                      │ Calculate innerExpDelta
                      ↓

  Step 2: Wrap in Seal (kind:13) - WITH EXPIRATION
  ┌─────────────────────────────────────────┐
  │  SealedRumorEvent (kind:13)             │
  │  ─────────────────────────────          │
  │  pubkey: sender's real pubkey           │
  │  content: nip44Encrypt(rumor JSON)      │
  │  tags: [                                │
  │    ["expiration", "1735689600"]         │  ✅ EXPIRATION ALLOWED
  │  ]                                      │     (by amended spec)
  │  createdAt: randomWithTwoDays()         │
  │  sig: schnorr signature                 │
  └─────────────────────────────────────────┘
           │
           │ Code would be:
           │
           │  val tags = expirationDelta?.let {
           │      arrayOf(ExpirationTag.assemble(
           │          createdAt + it + TimeUtils.twoDays()
           │      ))
           │  } ?: emptyArray()
           │
           ↓

  Step 3: Wrap in GiftWrap (kind:1059)
  ┌─────────────────────────────────────────┐
  │  GiftWrapEvent (kind:1059)              │
  │  ─────────────────────────────          │
  │  pubkey: random ephemeral key           │
  │  content: nip44Encrypt(seal JSON)       │
  │  tags: [                                │
  │    ["p", "recipient_pubkey"],           │
  │    ["expiration", "1735689600"]         │  ✅ ALSO HAS EXPIRATION
  │  ]                                      │     (redundant but safe)
  │  createdAt: randomWithTwoDays()         │
  │  sig: schnorr (with ephemeral key)      │
  └─────────────────────────────────────────┘
                      │
                      ↓

  ═══════════════════════════════════════════════════════════════════════════════
  DEFENSE-IN-DEPTH: Multiple Expiration Layers
  ═══════════════════════════════════════════════════════════════════════════════

                      ┌─────────────────────────────────┐
                      │   NORMAL OPERATION              │
                      │   (Current Amethyst behavior)   │
                      └─────────────────────────────────┘
                                    │
                                    ↓
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Relay receives: GiftWrap (kind:1059)                                   │
  │  ┌────────────────────────────────────────┐                             │
  │  │ tags: [["expiration", "1735689600"]]   │  ← Primary protection       │
  │  └────────────────────────────────────────┘                             │
  │                                                                          │
  │  Encrypted inside:                                                      │
  │  ┌────────────────────────────────────────┐                             │
  │  │ Seal (kind:13)                         │                             │
  │  │ tags: [["expiration", "1735689600"]]   │  ← Never seen by relay      │
  │  └────────────────────────────────────────┘                             │
  └─────────────────────────────────────────────────────────────────────────┘


                      ┌─────────────────────────────────┐
                      │   EDGE CASE / BUG SCENARIO      │
                      │   (Hypothetical failure mode)   │
                      └─────────────────────────────────┘
                                    │
                                    ↓
  ┌─────────────────────────────────────────────────────────────────────────┐
  │  Scenario: Bug causes bare Seal to be broadcasted                       │
  │                                                                          │
  │  Relay receives: Seal (kind:13)                                         │
  │  ┌────────────────────────────────────────┐                             │
  │  │ tags: [["expiration", "1735689600"]]   │  ← Backup protection! ✅    │
  │  └────────────────────────────────────────┘                             │
  │                                                                          │
  │  Result: Relay will auto-delete after expiry                            │
  │         - Limits exposure window                                        │
  │         - Prevents permanent storage                                    │
  │         - Reduces metadata leak duration                                │
  └─────────────────────────────────────────────────────────────────────────┘


  ═══════════════════════════════════════════════════════════════════════════════
  COMPARISON TABLE: Current vs Amended Spec
  ═══════════════════════════════════════════════════════════════════════════════

  ┌────────────────────────┬─────────────────┬─────────────────┬─────────────────┐
  │      Scenario          │  Current Spec   │  Amended Spec   │   Difference    │
  │                        │  (Empty tags)   │  (Allow expiry) │                 │
  ├────────────────────────┼─────────────────┼─────────────────┼─────────────────┤
  │ Normal operation:      │                 │                 │                 │
  │ - GiftWrap sent        │ ✅ Protected    │ ✅ Protected    │ No difference   │
  │ - Seal encrypted       │ ✅ Hidden       │ ✅ Hidden       │ No difference   │
  │ - Relay deletes        │ ✅ Yes          │ ✅ Yes          │ No difference   │
  ├────────────────────────┼─────────────────┼─────────────────┼─────────────────┤
  │ Bug: Bare seal leaked  │ ❌ Lives        │ ✅ Expires      │ SAFER ✅        │
  │                        │    forever      │    eventually   │                 │
  ├────────────────────────┼─────────────────┼─────────────────┼─────────────────┤
  │ Spec compliance        │ ✅ Compliant    │ ✅ Compliant    │ Both good       │
  ├────────────────────────┼─────────────────┼─────────────────┼─────────────────┤
  │ Implementation burden  │ Simple          │ Simple          │ No difference   │
  ├────────────────────────┼─────────────────┼─────────────────┼─────────────────┤
  │ Metadata in seal       │ Minimal         │ +8 bytes        │ Negligible      │
  │                        │ (tags=[])       │ (expiry tag)    │                 │
  └────────────────────────┴─────────────────┴─────────────────┴─────────────────┘


  ═══════════════════════════════════════════════════════════════════════════════
  CODE CHANGES NEEDED (If Spec Amended)
  ═══════════════════════════════════════════════════════════════════════════════

  File: quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip59Giftwrap/seals/
        SealedRumorEvent.kt

  REVERT to previous implementation:

  + import com.vitorpamplona.quartz.nip40Expiration.ExpirationTag

    suspend fun create(...) {
        val msg = Rumor.toJson(rumor)

  -     // NIP-59: "Tags MUST always be empty in a kind:13"
  -     // Expiration belongs on the gift wrap (kind 1059), not the seal.
  -     val tags = emptyArray<Array<String>>()

  +     // NIP-59 (amended): "Tags MAY contain expiration for defense-in-depth"
  +     // Primary expiration is on gift wrap; seal expiration is backup protection.
  +     val tags =
  +         expirationDelta?.let {
  +             arrayOf(ExpirationTag.assemble(createdAt + it + TimeUtils.twoDays()))
  +         } ?: emptyArray()

        return signer.sign(...)
    }


  ═══════════════════════════════════════════════════════════════════════════════
  RISK ANALYSIS: Why This Amendment Might Make Sense
  ═══════════════════════════════════════════════════════════════════════════════

  ┌───────────────────────────────────────────────────────────────────────────┐
  │ ARGUMENTS FOR AMENDMENT                                                    │
  ├───────────────────────────────────────────────────────────────────────────┤
  │ ✅ Defense-in-depth: Multiple layers of protection                        │
  │ ✅ Bug resilience: Protects against implementation errors                 │
  │ ✅ Future-proof: Handles unforeseen scenarios                             │
  │ ✅ Minimal cost: Only 8 extra bytes per seal                              │
  │ ✅ No breaking change: Old clients still work                             │
  │ ✅ Clear intent: Expiration is security-critical metadata                 │
  └───────────────────────────────────────────────────────────────────────────┘

  ┌───────────────────────────────────────────────────────────────────────────┐
  │ ARGUMENTS AGAINST AMENDMENT                                                │
  ├───────────────────────────────────────────────────────────────────────────┤
  │ ⚠️  Unnecessary: Seals should never be bare in correct implementations    │
  │ ⚠️  Metadata leak: Tags visible if seal parsing bugs exist                │
  │ ⚠️  Spec complexity: Adds exception to simple "empty tags" rule           │
  │ ⚠️  False sense of security: Real fix is "don't broadcast bare seals"     │
  └───────────────────────────────────────────────────────────────────────────┘


  ═══════════════════════════════════════════════════════════════════════════════
  EXPIRATION TIMELINE (With Amendment)
  ═══════════════════════════════════════════════════════════════════════════════

  Time: T=0 (Message sent)
  ─────────────────────────────────────────────────────────────────────────►

    Rumor (kind:14)
    └─ expiration: T+7days
            │
            ├──► Seal (kind:13)
            │    └─ tags: [["expiration", "T+7days+2days"]] ← T+9days
            │                                                   (buffer included)
            │
            └──► GiftWrap (kind:1059)
                 └─ tags: [["expiration", "T+7days+2days"]] ← T+9days
                                                                (same as seal)

    T=0         T+2days      T+7days           T+9days
    │             │            │                 │
    Create        Random       Rumor expires     GiftWrap + Seal expire
    Message       window       (in plaintext)    (relay deletes both)


  If bare Seal leaked at T+1day:
    T+1day                                      T+9days
    │                                             │
    Seal leaked                                   Seal deleted by relay ✅
    (encrypted, but visible)                      (8 day exposure window)


  ═══════════════════════════════════════════════════════════════════════════════
  RECOMMENDATION
  ═══════════════════════════════════════════════════════════════════════════════

  ┌───────────────────────────────────────────────────────────────────────────┐
  │ PROPOSAL: Amend NIP-59 to allow expiration tags on kind:13 seals          │
  │                                                                            │
  │ Rationale:                                                                 │
  │   • Provides defense-in-depth security                                     │
  │   • Minimal implementation cost                                            │
  │   • Protects against unforeseen bugs or protocol evolution                 │
  │   • Expiration is fundamentally security metadata, not content metadata    │
  │                                                                            │
  │ Proposed Spec Language:                                                    │
  │   "Tags SHOULD be empty in a kind:13 seal. An expiration tag MAY be       │
  │    included for defense-in-depth security in case the seal is transmitted  │
  │    outside of a gift wrap."                                                │
  └───────────────────────────────────────────────────────────────────────────┘

  Key Insight: With the amendment, you get redundant expiration protection at both the Seal and GiftWrap layers. In normal operation (which
  Amethyst implements correctly), it makes no functional difference. But if there's ever a bug or future protocol change that causes bare seals to
   be broadcasted, the expiration tag provides a safety net.

staab:
I'm not going to read all that, but I agree with @vitorpamplona, expiration tags (and other things like `-`) are ok. People just need to understand what they're doing and that it makes sense. And yes, expiration should probably be attached to the seal, although that's very marginal.

vitorpamplona:
Closing this since we changed the NIP

PR #1534, opened 2025-10-27
NIP-17: Don't broadcast gift wraps without kind:10050 DM relay list

alltheseas:
## Summary

NIP-17 states: **"If that is not found that indicates the user is not ready to receive messages under this NIP and clients shouldn't try."**

Previously, when a recipient had no kind:10050 DM relay list, Amethyst would fall back to broadcasting gift wraps to their public relays via `computeRelayListForLinkedUser()`. This violates the spec.

## Changes

- Modified `Account.kt:718-721` to remove fallback to public relays
- When no DM inbox relay list exists, gift wrap is simply not sent
- Added comment explaining NIP-17 requirement

## Impact

This fix prevents two problems:

1. **Metadata leak**: Broadcasting gift wraps to public relays reveals the existence of the DM
2. **Delivery failure**: Recipient isn't monitoring public relays for DMs, so message won't arrive

The sender should be informed that the recipient is not ready to receive NIP-17 messages.

## Testing

- ✅ Unit tests passing
- ✅ Spotless formatting checks
- ✅ Successfully built and installed to Android device

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -715,10 +715,11 @@ class Account(
                         ?.ifEmpty { null }
                 if (relayList != null) {
                     client.send(event, relayList.toSet())
-                } else {
-                    val publicRelayList = computeRelayListForLinkedUser(receiver)
-                    client.send(event, publicRelayList)
                 }
+                // NIP-17: "If that is not found that indicates the user is not ready to
+                // receive messages under this NIP and clients shouldn't try."
+                // Do NOT fall back to public relays - this leaks DM metadata and
+                // is unlikely to reach the intended inbox.
             } else {
                 return emptySet()
             }


alltheseas:
@vitorpamplona please note that the above PR does _not_ handle the communication of the fact that DM is not sent in the case not relay list kind-10050 is detected.  Above just resolves the privacy issue.

Do you have guidance on how we might surface this error/unhappy path communication to the DM enjoyer?

davotoula:
Are the 3 PRs independent enhancements that can go in separately or are they dependent on each other?

vitorpamplona:
There is a lot of overlap from the first one. But I think I found the difference. This one has an extra change in Account file to not send GiftWraps when the user hits the Rebroadcast menu option in a post. The issue here is that we don't know if the GiftWrap is a NIP-17 DM or not, so the current code sends it to the DM Inbox AND the public inbox of the receiver to be safe. This only happens when the user forces a rebroadcasting of the event which supposedly only happens when the regular send didn't work. The regular flow only sends to the DM inbox relays already. 

Since this is a failsafe option, I think it is ok to keep sending it to both relays to try to reach the user, regardless of which event kind this GiftWrap has inside.

alltheseas:
> Are the 3 PRs independent enhancements that can go in separately or are they dependent on each other?

Hey @davotoula sorry for the confusion, still getting used to giving y'all the bare minimum of things that "just work", and not jumbling things together.

The extra/non-nip-17 commits were attempts to fix the build pipeline (e.g. a blossom test failed when the nip-17 commits/PRs did not touch this code, as far as I'm aware) in order to build to a connected device directly. 

That said I had trouble compiling to device using the compile to device command. I did get the APK to build, and install but not the former. Maybe I have an older android device that doesn't go well with the build pipeline ( DC-1 ).

alltheseas:
The "don't broadcast gift wrap with out kin:10050 DM relay list" commit is - https://github.com/vitorpamplona/amethyst/pull/1534/commits/8756a9e6377d12e6eb47bb7f8c88a3f21fe6d103. 

Sorry for jumbling things. Let me see if I can strip out all of some of the commits prior to the 10050 relay list commit https://github.com/vitorpamplona/amethyst/pull/1534/commits/8756a9e6377d12e6eb47bb7f8c88a3f21fe6d103.

vitorpamplona:
> The "don't broadcast gift wrap with out kin:10050 DM relay list" commit is - https://github.com/vitorpamplona/amethyst/commit/8756a9e6377d12e6eb47bb7f8c88a3f21fe6d103.

Exactly. But those lines are broadcasting all GiftWraps, not only NIP-17 DMs. All Giftwraps that are not NIP-17 should go to the Public Inbox relay of each user. If the wrap is from somebody else and we cannot decrypt in to know what's inside or the user has rejected the permission to decrypt from the signer, the wrap should then go to both places.

Issue #1535, opened 2025-10-27
[BUG] I see "follow back" for all user accounts 


davotoula:
The bug:
If you add yourself to one of your follow lists then all user profiles will show "follow back" irrespective whether they are following your or not

reproduce:
Create a new list and add yourself to it
Go to any random profile in Global feed

expected:
The user is not following you so label should say "Follow"

error:
The label on random user profile says "Follow back"

Surely not all of nostr is following me.

Seems to be account specific as another account doesn't do this. 

Version: main

davotoula:
Bug was in main
Fix merged to main
done

PR #1540, opened 2025-10-29
only check follow sets when user1 is the logged-in user

davotoula:
If you add yourself to one of your follows lists then all user profiles will show "follow back".

This fix is to check follow sets only if user1 is logged in user (in effect `isUserFollowingLoggedIn` is being called).

Maybe it's time to split up the `observeUserIsFollowing` for those two different cases?

fix for: #1535


vitorpamplona:
Nice

PR #1550, opened 2025-11-06
Reduce errors in log: concurrent modification exception

davotoula:
Defensive copies and synchronised blocks to prevent concurrent modifications.

I feel the main feed UI is snappier after these changes.

Errors fixed:
`java.util.ConcurrentModificationException
	at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1061)
	at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1096)
	at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1093)
	at com.vitorpamplona.amethyst.service.relayClient.reqCommand.user.watchers.UserReportsSubAssembler.findMinimumEOSEsForUsers(UserReportsSubAssembler.kt:166)`

`java.util.ConcurrentModificationException
		at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1061)
		at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:1084)
		at java.util.AbstractCollection.addAll(AbstractCollection.java:335)
		at java.util.LinkedHashSet.<init>(LinkedHashSet.java:182)
		at kotlin.collections.CollectionsKt___CollectionsKt.toMutableSet(_Collections.kt:1710)
		at kotlin.collections.CollectionsKt___CollectionsKt.intersect(_Collections.kt:1687)
		at com.vitorpamplona.amethyst.service.relayClient.reqCommand.user.watchers.UserReportsSubAssembler.groupByRelayPresence(UserReportsSubAssembler.kt:99)`

`	java.util.ConcurrentModificationException
		at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1061)
		at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1096)
		at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1093)
		at com.vitorpamplona.amethyst.service.relayClient.reqCommand.account.follows.AccountFollowsLoaderSubAssembler.updateSubscriptions(AccountFollowsLoaderSubAssembler.kt:247)`

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip51Lists/peopleList/FollowListsState.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip51Lists/peopleList/FollowListsState.kt
@@ -46,7 +46,6 @@ import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.transformLatest
 import kotlinx.coroutines.flow.update
-import kotlin.collections.map
 
 /**
  * Maintains several stateflows for each step in processing PeopleLists
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/UserCompose.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/UserCompose.kt
@@ -34,7 +34,6 @@ import com.vitorpamplona.amethyst.R
 import com.vitorpamplona.amethyst.model.User
 import com.vitorpamplona.amethyst.service.relayClient.reqCommand.account.observeAccountIsHiddenUser
 import com.vitorpamplona.amethyst.service.relayClient.reqCommand.user.observeUserIsFollowing
-import com.vitorpamplona.amethyst.ui.navigation.navs.EmptyNav.nav
 import com.vitorpamplona.amethyst.ui.navigation.navs.INav
 import com.vitorpamplona.amethyst.ui.navigation.routes.Route
 import com.vitorpamplona.amethyst.ui.navigation.routes.routeFor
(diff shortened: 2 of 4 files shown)


vitorpamplona:
Very interesting. 

FYI, I usually try to do everything to avoid synchronized because they block the thread underneath the coroutine. If there is too much blocking  running in the Main thread or dispatcher, even for simple/fast code, even in a coroutine, the UI will suffer because Main uses just one thread and the thread itself will be blocked not only the coroutine. The same issue happens with the Default dispatcher, since it only has a few threads that match the number of CPUs. Too many blocks on the Default thread pool reduces the performance of everything. That won't happen in IO dispatchers because the threadpool there is flexible and if everything is blocked a new thread is made. This issue happens in all syncronized implementations in Kotlin or any of the Java dependencies. 

Also, fun fact, LruCache also blocks inside of it. So, when we call it from Main, there could be hickups that are significant. We might need to avoid using it on the Main Dispatcher too. :( 

I hate that we never know if a library is using syncronized or not inside their codebases.

davotoula:
I agree with hating synchronised... optimistic unsynchronised is a good start until we see concurrent modification exceptions in logs 🤣

The ui felt smoother after the fix, maybe avoiding the CME causes more stable ui?

But it was just a "feeling", didn't profile it nor instrument it.

davotoula:
Oh, talking about a blocking lru cache, I see significant ui staggering on empty cache startup.

I don't see it on release version. Something to watch

vitorpamplona:
Yeah, all of the debugging bloat is synchronized. :)

PR #1555, opened 2025-11-10
Support for nip 46

greenart7c3:


greenart7c3:
@vitorpamplona what should i do in the connect method?

after the connect whoever is using this needs to also make the getPublicKey since the remoteKey can be different from the real user key

Should I make the getPublicKey inside the connect and return both pubkeys?

vitorpamplona:
I am not sure. We could stay low-level on quartz and require users to request it for now. Then build a helper function later (via extension?) to facilitate. 

I think we will need to use the user's key as a reference to save in the local db on Amethyst. But maybe other systems don't have such needs.

vitorpamplona:
Hey greenart, let's avoid big PRs if possible. You can merge these structures if they work and change them later if you need.

greenart7c3:
Sure, you can merge this

PR #1586, opened 2025-11-21
Bookmark Lists.

KotlinGeekDev:
This PR introduces support for bookmark lists, also known as bookmark groups.


vitorpamplona:
This is great. 

1. Just prefer building routes instead of dialogs.
2. I think we will remove "t" (hashtags), "r" (URLs) from 30003 because there is already an interest list (30015) for hashtags and `r` is being phased out in favor of NIP-73. So maybe don't even bother coding support for them. Just display any `e` or `a` event from the list.

vitorpamplona:
Also, let's keep PRs small. When you get these commits in a place you like, mark ready for review, I will merge them and you can change it later if necessary.

KotlinGeekDev:
> This is great.
> 
>     1. Just prefer building routes instead of dialogs.
> 
>     2. I think we will remove "t" (hashtags), "r" (URLs) from 30003 because there is already an interest list (30015) for hashtags and `r` is being phased out in favor of NIP-73. So maybe don't even bother coding support for them. Just display any `e` or `a` event from the list.

Got it. I was already thinking of making support for links and hashtags. I'll remove them then.

KotlinGeekDev:
OK, Done. A video of how it works:

https://github.com/user-attachments/assets/5fefb0e3-412d-4c3e-8aa7-0f682eb9b3c6

KotlinGeekDev:
@vitorpamplona PR is ready.

PR #1602, opened 2025-12-13
Add iosX64 target

DestBro:
Did not check if it compiles


DestBro:
Seems it compiles fine now

PR #1609, opened 2025-12-15
Create KIND 1222 (voice notes) in amethyst

davotoula:
- Voice message posting flow added with recording UI, upload progress, and guarded posting
- No text entry allowed for voice messages
- Waveform preview
- Media server selection
- Deleting of temporary voice files on success / cancel 
- Abort post on upload failure
- Synchronised cleanup, dedicated scope, and exception handling for short/invalid recordings

https://github.com/user-attachments/assets/d34a6599-09ad-4e2d-aee9-e67ed0bcb18c

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/ShortNotePostViewModel.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/ShortNotePostViewModel.kt
@@ -923,6 +923,8 @@ open class ShortNotePostViewModel :
     }
 
     fun selectVoiceRecording(recording: RecordingResult) {
+        // Delete any existing temp file before replacing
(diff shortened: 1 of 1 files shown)


davotoula:
@vitorpamplona @greenart7c3 @KotlinGeekDev Ready for review / merge

vitorpamplona:
Very cool.

davotoula:
Let's make nostr chatty! 😂

vitorpamplona:
FYI, we may need to change the reply (click and hold to record) logic. I don't think people like this click and hold idea and it doesn't offer a way to check before sending it. Maybe we just open a new reply screen specifically made for audio posts with a widget to record/check/choose server, etc. 

The screen should be custom since YakBaks don't accept text replies. So.. more work :)

davotoula:
I agree, no preview and a silent UI while reply is being uploaded is confusing.

I can have a look at that next

PR #1624, opened 2025-12-27
Quartz Feature Parity Table on Multiplatform

DestBro:
This info is important for any new developer checking this repository. Based on the investigation of `TODO("Not yet implemented")` markers across the codebase, there is a significant gap in feature parity for the iOS implementation of the `quartz` library. While the Android and JVM versions are largely complete, many core cryptographic, utility, and data management features are currently stubs in the iOS target.

#### Detailed Observations

1.  **Core Cryptography Gap:** The most critical missing piece for iOS is the `Secp256k1Instance`. Without this, the library cannot perform the fundamental tasks of a Nostr client, such as signing events or verifying signatures.
2.  **Serialization:** The `OptimizedJsonMapper` on iOS is currently non-functional. Since Nostr relies heavily on JSON for event transmission, this prevents any real-world usage on iOS.
3.  **Data Structures:** The `LargeCache` and `BitSet` implementations on iOS are essentially empty. This will likely lead to performance issues or crashes if the common code expects these to behave as they do on Android.
4.  **Incomplete Common Code:** Some "Not yet implemented" markers exist in `commonMain` (e.g., `NostrSignerRemote`, `BitcoinExplorer`), indicating features that are currently under development or stubbed out for all platforms, not just iOS.
5.  **Amethyst UI:** There is an `EmptyNav` in the Amethyst application that has a `TODO` for `navigationScope`, suggesting some navigation paths are still being drafted.
    
#### Conclusion
The iOS target is currently in an early stage of development and is not yet ready for production use, as it lacks the essential cryptographic and serialization primitives required for Nostr protocol operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,6 +369,28 @@ When your app goes to the background, you can use NostrClient's `connect` and `d
 methods to stop all communication to relays. Add the `connect` to your `onResume` and `disconnect`
 to `onPause` methods.
 
+### Feature Parity Table
+
+| Feature Category | Feature / Component | Android / JVM Support | iOS Support | Notes |
+| :--- | :--- | :---: | :---: | :--- |
+| **Cryptography** | Secp256k1 (Schnorr, Keys) | ✅ Full | ❌ No | Core Nostr signing/verification is missing on iOS. |
+| | LibSodium (ChaCha20, Poly1305) | ✅ Full | ❌ No | AEAD and stream ciphers are unimplemented. |
+| | AES Encryption (CBC & GCM) | ✅ Full | ❌ No | `AESCBC` and `AESGCM` are stubs on iOS. |
+| | Hashing (SHA-256, etc.) | ✅ Full | ❌ No | `DigestInstance` is unimplemented. |
+| | MAC (HmacSHA256, etc.) | ✅ Full | ❌ No | `MacInstance` is unimplemented. |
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Nice add on

DestBro:
> Nice add on

Small thing I wanted to flag: when I send refactoring suggestions in DMs and there’s no reply, I’m not sure how to interpret it. A quick response would be appreciated.

vitorpamplona:
Weird. I did reply. But I am not sure if your DM inbox relays are accepting DMs. You have alby there that def doesn't accept DMs. The other /lang/zh might not accept them too. Ditto is very spotty and sometimes doesn't save events at all if it is under load. See if resetting to the default DM relays helps.

You should also remove purplepages from your outbox/inbox list since it doesn't keep your content or receive notifications to you.

DestBro:
Oh, you are right, that's crazy, well that thing adds more complexity for new comers to Nostr - they should know what stands behind all these relays in part. Also that there are "Content" relays and "Private messages" relays, and "Search" relays - whyle my general understanding was that each relay is the same and/or tends to cover as much NIPs as possible.

DestBro:
I fixed the realys, should work now, can you somehow trigger a message sync ?

DestBro:
I added amethyst relay as well. I expected that meassages will appear, but they did not, seems like you posted them only to these relays that do not work

vitorpamplona:
Hum.. amethyst doesn't have any relay.. 🤔

DestBro:
grafik

PR #1672, opened 2026-01-15
Voice notes enhancements and code review fixes

davotoula:
- Limit voice recording to 180s (to avoid hypothetical crashes during recording and/or anonymisation) 
- Moved re-record button inline and added to VoiceMessagePreview

Minor:
  - remove unused imports
  - sonar fixes
  - Do something with the "Boolean" value returned by "delete".
  - preallocate list capacity
  - Moved processingPreset = preset to execute synchronously before launching the coroutine
  - use colorScheme.onSurfaceVariant instead of Gray
  - guard for null inputFile.parentFile
  - add docs
  - confirmed: <1 results in Higher pitch. Adjusted NEUTRAL to slightly lower pitch
  - guard encoder / decoder .stop

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/types/Nip.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/types/Nip.kt
@@ -20,8 +20,6 @@
  */
 package com.vitorpamplona.amethyst.ui.note.types
 
-import android.R.attr.label
-import android.R.attr.maxLines
 import androidx.compose.foundation.border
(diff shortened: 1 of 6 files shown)


davotoula:
@vitorpamplona the 3 minute limit is just a safe-guard to avoid hypothetical out of memory issues since the recording / processing is not done streaming.

Do you think the limit is unnecessary?

vitorpamplona:
3 minutes may be too short. Maybe 10 minutes? Some people really like to talk

davotoula:
Ok.. We'll watch out for any oom and I'll change it to streaming if needed!

PR #1693, opened 2026-01-30
fix crash in Jackson deserialization for empty Filters

greenart7c3:


quartz/src/jvmAndroidTest/kotlin/com/vitorpamplona/quartz/nip01Core/jackson/JacksonMapperTest.kt
```diff
@@ -152,4 +153,25 @@ class JacksonMapperTest {
...
+        val deserialized = JacksonMapper.fromJsonTo<Filter>(json)
+
+        assertEquals(null, deserialized.ids)
+    }
+
+    @Test
+    fun shouldNotThrowExceptionWhenDeserializingFilterTags() {
+        val json =
+            Filter(
+                tags = mapOf("p" to listOf("123")),
```
davotoula: Consider a valid  64‑char hex key to avoid warning in test logs. Eg 3c39a7b53dec9ac85acf08b267637a9841e6df7b7b0f5e2ac56a8cf107de37da

quartz/src/jvmAndroidTest/kotlin/com/vitorpamplona/quartz/nip01Core/jackson/JacksonMapperTest.kt
```diff
@@ -152,4 +153,25 @@ class JacksonMapperTest {
...
+                tagsAll = mapOf("p" to listOf("123")),
+            ).toJson()
+        val deserialized = JacksonMapper.fromJsonTo<Filter>(json)
+
+        assertEquals(true, deserialized.tags?.keys?.contains("p"))
+        assertEquals(true, deserialized.tagsAll?.keys?.contains("p"))
```
davotoula, line 177: Consider testing also that the deserialised value is same as original value:
assertEquals(listOf(expectedTagValue), deserialized.tags?.get("p"))

PR #1726, opened 2026-03-01
Fix image/URL display when no space before URL

### LINKED ISSUE #1149: [BUG] Amethyst fails to display images if there is no space before "https://"
One common reason images fail to load is the syntax created by one client being interpreted incorrectly by another.

In this case, images uploaded by Primal are breaking on Amethyst because the client isn't adding an extra space or line break before the URL.

Most other clients can display these images normally, but Amethyst does not, only showing the image URL instead. 

I'm asking for Amethyst to follow what other clients already do:  recognize the user intent to display an image, and parse these links correctly. 

Examples:

https://nostrudel.ninja/#/n/nevent1qqs2j3p0lxyw7xnjhc2j55vl63ael0mfkdxdyqc479tuvdy4zfqf3uqpzfmhxue69uhhqatjwpkx2urpvuhx2ucpz3mhxue69uhhyetvv9ujuerpd46hxtnfduq3camnwvaz7tmwdaehgu3wvf5hgcm0d9hx2u3wwdhkx6tpdsqs6amnwvaz7tmwdaejumr0dsl7jwdl

https://iris.to/nevent1qqs2j3p0lxyw7xnjhc2j55vl63ael0mfkdxdyqc479tuvdy4zfqf3uqpzfmhxue69uhhqatjwpkx2urpvuhx2ucpz3mhxue69uhhyetvv9ujuerpd46hxtnfduq3camnwvaz7tmwdaehgu3wvf5hgcm0d9hx2u3wwdhkx6tpdsqs6amnwvaz7tmwdaejumr0dsl7jwdl

https://coracle.social/nevent1qqs2j3p0lxyw7xnjhc2j55vl63ael0mfkdxdyqc479tuvdy4zfqf3uqpzfmhxue69uhhqatjwpkx2urpvuhx2ucpz3mhxue69uhhyetvv9ujuerpd46hxtnfduq3camnwvaz7tmwdaehgu3wvf5hgcm0d9hx2u3wwdhkx6tpdsqs6amnwvaz7tmwdaejumr0dsl7jwdl

image

dmnyc:
Closes #1149

## Summary

- Fixes images and URLs not rendering when there is no space before the URL scheme (e.g., `image!https://example.com/photo.jpg`)
- The rich text parser's URL detector (LinkedIn's library) correctly finds URLs even without preceding spaces, but the text segmenter splits by spaces and then does exact word matching against the detected URL set — so `"image!https://..."` never matched `"https://..."`
- Fix: when a space-split word contains a detected URL but isn't an exact match, split it into prefix text + URL/image segment + suffix text

## Screenshots

Before:
image
image

After:
image

## Test plan

- [ ] Post a note with no space before an image URL (e.g., `text!https://image.nostr.build/...gif`)
- [ ] Verify the image renders correctly in the thread view
- [ ] Verify normal posts with properly spaced URLs still render correctly
- [ ] Verify URLs with text after them also split correctly

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
@@ -178,7 +178,13 @@ class RichTextParser {
             val wordList = paragraph.trimEnd().split(' ')
             val segments = ArrayList<Segment>(wordList.size)
             wordList.forEach { word ->
-                segments.add(wordIdentifier(word, images, videos, urls, emojis, tags))
+                val schemeIndex = findUrlSchemeIndex(word)
+                if (schemeIndex > 0) {
+                    segments.add(RegularTextSegment(word.substring(0, schemeIndex)))
+                    segments.add(wordIdentifier(word.substring(schemeIndex), images, videos, urls, emojis, tags))
+                } else {
+                    segments.add(wordIdentifier(word, images, videos, urls, emojis, tags))
+                }
             }
 
             paragraphSegments.add(ParagraphState(segments.toPersistentList(), isRTL))
@@ -197,6 +203,16 @@ class RichTextParser {
                     )
                 }
             }.toImmutableList()
+    }
+
+    private fun findUrlSchemeIndex(word: String): Int {
+        val https = word.indexOf("https://")
+        val http = word.indexOf("http://")
+        return when {
+            https > 0 && (http <= 0 || https <= http) -> https
+            http > 0 -> http
+            else -> -1
+        }
     }
 
     private fun isNumber(word: String) = numberPattern.matches(word)


vitorpamplona:
Too slow. It compares every word in the post with all urls.. need to figure out a faster way of doing this.

dmnyc:
Updated — replaced the O(words × urls) approach with a simple `indexOf` check for `https://` / `http://` schemes mid-word. No iteration over the URL set at all, just two constant-time string searches per word. When a scheme is found at position > 0, the word is split at that index and the URL portion is passed through the existing `wordIdentifier`. Tested and confirmed working.

*This fix was developed with [Claude Code](https://claude.ai/claude-code).*

vitorpamplona:
Still slow. To be fair, I think I need to completely change how the parser is done so that each character is only tested once for all the 12 or so different kinds of things it could be. 

This is happening during scrolling. That's why performance is so important. 

I rejected two other Claude PRs on this big refactoring already. I am working on it.

dmnyc:
Makes sense, context appreciated. Happy to close this in favor of your rewrite.

vitorpamplona:
I just merged a commit that brings in-house the UrlDetector we were using as a lib. I made a lot of changes and added a solution to this issue directly in the main parser. I hope it solves all the issues there.

PR #1727, opened 2026-03-01
fix: treat multibyte characters as URL terminators in RichTextParser

kojira:
### Problem

On Android, the ICU regex engine treats `\w` as Unicode-aware, causing it to
match multibyte characters (e.g. Japanese, Arabic, Korean). This means that
when a URL or email address is embedded in text without surrounding spaces,
multibyte characters are incorrectly included as part of the URL/email.

The wrong range is detected as a link, so it cannot be opened correctly.
Additionally, when the trailing text is long enough, the entire segment is
rendered as a truncated link, making the surrounding text completely unreadable.

For example, `ほむほむaaaaa@example.com` or `example.comあああ` is parsed as
a single URL segment, displayed as a link that includes the multibyte characters.

### Changes

- **`noProtocolUrlValidator`**: Replace `\w`/`\d` with explicit ASCII character
  classes `[a-zA-Z0-9_]`. This also removes the Android 9 try/catch workaround,
  which is no longer needed with the simplified regex.

- **`wordBoundaryRegex` + split change**: Add a compiled regex that splits text
  at spaces *and* at ASCII/multibyte character boundaries. As a result,
  `ああexample.com` is split into `["ああ", "example.com"]` before segment
  parsing, so only the ASCII URL part is recognized as a link.

## Test plan

- [ ] `ああexample.com` → `ああ` (text) + `example.com` (link)
- [ ] `example.comあああ` → `example.com` (link) + `あああ` (text)
- [ ] `ほむほむaaaaa@example.comほげほげ` → text + email + text
- [ ] Unit tests in `RichTextParserMultibyteTest` pass

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
@@ -175,7 +175,7 @@ class RichTextParser {
         lines.forEach { paragraph ->
             val isRTL = isArabic(paragraph)
 
-            val wordList = paragraph.trimEnd().split(' ')
+            val wordList = paragraph.trimEnd().split(wordBoundaryRegex).filter { it.isNotEmpty() }
             val segments = ArrayList<Segment>(wordList.size)
             wordList.forEach { word ->
                 segments.add(wordIdentifier(word, images, videos, urls, emojis, tags))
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Nice! Let me know when you are ready for a merge. 

Please add tests as we are likely to migrate this parser later for something that only walks through each character once to improve performance.

vitorpamplona:
I am reconsidering this because it is adding spaces when the parser rebundles them into a single text. 

It was breaking in Latin 1 chars that are used in the middle of words, like "caução". I fixed that one by adding ASCII + Latin1. But then quotes ` ’ “ ” ` and other punctuation started to have spaces. 

I am also concerned about non-ASCII URLs, like: http://example.com/スター・ウォーズ×マンガ/  which are now possible. 

Maybe we need a new parser.

vitorpamplona:
@kojira I just merged a commit that brings in-house the UrlDetector we were using as a lib. I made a lot of changes and added your approach (ascii vs non-ascii domains) directly into the main parser. It looks like it still works with your changes but let me know if I broke anything in japanese urls. 

https://github.com/vitorpamplona/amethyst/commit/f4d401bc56d8c284e1e857642aff8bd33accce52

kojira:
Thank you for merging this! I verified it works correctly on device — URLs are properly detected and linked even when surrounded by multibyte characters on both sides. The fix behaves exactly as expected. Really appreciate the quick review and merge! 🙏

PR #1734, opened 2026-03-02
feat: Namecoin NIP-05 identity verification via ElectrumX

mstrofnone:
## Namecoin NIP-05 Identity Verification via ElectrumX

Add censorship-resistant NIP-05 verification and **search resolution** using the Namecoin blockchain. Users can set their `nip05` field to a `.bit` domain (e.g. `alice@example.bit`) or direct Namecoin name (`d/example`, `id/alice`) and Amethyst will resolve the pubkey mapping via ElectrumX instead of HTTP.

### What's new

**NIP-05 verification** — profiles with `.bit` in their `nip05` field are verified against the Namecoin blockchain instead of HTTP. No web server to seize, no DNS to hijack.

**Search bar integration** — typing any Namecoin identifier into the search bar resolves it via ElectrumX and shows the user profile:

| Search query | What happens |
|---|---|
| `alice@example.bit` | Looks up `d/example`, resolves `alice` entry |
| `example.bit` | Looks up `d/example`, resolves root `_` entry |
| `d/example` | Direct domain namespace lookup |
| `id/alice` | Direct identity namespace lookup |

**Tor integration** — ElectrumX connections respect the user's Tor settings. When "NIP-05 verifications via Tor" is enabled, connections route through the SOCKS proxy and prefer a `.onion` ElectrumX server. No IP leaks.

**Name expiry checking** — names are validated against the current block height. Expired names (≥36,000 blocks since last update, ~250 days) are rejected. The `expiresIn` field is populated on active names for downstream use.

### How it works

Resolution uses the standard Electrum protocol (scripthash-based lookups):
1. Build canonical name index script matching ElectrumX-NMC indexing
2. Query `blockchain.scripthash.get_history` for the name's tx history
3. Parse `NAME_UPDATE` script from the latest transaction output
4. Extract Nostr pubkey from the name's JSON value
5. Query `blockchain.headers.subscribe` for current block height
6. Reject if `currentHeight - nameHeight ≥ 36,000` (expired)

Supports both `d/` (domain) and `id/` (identity) Namecoin namespaces, simple and extended NIP-05-like value formats with relay hints, LRU caching with 1h TTL, and self-signed TLS certificates.

### Tor & proxy support

The `ElectrumxClient` accepts an injected `SocketFactory` lambda so all TCP connections go through the user's configured proxy (Tor SOCKS5). A `ProxiedSocketFactory` handles the SOCKS routing. SSL is layered on top of the proxied socket.

Server selection is dynamic:

| Tor setting | Primary server | Fallback |
|---|---|---|
| Off | `electrumx.testls.space:50002` | `ulrichard.ch:50006`, `nmc2.lelux.fi:50006` |
| On | `.onion:50002` | `electrumx.testls.space:50002` (via Tor) |

Both the socket factory and server list are lambdas evaluated per-request — toggling Tor takes effect immediately.

### Name expiry

Namecoin names expire 36,000 blocks (~250 days) after their last `NAME_UPDATE` transaction. The client queries the current block height via `blockchain.headers.subscribe` and rejects expired names, returning `null` (treated as "not found" by callers). Unconfirmed transactions (height ≤ 0) are treated as active. Active names have their `expiresIn` field populated with the remaining block count.

### Files

**New (quartz/):**
- `ElectrumxClient` — TCP/TLS client with injected `SocketFactory`, expiry checking
- `NamecoinNameResolver` — Identifier parsing, value extraction, dynamic server selection
- `NamecoinLookupCache` — LRU cache with TTL
- `NamecoinNameResolverTest` — Unit tests

**New (amethyst/):**
- `NamecoinNameService` — App singleton, initialized with proxy-aware client
- `ProxiedSocketFactory` — `SocketFactory` routing through SOCKS5 proxy (Tor)

**Modified:**
- `Nip05Client` — Routes `.bit` identifiers to Namecoin resolver
- `AppModules` — Wires up resolver with Tor-aware socket factory and server list
- `RoleBasedHttpClientBuilder` — Added `socketFactoryForNip05()` for proxy-aware socket creation
- `SearchBarViewModel` — Namecoin search resolution via `namecoinResolvedUser` flow

**Docs:**
- `docs/namecoin-nip05-design.md` — Full architecture, protocol, Tor integration, and expiry details

### How to test

**Unit tests:**
```bash
./gradlew :quartz:jvmTest --tests "*NamecoinNameResolverTest*"
```

**Manual testing on emulator or device:**

Build and install:
```bash
./gradlew assemblePlayDebug
adb install -r amethyst/build/outputs/apk/play/debug/amethyst-play-universal-debug.apk
```

Search bar tests — open the search bar and enter each:

| Search query | Expected result |
|---|---|
| `m@testls.bit` | Resolves to Vitor Pamplona's profile (pubkey `6cdebcca...18667d`) |
| `testls.bit` | Resolves to same profile (root lookup) |
| `d/testls` | Resolves to same profile (direct namespace) |

**Tor tests** — enable Tor and "NIP-05 verifications via Tor":
1. Search for `m@testls.bit` — resolves via onion server
2. `tcpdump` should show only local Tor SOCKS connections, no direct ElectrumX traffic
3. Toggle Tor off — next search uses clearnet servers immediately

**Expiry test** — an expired Namecoin name should return no results in the search bar. Active names show the resolved profile normally.

### Default ElectrumX servers

**Clearnet:**
- `electrumx.testls.space:50002` (TLS, self-signed)
- `ulrichard.ch:50006` (TLS)
- `nmc2.lelux.fi:50006` (TLS)

**Tor:**
- `i665jpwsq46zlsdbnj4axgzd3s56uzey5uhotsnxzsknzbn36jaddsid.onion:50002` (TLS, self-signed) — onion service for `electrumx.testls.space`
- `electrumx.testls.space:50002` (fallback, routed through Tor)

See `docs/namecoin-nip05-design.md` for full architecture details.


mstrofnone:
## Suggestions from porting this to Primal Android

I ported this PR to [PrimalHQ/primal-android-app#934](https://github.com/PrimalHQ/primal-android-app/pull/934) and hit several real-world issues during testing. Here are improvements that would also benefit the Amethyst implementation:

---

### 1. 🐛 Race condition: stale intermediate lookups overwrite final result

**The problem:** `SearchBarViewModel` debounces `searchValueFlow` at 400ms, but if a user types `d/testls` character by character, the debounce fires for intermediate values like `d/test`, `d/testl`, etc. Each triggers an ElectrumX lookup that takes 1-3 seconds. The lookups complete out of order — a later query (`d/testls`) may resolve correctly, but then an earlier stale query (`d/testl`) finishes and overwrites the result with `null`.

The `namecoinResolvedUser` flow in `SearchBarViewModel` uses `.map {}` on debounced input, which does **not** cancel previous emissions. Each intermediate lookup runs to completion.

**The fix:** Use `mapLatest` instead of `map` for the `namecoinResolvedUser` flow. `mapLatest` cancels the previous transform when a new value arrives:

```kotlin
private val namecoinResolvedUser =
    searchValueFlow
        .debounce(400)
        .distinctUntilChanged()
        .filter { NamecoinNameResolver.isNamecoinIdentifier(it) }
        .mapLatest { term ->          // ← mapLatest, not map
            try {
                val result = NamecoinNameService.getInstance().resolve(term)
                if (result != null) LocalCache.getOrCreateUser(result.pubkey) else null
            } catch (_: Exception) {
                null
            }
        }
        .flowOn(Dispatchers.IO)
        .stateIn(viewModelScope, WhileSubscribed(5000), null)
```

---

### 2. 🐛 CancellationException swallowed in catch block

Related to the above: if you do switch to `mapLatest`, the cancelled coroutine will throw `CancellationException`. The current `catch (_: Exception)` block catches it and returns `null`, which then propagates downstream as a valid "not found" result and clears the UI.

**The fix:** Re-throw `CancellationException`:

```kotlin
.mapLatest { term ->
    try {
        val result = NamecoinNameService.getInstance().resolve(term)
        if (result != null) LocalCache.getOrCreateUser(result.pubkey) else null
    } catch (e: CancellationException) {
        throw e
    } catch (_: Exception) {
        null
    }
}
```

This is standard Kotlin coroutine practice but easy to miss.

---

### 3. 🐛 Root lookup fails when `names` object has no `_` key

**The problem:** In `extractFromDomainValue`, the extended form tries `names[parsed.localPart]` (which is `names["_"]` for root lookups) then falls back to `names["_"]` — the same thing. If the on-chain data only has named entries (e.g. `{"names": {"m": "6cdebcca..."}}`), root lookups return `null`.

Real-world example: A name with `{"nostr": {"names": {"m": "6cdebcca..."}}}` and no `_` entry. Searching `d/example` or `example.bit` fails silently.

**The fix:** For root lookups (`localPart == "_"`), fall back to the first available entry when `_` is not present:

```kotlin
val pubkeyElem = names[parsed.localPart]
    ?: names["_"]
    ?: if (parsed.localPart == "_") names.entries.firstOrNull()?.value else null
    ?: return null
```

Similarly update the `localPart` in the returned result to reflect which key was actually matched.

---

### 4. 💡 Global Mutex serializes all lookups

`ElectrumxClient` uses a single `Mutex` around `connectAndQuery`, which means `nameShowWithFallback` is fully serialized — even when querying **different** names on **different** servers. If the first server times out (10s connect + 15s read), every other lookup waits behind it.

During testing, this meant intermediate lookups (from the debounce race) blocked the final correct lookup for 20+ seconds.

**Suggestion:** Either remove the mutex entirely (each call creates its own socket, so there is no shared state to protect), or scope it per-server if the intent is to limit concurrent connections:

```kotlin
// Per-server mutex to limit connections without blocking unrelated lookups
private val serverMutexes = ConcurrentHashMap<String, Mutex>()

private fun mutexFor(server: ElectrumxServer) =
    serverMutexes.getOrPut("${server.host}:${server.port}") { Mutex() }
```

---

### 5. 💡 Add a loading indicator in the UI

ElectrumX lookups take 1-5 seconds depending on the server. During this time the user sees no feedback that anything is happening. 

**Suggestion:** Emit a loading state before the resolve starts. In the Primal port I added a `namecoinResolving: Boolean` flag to the UI state and show a spinner with "Resolving via Namecoin blockchain…" text while the lookup is in flight.

For Amethyst, you could do this with a sealed class wrapper or a separate `isResolvingNamecoin` StateFlow that the search UI observes.

---

### 6. 💡 `NamecoinNameService` singleton — consider structured concurrency

`NamecoinNameService` creates its own `CoroutineScope(Dispatchers.IO + SupervisorJob())` that is never cancelled. If the service is re-initialized or the app goes to background, orphaned coroutines from `resolveLive()` will keep running.

**Suggestion:** Either tie the scope to the `ServiceManager` lifecycle, or make `resolveLive` accept an external scope:

```kotlin
fun resolveLive(identifier: String, scope: CoroutineScope): StateFlow<NamecoinResolveState>
```

---

### 7. 💡 `customServers` is set but never read

`NamecoinNameService.setCustomServers()` stores the list but `resolver` was constructed with the default `serverListProvider`. The custom servers are never actually used.

**Suggestion:** Wire it through:

```kotlin
private val resolver = NamecoinNameResolver(
    electrumxClient = electrumxClient,
    serverListProvider = { customServers.ifEmpty { ElectrumxClient.DEFAULT_SERVERS } },
)
```

---

These are all things I hit during real testing on the Primal port. The race condition (#1) and CancellationException (#2) are the most impactful — they cause the resolved profile to appear briefly then vanish, or the wrong profile to stick. Happy to submit patches for any of these.

PR #1771, opened 2026-03-06
fix: address race conditions and improve Namecoin NIP-05 resolution

mstrofnone:
## Summary

Fixes 6 issues discovered while porting the Namecoin NIP-05 feature (#1734) to [Primal Android](https://github.com/PrimalHQ/primal-android-app/pull/934) and testing against live ElectrumX servers.

Based on the `namecoin-nip05` branch from #1734. See also: [detailed analysis in #1734 comment](https://github.com/vitorpamplona/amethyst/pull/1734#issuecomment-4008647487)

### Bug fixes

**1. Race condition: stale intermediate lookups overwrite final result**

`SearchBarViewModel.namecoinResolvedUser` used `.map {}` which does not cancel previous emissions. Typing `d/testls` character by character triggers lookups for `d/test`, `d/testl`, etc. These complete out of order — the correct `d/testls` result arrives first, then a stale `d/testl → null` overwrites it.

**Fix:** Changed `map` → `mapLatest`, which cancels the previous transform when a new value arrives.

**2. CancellationException swallowed in catch block**

With `mapLatest`, cancelled coroutines throw `CancellationException`. The generic `catch (_: Exception)` was catching it and emitting `null`, wiping the valid result from the final query.

**Fix:** Re-throw `CancellationException` before the generic catch.

**3. Root lookup fails when `names` object has no `_` key**

`extractFromDomainValue` tried `names[localPart]` then `names["_"]` as fallback — but for root lookups these are the same key. If the on-chain data only has named entries (e.g. `{"names": {"m": "pubkey..."}}`), root lookups silently returned `null`.

**Fix:** For root lookups, fall back to the first available entry when `_` is not present. Non-root lookups still fail correctly.

**4. Global Mutex serializes all ElectrumX lookups**

A single `Mutex` in `ElectrumxClient.nameShow()` meant that a lookup to `nmc2.bitcoins.sk` had to wait for a lookup to `electrumx.testls.space` to finish — even though they use completely separate sockets. A 25-second timeout on one server blocked everything.

**Fix:** Per-server mutexes via `ConcurrentHashMap<String, Mutex>`.

### Improvements

**5. `customServers` was set but never read**

`NamecoinNameService.setCustomServers()` stored the list but the `resolver` was constructed with the default `serverListProvider` and never referenced `customServers`.

**Fix:** Wire `customServers` through to the resolver via the `serverListProvider` lambda.

**6. `resolveLive` orphaned CoroutineScope**

`NamecoinNameService` creates `CoroutineScope(Dispatchers.IO + SupervisorJob())` that is never cancelled. `resolveLive()` launches into this scope, so coroutines outlive any caller lifecycle.

**Fix:** `resolveLive` now accepts an optional `scope` parameter so callers can tie resolution to their own lifecycle (e.g. `viewModelScope`). Falls back to the internal scope for backward compatibility.

## Tests

- 2 new unit tests: root fallback to first entry, non-root no-fallback
- All existing tests pass
- Tested on emulator against live Namecoin ElectrumX servers

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
@@ -163,9 +163,6 @@ class AppModules(
                 }
             },
         )
-    val namecoinNameService =
-        com.vitorpamplona.amethyst.service.namecoin.NamecoinNameService
-            .init(namecoinElectrumxClient)
     val nip05Client = Nip05Client(nip05Fetcher, namecoinResolver)
 
     // Application-wide block height request cache


mstrofnone:
Added a new commit that propagates distinct error types from ElectrumX through to the search UI. Discovered while porting Namecoin NIP-05 to notedeck (https://github.com/damus-io/notedeck/pull/1314).

**What changed:**

`NamecoinLookupException` sealed class with three variants:
- `NameNotFound` — name queried successfully, doesn't exist on blockchain
- `NameExpired` — name exists but expired (>36000 blocks)
- `ServersUnreachable` — all ElectrumX servers failed with connection errors

Previously all three cases returned `null` from `nameShowWithFallback`, so the search UI couldn't tell the user *why* resolution failed.

**Key design decisions:**
- `NameNotFound`/`NameExpired` are **definitive answers** from the blockchain → `nameShowWithFallback` short-circuits (no point trying other servers)
- Connection errors try the next server; `ServersUnreachable` only thrown when all servers fail
- `SearchBarViewModel` now exposes a typed `namecoinSearchState` flow (`Idle`/`Loading`/`Resolved`/`NotFound`/`Expired`/`ServersUnreachable`) — the Compose UI can observe this to show spinners, profiles, or error messages
- The existing `namecoinResolvedUser` is now derived from `namecoinSearchState` (no duplicate resolution)

amethyst/src/main/java/com/vitorpamplona/amethyst/service/namecoin/NamecoinNameService.kt
```diff
@@ -43,12 +44,33 @@ class NamecoinNameService(
...
+    private val resolver =
+        NamecoinNameResolver(
+            electrumxClient = electrumxClient,
+            serverListProvider = { customServers.ifEmpty { DEFAULT_ELECTRUMX_SERVERS } },
+        )
+    private val cache = NamecoinLookupCache()
+
+    companion object {
+        @Volatile
+        private var instance: NamecoinNameService? = null
```
vitorpamplona: Why is this back? It doesn't look like any part of the code is using it.

PR #1772, opened 2026-03-06
perf: optimize state management and clean up styling in ReactionsSettingsScreen

greenart7c3:
- Change `items` state initialization to use `toList()` instead of `toMutableList()` to ensure a fresh immutable snapshot from the flow.
- Reformat chained modifiers in `DraggableItem` for better readability.


davotoula:
Lint catches some... But I'm getting a lot of false positives about unused variables... Lint doesn't understand re-compose behaviour 😭

greenart7c3:
Yes, the false positives started in some gradle upgrade. Don't know if they will change it or we should start changing to `= remember` instead of `by remember`

vitorpamplona:
quick tip: Add the https://plugins.jetbrains.com/plugin/28767-compose-stability-analyzer

PR #1789, opened 2026-03-09
fix(quartz): fix NIP-44 key mutation and NIP-46 connect response handling

nrobi144:
## Summary

Two bugs in the Quartz NIP-46 remote signer library:

1. **NIP-44 HMAC key mutation** — `FixedKey.getEncoded()` returned a raw byte array reference. `javax.crypto.Mac.destroy()` zeroes the key material via `key.fill(0)`, corrupting the shared `saltPrefix` ("nip44-v2") in `Nip44v2` after first use. All subsequent NIP-44 decrypt operations fail with HMAC validation errors.

2. **NIP-46 connect response handling** — `connect()` used `PubKeyResponse::parse`, expecting a hex pubkey in the result. Per the NIP-46 spec and Amber maintainer feedback, connect never returns a pubkey — the response is `"ack"` (bunker://) or the secret string (nostrconnect://). Introduced `ConnectResponse` parser that checks `result`/`error` fields directly on the base `BunkerResponse` (matching Jackson deserialization behavior). `connect()` now returns `Unit` — callers call `getPublicKey()` separately as a domain concern.

## Changes

- **`FixedKey.getEncoded()`** — returns `key.copyOf()` instead of raw reference (prevents mutation)
- **`ConnectResponse`** — new parser that checks `result`/`error` fields directly on base `BunkerResponse`, not subclass pattern matching
- **`ConnectResult`** — sealed interface with `Ack` and `AlreadyConnected` (removed `PubKey` — connect never returns one)
- **`NostrSignerRemote.connect()`** — returns `Unit`, pure protocol handshake; callers call `getPublicKey()` separately

## Test plan

- [x] `Nip44v2JvmTest` — 5 tests: spec vectors, ECDH symmetry, encrypt/decrypt round-trip, cross-key-pair (NIP-46 pattern)
- [x] `ResponseParserTest` — 6 ConnectResponse tests (ack, secret string, already-connected, case-insensitive, error, no-result) + all other response parser tests (28 total)
- [x] All existing quartz + desktop tests pass

## Affected platforms

`FixedKey.kt` is in `jvmAndroid` source set — affects both Desktop JVM and Android.

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip46RemoteSigner/signer/ConnectResponse.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip46RemoteSigner/signer/ConnectResponse.kt
@@ -21,33 +21,23 @@
 package com.vitorpamplona.quartz.nip46RemoteSigner.signer
 
 import com.vitorpamplona.quartz.nip46RemoteSigner.BunkerResponse
-import com.vitorpamplona.quartz.nip46RemoteSigner.BunkerResponseAck
-import com.vitorpamplona.quartz.nip46RemoteSigner.BunkerResponseError
-import com.vitorpamplona.quartz.nip46RemoteSigner.BunkerResponsePublicKey
 
 class ConnectResponse {
     companion object {
-        fun parse(response: BunkerResponse): SignerResult.RequestAddressed<ConnectResult> =
-            when (response) {
-                is BunkerResponsePublicKey -> {
-                    SignerResult.RequestAddressed.Successful(ConnectResult.PubKey(response.pubkey))
-                }
-
-                is BunkerResponseAck -> {
-                    SignerResult.RequestAddressed.Successful(ConnectResult.Ack)
-                }
-
-                is BunkerResponseError -> {
-                    if (response.error?.contains("already connected", ignoreCase = true) == true) {
-                        SignerResult.RequestAddressed.Successful(ConnectResult.AlreadyConnected)
-                    } else {
-                        SignerResult.RequestAddressed.Rejected()
-                    }
-                }
-
-                else -> {
-                    SignerResult.RequestAddressed.ReceivedButCouldNotPerform()
+        fun parse(response: BunkerResponse): SignerResult.RequestAddressed<ConnectResult> {
+            if (response.error != null) {
+                return if (response.error.contains("already connected", ignoreCase = true)) {
+                    SignerResult.RequestAddressed.Successful(ConnectResult.AlreadyConnected)
+                } else {
+                    SignerResult.RequestAddressed.Rejected()
                 }
             }
+
+            if (response.result != null) {
+                return SignerResult.RequestAddressed.Successful(ConnectResult.Ack)
+            }
+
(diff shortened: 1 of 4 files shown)


nrobi144:
## NIP-46 `connect` Response Handling — Review

Investigated the 2nd fix (connect response handling) against the NIP-46 spec and Amber's source.

### NIP-46 spec

`connect` result: `"ack"` OR `<required-secret-value>` (echo-back for `nostrconnect://` anti-spoofing).

### Amber behavior

Amber returns `"ack"` for connect. Confirmed in `BunkerRequestUtils.kt`:
```kotlin
"connect" -> "ack"
```

### The bug (before this PR)

`connect()` used `PubKeyResponse::parse` — only accepted `BunkerResponsePublicKey` (64-char hex). When Amber sent `"ack"`, deserializer correctly mapped to `BunkerResponseAck`, but `PubKeyResponse::parse` hit the `else` branch → `ReceivedButCouldNotPerform` → login failure.

### The fix

`ConnectResponse::parse` now handles all known signer behaviors:

| Signer response | Deserializer type | ConnectResult | Action |
|---|---|---|---|
| 64-char hex pubkey | `BunkerResponsePublicKey` | `PubKey` | Use directly |
| `"ack"` | `BunkerResponseAck` | `Ack` | Fallback `getPublicKey()` |
| error "already connected" | `BunkerResponseError` | `AlreadyConnected` | Fallback `getPublicKey()` |
| other error | `BunkerResponseError` | `Rejected` | Throw |
| anything else | generic `BunkerResponse` | `ReceivedButCouldNotPerform` | Throw |

### One minor gap: secret echo-back

NIP-46 spec allows signers to return the **secret value** instead of `"ack"`. If a signer returned e.g. `"my-secret-123"`, the deserializer would produce a generic `BunkerResponse` (not hex, not ack, not JSON) → `ConnectResponse::parse` hits `else` → `ReceivedButCouldNotPerform` → login failure.

No known signer does this today (Amber sends `"ack"`), but it's technically spec-compliant. A defensive fix would treat generic `BunkerResponse` with non-null `result` as an ack-like success + validate the secret matches what was sent. Low priority.

### Verdict

Fix is **correct** for all known signers. The `ack` → `getPublicKey()` fallback is the right approach.

vitorpamplona:
@greenart7c3

greenart7c3:
The connect never returns a pubkey. 
It's ack or secret if using nostrconnect:// uri, secret is never returned if using bunker:// uri
While some bunkers like amber still use the main user pubkey in the requests the correct approach is always after a success connect call the getPublickey method

In Amber currently it's doing request p tag == getpublickey pubkey because at the time it's was not clear that i should have used a new key for each application

nrobi144:
@greenart7c3 Thanks for the correction — updated.

**Changes made:**
- Removed `ConnectResult.PubKey` — connect never returns a pubkey
- Rewrote `ConnectResponse.parse()` to check `result`/`error` fields directly on base `BunkerResponse` (not subclass pattern matching). Any non-null `result` (ack or secret) = success
- `connect()` now returns `Unit` — pure protocol handshake. Callers call `getPublicKey()` separately
- Tests updated to use base `BunkerResponse` to match Jackson deserialization behavior, added secret string test case

PR #1819, opened 2026-03-12
Add last seen timestamp tracking for user profiles

vitorpamplona:
## Summary
This PR adds functionality to track and display the last seen timestamp for user profiles in the Amethyst application. When viewing a user's profile, the app now fetches the most recent event from that user and displays when they were last seen.

## Key Changes
- **New filter assembler**: `UserProfileLastSeenFilterSubAssembler` - Manages subscriptions to fetch the latest event from a user to determine their last seen timestamp
- **New filter function**: `filterUserProfileLastSeen()` - Constructs relay-based filters to query a user's most recent event, prioritizing outbox relays, then all used relays, then relay hints
- **User model enhancement**: Added `lastSeenFlow()` and `updateLastSeen()` methods to the `User` class to track and expose the last seen timestamp as a reactive flow
- **UI component**: New `DisplayLastSeen()` composable function that displays the last seen time in a human-readable format (e.g., "2 hours ago")
- **Profile integration**: Integrated the last seen display into the user profile header via `DrawAdditionalInfo()`
- **Filter group update**: Added the new `UserProfileLastSeenFilterSubAssembler` to the `UserProfileFilterAssembler` group (increasing from 5 to 6 subscriptions per visible user profile)
- **Localization**: Added string resource for the "last seen" label

## Implementation Details
- The last seen timestamp is only updated if the new timestamp is greater than the previously stored value
- The filter queries for a single event (`limit = 1`) from the user's authors list
- EOSE (End of Stored Events) signals and live events both trigger timestamp updates
- The timestamp display uses the existing `timeAgo()` utility for consistent time formatting across the app

https://claude.ai/code/session_01EHAv38JZVdsw1mv2rA5DCB


davotoula:
Maybe an apply spotless before pushing would fix these automatically?

vitorpamplona:
I need to figure out how to get Claude to do that automatically.. internet says that I need to add to pre-commit git hooks, but that seems harsh.... idk...

davotoula:
Just putting it in Claude.md should be enough!

vitorpamplona:
Merged manually

PR #1833, opened 2026-03-13
feat: NIP-66 relay liveness filtering for outbox selection

alltheseas:
## Summary

Uses NIP-66 relay monitor data to filter dead relays from outbox relay selection, reducing wasted connections and improving feed load performance.

- Adds `RelayLivenessState` that fetches the alive relay set from `api.nostr.watch/v1/online` (HTTP API, refreshed hourly)
- Filters dead relays from outbox paths: `FollowListOutboxOrProxyRelays`, `FollowListReusedOutboxOrProxyRelays`, `FollowsPerOutboxRelay`
- Filters dead relays from search relay selection (`SearchRelayListState`) with a never-filter-to-zero safety guard
- Unions dead relays into `cannotConnectRelays` for user discovery/metadata loading (`AccountFollowsLoaderSubAssembler`)

### Safety

- **Fail-open**: `aliveRelaysFlow` starts as `emptySet()` — no filtering until data arrives
- **`.onion` preservation**: Tor relays always pass through (can't be validated by clearnet monitors)
- **Never filter to zero**: Search relay path preserves all relays if filtering would leave an empty set

### Benchmark Results (A/B comparison, 5 profiles × 3 algorithms, 1-year window)

| Metric | Without NIP-66 | With NIP-66 | Change |
|--------|---------------|-------------|--------|
| Relay success rate | 70–95% | 85–100% | **+5 to +30pp** |
| Dead relay pruning | — | 40–65% of candidates removed | — |
| Event recall | 14–33% | 15–32% | **~neutral** (+0.9pp mean) |
| Timeout tax | 0–15s | 0s | eliminated |

Relay pruning percentages by profile: fiatjaf 40%, hodlbod 48%, jb55 59%, ODELL 63%, Gato 68%, Telluride 65%.

## Test plan

- [x] `./gradlew test` passes (all 158 tasks, pre-push hook)
- [x] `./gradlew spotlessApply` passes
- [ ] Manual: verify `RelayLivenessState` logs alive set size after startup
- [ ] Manual: compare relay connection count before/after

### DIFF
--- a/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/ThreadDualAxisChartAssemblerTest.kt
+++ b/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/ThreadDualAxisChartAssemblerTest.kt
@@ -54,17 +54,16 @@ class ThreadDualAxisChartAssemblerTest {
         val keyPair = KeyPair()
         val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
 
-        val client =
-            NostrClient(
-                OkHttpWebSocket.Builder {
-                    OkHttpClient
-                        .Builder()
-                        .followRedirects(true)
-                        .followSslRedirects(true)
-                        .build()
-                },
-                scope,
-            )
+        val websocketBuilder =
+            OkHttpWebSocket.Builder {
+                OkHttpClient
+                    .Builder()
+                    .followRedirects(true)
+                    .followSslRedirects(true)
+                    .build()
+            }
+
+        val client = NostrClient(websocketBuilder, scope)
 
         val account =
             Account(
@@ -75,6 +74,7 @@ class ThreadDualAxisChartAssemblerTest {
                 otsResolverBuilder = EmptyOtsResolverBuilder,
                 cache = LocalCache,
                 client = client,
+                websocketBuilder = websocketBuilder,
                 scope = scope,
             )
 
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt
@@ -268,6 +268,7 @@ class AppModules(
             otsResolverBuilder = otsResolverBuilder,
             cache = cache,
             client = client,
+            websocketBuilder = websocketBuilder,
         )
 
     val sessionManager =
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -138,6 +138,7 @@ import com.vitorpamplona.quartz.nip01Core.relay.client.INostrClient
 import com.vitorpamplona.quartz.nip01Core.relay.client.accessories.downloadFirstEvent
 import com.vitorpamplona.quartz.nip01Core.relay.filters.Filter
 import com.vitorpamplona.quartz.nip01Core.relay.normalizer.NormalizedRelayUrl
+import com.vitorpamplona.quartz.nip01Core.relay.sockets.WebsocketBuilder
 import com.vitorpamplona.quartz.nip01Core.signers.EventTemplate
 import com.vitorpamplona.quartz.nip01Core.signers.NostrSigner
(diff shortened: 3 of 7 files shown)


alltheseas:
@vitorpamplona this is based on nostrability/outbox benchmark learnings.

The core benefit is that over one out of two relays does not return useful for a variety of reasons (e.g. dead relay, offline relay, inaccurate relay lists etc.). 

By filtering outbox paths with NIP-66 relay liveness, load time is shown to be reduced by ~~40%. Less connections are wasted to otherwise useless stated relays. 

More info on outbox nip-66 benchmarks here: https://github.com/nostrability/outbox

cc @dskvr

vitorpamplona:
> By filtering outbox paths with NIP-66 relay liveness, load time is shown to be reduced by ~~40%. Less connections are wasted to otherwise useless stated relays.

Is this stat measured on Amethyst or other clients? 

Why is this PR using the HTTP call instead of NIP-66 relay events? Wouldn't loading the same list of online relays via NIP-66 events be better? This makes the app dependent on `nostr.watch` as a provider and DNS-locked without any way for users to change who owns who they can coneect to... which is not good. 

> .onion preservation: Tor relays always pass through (can't be validated by clearnet monitors)

This is a problem since by default almost all relay connections to trough Tor on Amethyst. The user has to opt out of tor to change it. We need to know their status on the Tor network.

vitorpamplona:
Also, @dskvr, since we are talking about `nostr.watch`, can you add CORS * to `https://api.nostr.watch/v1/online` so that webapps can access it?

alltheseas:
> HTTP

This is a blunder. Latest commits now use native NIP-66 events. 

> Is this stat measured on Amethyst or other clients?

I extracted various outbox algos, including Amethyst, into a Typescript benchmarking script. You can recreate results by visiting - https://github.com/nostrability/outbox/blob/main/Benchmark-recreation.md.

> Tor

I did not test on Tor, good point. 

Pushed commits now if Tor mode is on for Amethyst outbox/nip-66 connection to not ask clearnet nip-66 monitors. I am not sure if any monitors serve Tor currently.

In clearnet mode, Amethyst asks nip-66 monitors. 

If no useful info comes back, lack of nip-66 response does not block connections.

vitorpamplona:
Cool, so if you run this version, does Amethyst actually load 40% faster? It's hard to believe that just removing offline relays can be that effective since each one is just a different thread in the OS and doesn't block anything else. Removing the offline relays would just delete threads that are already paused, waiting for the server to reply. 

Also, I just noticed that this also blocks any relay that the monitor doesn't know it exists, and all of the personalized addresses, like wss://relay.com/npub1... Or was://npub1xxx.relay.com

alltheseas:
> It's hard to believe that just removing offline relays can be that effective since each one is just a different thread in the OS and doesn't block anything else

The mechanism isn't thread blocking. 

In benchmarking I found that 1) there are imperfect, changing relay lists, and 2) more than half of queried relays don't return useful info. 

By using nip-66 you are eliminating otherwise wasted outbox connections, and getting to querying live relays at a higher probability. 

The speed up comes from not waiting for relays that can't return useful info.

alltheseas:
> I just noticed that this also blocks any relay that the monitor doesn't know it exists, and all of the personalized addresses, like wss://relay.com/npub1... Or was://npub1xxx.relay.com

Will check this later this weekend

vitorpamplona:
> > It's hard to believe that just removing offline relays can be that effective since each one is just a different thread in the OS and doesn't block anything else
> 
> The mechanism isn't thread blocking. 
> 
> In benchmarking I found that 1) there are imperfect, changing relay lists, and 2) more than half of queried relays don't return useful info. 
> 
> By using nip-66 you are eliminating otherwise wasted outbox connections, and getting to querying live relays at a higher probability. 
> 
> The speed up comes from not waiting for relays that can't return useful info. 

That's why it's weird to understand.. why would clients be waiting for any relay..ever? The way Amethyst works is that each relay has two threads, one to read and one to write. If the server isnt there that thread is just paused until it comes back online. So at least for us, the only cost we are saving is the cost of creating threads for the unavailable relays. Which shoudn't be more than a few microseconds per relay.

PR #1848, opened 2026-03-15
Fix ios build issues

davotoula:
Was a full build ever working for you all?

After resolving the below issues I am able to do a full clean and full build (`./gradlew build`). Took around 17 minutes.

I have xcode, ios simulator installed locally. Also libsodium via brew.

Fixes:
- config: Use different binaries for ios device and ios simulator
- test: Correct toHexKey function that is used (a java function was used before)
- Workaround for compose.stability.analyzer plugin (v0.7.0) not declaring task dependencies properly on Gradle 9.x
- Fix ArrayIndexOutOfBoundsException in GZipTest
- Fix :quartz:testAndroidHostTest
	- install libsodium using brew and configured tests to use local dependencies
	- fix invalid nsec string
	- fix Jackson null return for unknown types
	- Excluded — Nip47WalletConnectTest from testAndroidHostTest because it requires android.net.Uri (should we consider Robolectric?)

note: a (slow) mac verification was run for binaries before ios tests but apparently it's a once only, should not happen on future builds. Only after xcode or ios simulater updates.
note: NostrClientRepeatSubTest.kt is intermittent as it uses live relay (wss://relay.damus.io) and expects exactly 100 events
note: there are around 118 failing quartz tests due to "Not implemented yet"


vitorpamplona:
Building works on a MacOS. But I have a feeling that our build files need to be fixed for other OSs. :(

davotoula:
Have you tried a clean build from recent main?

Do you have libsodium installed?

quartz/src/commonTest/kotlin/com/vitorpamplona/quartz/nip47WalletConnect/NwcInfoEventTest.kt
```diff
@@ -28,7 +28,7 @@ import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 class NwcInfoEventTest {
-    private val signer = DeterministicSigner("nsec1w4uucmeyyng0kegm7486r23sv4majkmvqsj6eypprq0xttxss55s5mgg9t".nsecToKeyPair())
+    private val signer = DeterministicSigner("nsec10g0wheggqn9dawlc0yuv6adnat6n09anr7eyykevw2dm8xa5fffs0wsdsr".nsecToKeyPair())
```
vitorpamplona, line 31: changing keys? Does this make sense? This test was taken from alby

davotoula: The Claude resolution comment was that key was invalid so it used a valid one from one of our tests. I didn't confirm why/if the original was invalid

vitorpamplona:
I pulled out the text fixes so that we can focus on the build stuff here

vitorpamplona:
Looks like the actions are running and compiling for macos, windows and linux.. did you get it to work in your machine?

davotoula:
I doesn't work on my local mac if i'm not on this branch.. but i haven't tried your new Action locally.

I'll keep investigating

davotoula:
Actually, it's the full `./gradlew build` that doesn't work but maybe that should never be used... Or it is still work in progress...

vitorpamplona:
I think I imported everything to main. I also fixed some test cases with libsodium that were in the wrong place

KotlinGeekDev:
I imported the recent changes from main some hours ago, and it worked for me. I'll test the latest changes and see...

davotoula:
full build now work on main, thanks for fixing it.

For fun I will try on fresh linux and windows VM.

closing this PR

PR #1863, opened 2026-03-16
Reduce compilation warnings

davotoula:
Just low risk mechanical changes. Compilation warnings from 250 -> 199

- Unnecessary ?. / ?: / !! on non-null
- Missing @OptIn annotations
- Parameter name mismatches
- Icons.Filled.*  / Icons.AutoMirrored.*
- No cast needed / remove inline
- LocalLifecycleOwner import fix
- NIP-51 name() → title()
- Chess gameId → startEventId
- Nullable DecimalFormat safe calls
- DelicateCoroutinesApi opt-in

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/chess/LiveChessGame.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/chess/LiveChessGame.kt
@@ -237,7 +237,7 @@ fun LiveChessGameScreen(
             ) {
                 // Game info - use currentPosition.activeColor for turn display
                 GameInfoHeader(
-                    gameId = gameState.gameId,
+                    gameId = gameState.startEventId,
                     opponentName = opponentName,
                     playerColor = gameState.playerColor,
                     currentTurn = currentPosition.activeColor,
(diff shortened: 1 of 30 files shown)


commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/chess/LiveChessGame.kt
```diff
@@ -237,7 +237,7 @@ fun LiveChessGameScreen(
             ) {
                 // Game info - use currentPosition.activeColor for turn display
                 GameInfoHeader(
-                    gameId = gameState.gameId,
+                    gameId = gameState.startEventId,
```
vitorpamplona, line 240: is this correct? Looks more than just a refactoring... Just checking

davotoula: Might have been a deprecation, let me double check before merging.

@nrobi144 ?

davotoula:
```
/** Legacy alias for startEventId */
    @Deprecated("Use startEventId instead", ReplaceWith("startEventId"))
    val gameId: String get() = startEventId
```

```
/** Start event ID - the game identifier in Jester protocol */
    val startEventId: String,
```

The code comments suggest is strongly.

davotoula:
This keeps happening:

`* What went wrong:
Execution failed for task ':amethyst:packageFdroidBenchmark'.
> A failure occurred while executing com.android.build.gradle.tasks.PackageAndroidArtifact$IncrementalSplitterRunnable
`

I think it was suggested that it's too heavy for CI build, should be divided or something

PR #1877, opened 2026-03-19
Deprecate NIP-04 and always use NIP-17 for DM sending

vitorpamplona:
## Summary
This PR deprecates NIP-04 (legacy unencrypted DMs) and makes NIP-17 (encrypted sealed messages) the only protocol used for sending direct messages. The changes also improve relay availability checking to prevent sending messages to recipients without DM relay lists.

## Key Changes

- **NIP-17 is now mandatory**: Removed all NIP-04 sending code paths and made `nip17` a read-only property that always returns `true`
- **Removed NIP-17 toggle UI**: Deleted the `toggleNIP04And24()` function and the toggle button that allowed users to switch between NIP-04 and NIP-17
- **Improved relay availability checking**: 
  - Renamed `requiresNIP17` to `recipientsMissingDmRelays` for clarity
  - Changed logic from checking group chat size to actually verifying if recipients have DM relay lists
  - Messages can now only be sent if all recipients have DM inbox relays (kind 10050) or NIP-65 inbox relays
- **Updated UI feedback**:
  - Added `RecipientMissingRelaysWarning()` component that displays when recipients lack DM relay lists
  - Changed the NIP-17 indicator from a toggleable button to a read-only icon showing NIP-17 is always active
  - Updated `canPost()` validation to block sending when `recipientsMissingDmRelays` is true
- **Simplified file upload logic**: Removed conditional NIP-04/NIP-17 branching in `ChatFileUploader` and `ChatFileSender`, now only using NIP-17 paths
- **Cleaned up imports**: Removed unused `AdvertisedRelayListEvent` and `PrivateDmEvent` imports where no longer needed

## Implementation Details

- The `updateRecipientRelayStatus()` method now checks each recipient's `dmInboxRelays()` to determine if message delivery is possible
- Desktop and mobile UIs now consistently show NIP-17 as the only available encryption method
- The `ChatNewMessageState` class was updated to remove NIP-04 sending logic and the `nip17` state flow
- Error messaging guides users to set up DM relay lists when recipients are missing them

https://claude.ai/code/session_01T7QhUW9cZogk4DxDXbbbJJ

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/privateDM/ChatroomView.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/privateDM/ChatroomView.kt
@@ -26,14 +26,16 @@ import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.viewmodel.compose.viewModel
+import com.vitorpamplona.amethyst.service.relayClient.reqCommand.event.EventFinderFilterAssemblerSubscription
 import com.vitorpamplona.amethyst.ui.feeds.WatchLifecycleAndUpdateModel
 import com.vitorpamplona.amethyst.ui.navigation.navs.INav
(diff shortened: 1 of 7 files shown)


TotallyAvailable:
@vitorpamplona ***Please*** remember to set the GITHUB_TOKEN permissions for ***all*** workflows to the ***absolute minimum*** required (including 'Secret Source: Actions') to avoid ***any*** further use of default **write-all** tokens.

Making it another 2-in-1:
@greenart7c3 This appears to affect most (if not all) of your repos and workflows as well.

(Edit afterwards due to "circumstances"... (re-)consider the necessity of 'persist-credentials' and the consequences of the current ***default***... There *might* even be *more* GitHub design choices worth pointing out. How all of this survived this long is *wild*.)

Apologies as this does have nothing to do with this PR, rather a "less obvious" reach out to get this issue hopefully taken care of after the ***many*** years it has existed. Not the fine art of course.

PR #1887, opened 2026-03-20
UI polish: feed filter

davotoula:
- modernise SpinnerSelectionDialog with Material 3 styling
- add grouped feed filter dialog with Material 3 styling 

Screenshot_20260320-140704


vitorpamplona:
NICE!!

PR #1903, opened 2026-03-23
fix(chess): normalize SAN for cross-client Jester interop

nrobi144:
## Summary
- Fixes chess games against jester.nyo.dev and other clients that use `0-0`/`0-0-0` (zeros) for castling notation
- Fixes JesterContent JSON serialization — `version`, `fen`, and `history` fields were omitted, breaking jester `isStartGameEvent` check
- Adds `offchain.pub` relay for better jester interop (2 shared relays instead of 1)
- Fixes `discoverUserGames()` silently dropping games classified as spectator
- 25 interop tests

## Problem
1. **Castling breaks games:** chesslib rejects `0-0`/`0-0-0` (zeros) causing ChessStateReconstructor to mark games as desynced. Games appear permanently stuck, even after restart.
2. **Challenges invisible to jester:** JesterContent fields `version`, `fen`, `history` were omitted from JSON. Jester checks `arrayEquals(json.history, [])` which fails when field is absent.
3. **Single shared relay:** Only relay.damus.io was shared with jester.

## Changes
- `ChessEngine.jvmAndroid.kt` — `normalizeSan()` converts `0-0` to `O-O`, strips annotations
- `JesterContent.kt` — `@EncodeDefault` on version, fen, history
- `ChessConfig.kt` — Added offchain.pub relay
- `ChessLobbyLogic.kt` — discoverUserGames adds spectator games instead of dropping
- `ChessInteropBugTest.kt` — 25 new tests

## Test plan
- [x] 25 new interop tests all pass
- [x] Zero regressions on existing chess tests
- [x] spotlessCheck passes

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip64Chess/jester/JesterContent.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip64Chess/jester/JesterContent.kt
@@ -20,18 +20,26 @@
  */
 package com.vitorpamplona.quartz.nip64Chess.jester
 
+import kotlinx.serialization.EncodeDefault
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.Serializable
 
 /**
- * JSON content structure for Jester events
+ * JSON content structure for Jester events.
+ *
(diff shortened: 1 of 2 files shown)


davotoula:
Ready to be merged?

PR #1905, opened 2026-03-23
feat(desktop): cache-centric architecture for desktop feeds

nrobi144:
## Summary

- Replaces direct relay event rendering with a cache-centric architecture for desktop
- Events flow through `DesktopLocalCache` → `DesktopFeedViewModel` → UI, enabling instant back-navigation and proper deduplication
- Fixes critical bug where FeedScreen showed 0 relays/notes/follows on launch (relay subscription deadlock)
- **Replaced `BoundedLargeCache` with Android's `LargeSoftCache` (WeakReference-based)** — notes are now GC-driven, not arbitrarily evicted
- **Extracted 3 State classes to commons** (BookmarkListState, Kind3FollowListState, Nip65RelayListState) — Desktop pins important notes via strong refs, matching Android's pattern
- Adds 37 integration tests covering the full cache → filter → ViewModel pipeline

### PoW

https://github.com/user-attachments/assets/453effa5-b2c5-43b1-b559-9b01a87df308

### Cache Layer — WeakReference Model (NEW)
- `LargeSoftCache` moved from `amethyst/` to `commons/jvmAndroid/` — shared between Android and Desktop
- Desktop uses `LargeSoftCache<K, WeakReference<V>>` instead of `BoundedLargeCache` (deleted)
- No hard size limits — GC respects the Note reference graph (replies, reactions, zaps keep clusters alive)
- `ICacheProvider` return types fixed from `Any?` to proper `User?`/`Note?`
- Single-pass `cleanUp()` (zero allocation, ~40% faster than original double-pass)

### Note Pinning via State Classes (NEW)
- `BookmarkListState` extracted to `commons/commonMain/` — pins bookmark AddressableNote
- `Kind3FollowListState` extracted with `Kind3FollowListRepository` interface
- `Nip65RelayListState` extracted with `Nip65RelayListRepository` interface
- Per-feature repository interfaces match existing `EphemeralChatRepository` pattern
- `DesktopIAccount` creates all 3 State objects — pinned notes survive GC

### Memory Cleanup (NEW)
- Periodic heap-aware cleanup: checks every 30s, runs at >75% heap or every 5 minutes
- `cleanMemory()` sweeps stale WeakRef entries, `cleanObservers()` clears unused NoteFlowSets
- Try-catch per operation, 2-minute startup grace period

### Feed Architecture
- `DesktopFeedViewModel`: cache-backed ViewModel with init hydration + event stream
- Feed filters: Global, Following, Profile, Thread, Notification, Search, Bookmarks, Reads

### Bug Fixes
- Missing relay subscriptions on FeedScreen and UserProfileScreen
- Relay deadlock: switched `connectedRelays` to `relayStatuses.keys`
- Profile metadata/counts cached and seeded on compose

### Tests (37)
- Cache pipeline + coordinator integration tests

## Test plan
- [x] All 35+ integration tests pass
- [x] `./gradlew spotlessCheck` passes
- [x] Android compiles with no behavior change
- [ ] Manual: desktop app feed loading + navigation persistence

### DIFF
--- a/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/ui/NotificationsScreen.kt
+++ b/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/ui/NotificationsScreen.kt
@@ -58,6 +58,7 @@ import com.vitorpamplona.amethyst.commons.ui.components.LoadingState
 import com.vitorpamplona.amethyst.commons.ui.feed.FeedHeader
 import com.vitorpamplona.amethyst.commons.util.toTimeAgo
 import com.vitorpamplona.amethyst.desktop.account.AccountState
+import com.vitorpamplona.amethyst.desktop.cache.DesktopLocalCache
 import com.vitorpamplona.amethyst.desktop.network.DesktopRelayConnectionManager
 import com.vitorpamplona.amethyst.desktop.subscriptions.DesktopRelaySubscriptionsCoordinator
 import com.vitorpamplona.amethyst.desktop.subscriptions.createNotificationsSubscription
@@ -109,6 +110,7 @@ sealed class NotificationItem(
 @Composable
 fun NotificationsScreen(
     relayManager: DesktopRelayConnectionManager,
+    localCache: DesktopLocalCache,
     account: AccountState.LoggedIn,
     subscriptionsCoordinator: DesktopRelaySubscriptionsCoordinator? = null,
 ) {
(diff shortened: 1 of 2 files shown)


vitorpamplona:
FYI this maximum size approach might not work well, as things will just disappear. 

On Mobile, we use WeakReferences over a Note "Graph" that are removed when nobody else in the app is using them. The GC deals with circular references. But that means you have to keep a reference to events you want to keep in memory elsewhere, like how we do it on the Account class that keeps most of the Note references for the user's lists and settings. Or like, when we load a note with likes, replies, etc, the main note becomes a lifecycle on itself and all replies are deleted only when the main note is removed from memory.

nrobi144:
> FYI this maximum size approach might not work well, as things will just disappear.
> 
> On Mobile, we use WeakReferences over a Note "Graph" that are removed when nobody else in the app is using them. The GC deals with circular references. But that means you have to keep a reference to events you want to keep in memory elsewhere, like how we do it on the Account class that keeps most of the Note references for the user's lists and settings. Or like, when we load a note with likes, replies, etc, the main note becomes a lifecycle on itself and all replies are deleted only when the main note is removed from memory.

@vitorpamplona you are right - that's a good point. I did think initially there's a low hanging fruit with caching here to make the UX nicer. But I want to get this right so it's not causing events users care about to be dropped or not loaded

nrobi144:
Thanks for the feedback! You were right — the `BoundedLargeCache` approach was wrong.

**What changed:**

1. **Moved `LargeSoftCache` to `commons/jvmAndroid/`** — Desktop now uses the same WeakReference-based cache as Android. `BoundedLargeCache` is deleted. No hard size limits, GC respects the reference graph.

2. **Extracted State classes to commons** — `BookmarkListState`, `Kind3FollowListState`, and `Nip65RelayListState` now live in `commons/commonMain/` with per-feature repository interfaces (matching the `EphemeralChatRepository`/`PublicChatListRepository` pattern). `DesktopIAccount` creates these State objects so pinned AddressableNotes survive GC — same "long-term reference" pattern as Android.

3. **`ICacheProvider` returns typed `User?`/`Note?`** instead of `Any?` — since these types already live in commons.

4. **Periodic heap-aware cleanup** — Desktop has no `onTrimMemory`, so the coordinator checks heap usage every 30s and runs `cleanUp()` + `cleanObservers()` at >75% heap or every 5 minutes.

Android is untouched behavior-wise — `LargeSoftCache` is just imported from commons instead of the local amethyst package now (one import change in `LocalCache.kt`).

Remaining State classes (MuteList, UserMetadata, etc.) will be extracted as Desktop builds features that need them.

vitorpamplona:
Is this still draft?

nrobi144:
> Is this still draft?

wanted to test it out a bit and merge in the long-form support so that gets caching as well. Finishing up soon

nrobi144:
@vitorpamplona this is now ready

PR #1909, opened 2026-03-23
Kmp completeness, part 3 - adios, Swift bridge.

KotlinGeekDev:
This replaces everything the swift bridge was used for, and removes it.


vitorpamplona:
Can you update this one? URLs.ios.kt‎ got deleted since we now have the URLDetector in KMP. There is no need for the iOS version anymore. 

Do you want to use your lib for java as well? Then we can drop the dependency on that RFC 3986 java implementation

KotlinGeekDev:
> Can you update this one? URLs.ios.kt‎ got deleted since we now have the URLDetector in KMP. There is no need for the iOS version anymore.
> 
Got it.

> Do you want to use your lib for java as well? Then we can drop the dependency on that RFC 3986 java implementation

No problem. I'll just move it to the commonMain dependencies block.

PR #1915, opened 2026-03-24
Render Namecoin identifiers as clickable Nostr profile links

mstrofnone:
Make Namecoin identifiers (`d/name`, `id/name`, `name@domain.bit`, `@domain.bit`, `domain.bit`) clickable in note content, resolving to Nostr profiles via ElectrumX blockchain lookups.

## Behavior

| Format | Example | Rendered as |
|--------|---------|-------------|
| Bare .bit domain | `testls.bit` | Clickable → resolves to Nostr profile |
| NIP-05 style | `m@testls.bit` | Clickable → resolves to Nostr profile |
| d/ namespace | `d/testls` | Clickable → resolves to Nostr profile |
| id/ namespace | `id/alice` | Clickable → resolves to Nostr profile |
| URL with scheme | `https://example.bit` | Opens browser (normal URL behavior) |
| URL with scheme | `http://www.example.bit` | Opens browser (normal URL behavior) |

## Why `https://example.bit` should open the browser

.bit domains are real DNS names on the Namecoin blockchain. They can host actual websites, resolvable via:

- **[ncdns](https://www.namecoin.org/docs/ncdns/)** — Namecoin-to-DNS bridge that works with most applications. The ncdns Windows installer configures the Windows certificate verifier, so Chromium and Edge work with .bit HTTPS out of the box.
- **Tor Browser** — ncdns installer also configures .bit resolution in Tor Browser.
- **Namecoin TLS** — .bit domains support [full TLS certificate chains](https://www.namecoin.org/docs/name-owners/tls/) stored on-chain, providing censorship-resistant HTTPS.

When a user explicitly writes `https://example.bit`, they intend to visit a website. The browser handles resolution (or shows an error if no Namecoin DNS bridge is configured). This is consistent with how other alternative TLDs work.

Only **bare identifiers without a URL scheme** are treated as Nostr identity lookups.

### Rich text parser (`commons`)
- New `NamecoinSegment` type in `RichTextParserSegments.kt`
- Namecoin identifier detection runs BEFORE schemeless URL classification in `RichTextParser.kt`
- Words starting with `https://` or `http://` are excluded from Namecoin detection — these are browser URLs
- Uses `NamecoinNameResolver.isNamecoinIdentifier()` from quartz with trailing punctuation stripping

### Note content rendering (`amethyst`)
- New `NamecoinIdentifierLink.kt` composable:
  - Resolves identifiers via `Nip05Client.namecoinResolver` (ElectrumX blockchain)
  - On success → navigates to `Route.Profile(pubkey)` (in-app Nostr profile)
  - While resolving / on failure → Namecoin-blue (#4A90D9) styled text
- Wired into both `RenderWordWithPreview` and `RenderWordWithoutPreview`

## Testing
- ✅ Tested on Android emulator (API 35)
- ✅ All unit tests pass
- ✅ spotless clean

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/NamecoinIdentifierLink.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/NamecoinIdentifierLink.kt
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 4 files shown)


mstrofnone:
Just had the thought that the value at the time of the note's creation needs to be resolved possibly or if the name has not expired since the note maybe it's fine.

PR #1925, opened 2026-03-24
Add cleanup of temporary files in upload pipeline

vitorpamplona:
## Summary
This PR adds proper cleanup of temporary files created during the upload pipeline, preventing accumulation of intermediate files from compression, stripping, and encryption operations.

## Key Changes
- Added `deleteTempUri()` method to safely delete temporary files created during upload processing
  - Only deletes files that differ from the original (intermediate temp files, not user content)
  - Includes error handling and logging for failed deletions
- Wrapped both `upload()` and `uploadEncrypted()` methods with try-finally blocks to ensure cleanup
  - In `upload()`: cleans up stripped and compressed temp files
  - In `uploadEncrypted()`: cleans up encrypted, stripped, and compressed temp files in proper order
- Added necessary imports: `Log` from quartz utils and `File` from java.io

## Implementation Details
- Cleanup is performed in reverse order of file creation to properly handle chains of temporary files
- The method safely handles cases where URIs are identical (no deletion needed)
- Exceptions during cleanup are caught and logged without interrupting the upload result
- Uses try-finally to guarantee cleanup even if upload operations fail

https://claude.ai/code/session_01YS3dbZ1k84N2EHS3AruYRV


vitorpamplona:
@davotoula can you check this out to see if it makes sense? The goal is to delete the temporary files, otherwise they will consume the disk space forever since nothing deletes them later

davotoula:
I'll have a look. 

We assumed that leaving them in /cache they will be cleared eventually. 

One exception, when sharing video to other apps I have a minute delay on it to allow for sharing to finish

vitorpamplona:
Yeah, I am not really sure that happens on all phones. We certainly have seen the app misbehaving because the cache was over 3GB and couldn't not store any more videos or images from the feed. I am not sure if this is the issue, but it could be.

davotoula:
Let's try it! I will have a look. 

For me language files always show as 1Gb for the app even after cleaning cache

davotoula:
Manual test plan:
  - [x] Upload JPEG with compression — verify temp copy + stripped file deleted
  - [x] Upload JPEG uncompressed — verify stripped file deleted after upload
  - [x] Upload video with compression — verify compressed + stripped video deleted
  - [x] Upload small/already-compressed video — verify abandoned compressed file deleted
  - [x] Record and send voice message — verify stripped audio deleted
  - [x] Record and send voice message when upload failed — verify stripped audio deleted
  - [x] Upload MP3 with ID3 tags — verify input + stripped MP3 deleted
  - [x] Send image in DM (encrypted upload) — verify compressed, stripped, and encrypted files all deleted
  - [ ] Cancel upload mid-progress — verify no temp files remain
  - [x] Share video to another app — verify receiving app gets video, temp cleaned after ~1 min
  - [x] Upload GIF/SVG (compression skipped) — verify no orphaned temp files

davotoula:
done, tested, ready for merge/review @vitorpamplona

PR #1934, opened 2026-03-24
fix: reduce NoteCompose recomposition and scroll jitter

vitorpamplona:
## Summary

- **Use `drawBehind` for backgroundColor in `ClickableNote`** — the biggest win. Previously `backgroundColor.value` was read during composition, causing the entire note subtree to recompose when the "new item" highlight fades after 5 seconds. Now the color is only read during the draw phase.
- **Only launch fade-out `LaunchedEffect` for actually-new notes** — previously every note spawned a coroutine with `delay(5000)` even when `routeForLastRead` was null or the note wasn't new. During fast scrolling this created many unnecessary coroutines.
- **Cache `baseNote.event` in local vals** — `InnerNoteWithReactions`, `NoteBody`, and `FirstUserInfoRow` each accessed `baseNote.event` multiple times via property reads. Caching in a local val avoids repeated reads and redundant `is` checks.
- **Hoist `Modifier.fillMaxWidth()` in `FeedLoaded`** — avoids creating a new modifier instance per LazyColumn item on each recomposition.

## Test plan

- [ ] Scroll through home feed — should feel smoother with less jitter when new items appear
- [ ] Verify "new note" highlight still appears and fades after ~5 seconds
- [ ] Verify note click/long-press actions still work correctly
- [ ] Verify boosted/quoted notes render correctly
- [ ] Check that reactions row still displays properly

https://claude.ai/code/session_01XUiYFd7aiZ4Gd1pyCLXpdE

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/feeds/FeedLoaded.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/feeds/FeedLoaded.kt
@@ -29,6 +29,7 @@ import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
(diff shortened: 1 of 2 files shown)


vitorpamplona:
lol, what a joke claude.

davotoula:
😂

PR #1937, opened 2026-03-25
fix: pin ElectrumX server certs for Samsung One UI 7 (Android 16) compatibility

mstrofnone:
## Problem

Namecoin `.bit` name resolution fails silently on **Samsung devices running One UI 7 (Android 16)**. When it fails, users have **no way to diagnose the issue**.

## Root Cause

Samsung One UI 7 ships a hardened Conscrypt/TLS stack (Knox) that **silently rejects TLS connections using a no-op `X509TrustManager`**. The `ElectrumXClient` used this "trust-all" pattern to connect to ElectrumX servers with self-signed certificates.

### 1. Pin ElectrumX server certificates
Replace `trustAllSslFactory()` with `pinnedSslFactory()` using actual pinned certs + system CAs.

### 2. OEM compatibility hardening
- Cached SSLSocketFactory (avoid rebuild per connection)
- Explicit TLSv1.2 (`SSLContext.getInstance("TLSv1.2")` — Xiaomi/OnePlus Conscrypt forks may default to TLS 1.0)
- Enforce TLSv1.2+ on SSLSocket enabled protocols
- KeyStore type fallback to PKCS12 (Xiaomi MIUI)

### 3. Test Connection diagnostics UI
- Per-server test button with streaming results (✅/❌ + response time + error detail)
- Human-readable errors: "TLS handshake failed", "Connection refused", "DNS failed", "Timed out", etc.
- Diagnostic card: last test timestamp, device info, TLS version negotiated
- `ElectrumXClient.testServer()` captures cert PEM + SHA-256 fingerprint

### 4. TOFU cert pinning for custom servers
- Test Connection captures server certificates (Trust On First Use)
- **User must explicitly confirm** each new cert via dialog showing SHA-256 fingerprint
- Pinned certs persisted to DataStore, restored on app startup
- `ElectrumXClient.addPinnedCert()` / `setDynamicCerts()` for runtime management
- UI shows cert fingerprint in test results

### 5. Tor .onion server handling
- `.onion` addresses use pinned factory first (cert is already pinned), fall back to trust-all
- Verified via Tor SOCKS proxy: the .onion shares the same cert as `electrumx.testls.space`

### 6. GrapheneOS / hardened ROM compatibility
Reviewed against [GrapheneOS](https://github.com/GrapheneOS) source code (hardened Conscrypt, strict TLS enforcement on Pixel devices):
- **`.onion` trust-all hardening**: GrapheneOS may reject no-op `X509TrustManager` at the Conscrypt/BoringSSL layer, even for proxied sockets. `.onion` connections now use `cachedPinnedSslFactory()` as the primary path (the server's cert is already in `PINNED_ELECTRUMX_CERTS`). `onionSslFactory()` (trust-all) is kept only as a fallback on `SSLHandshakeException`.
- **Explicit trust decisions**: Auto-pinning replaced with confirmation dialog showing full SHA-256 fingerprint — aligns with GrapheneOS's security philosophy of explicit user consent for trust decisions.
- **`trustAllCerts` → `usePinnedTrustStore` rename**: The field no longer means "trust all certificates" — it means "use the pinned trust store." Renamed across all 4 files to prevent future contributors from misinterpreting the flag and reintroducing trust-all.
- **`trustAllCerts` → `usePinnedTrustStore` rename**: The field no longer means "trust all certificates" — it means "use the pinned trust store." Renamed across all 4 files to prevent future contributors from misinterpreting the flag and reintroducing trust-all.
- **Hostname verification note**: Raw `SSLSocket.createSocket()` uses the host parameter for SNI only, not hostname verification. This is a pre-existing gap (not introduced here) — a follow-up should add `endpointIdentificationAlgorithm="HTTPS"` or fingerprint-based verification for pinned certs.

### ✅ Directly tested (emulator instrumented tests)
| Device | OS | API | Result |
|--------|-----|-----|--------|
| Google Pixel 6 (AOSP emulator) | Android 16 | 36 | ✅ 8/8 tests pass |

### ✅ Specific tests passed on emulator
| Test | Result |
|------|--------|
| All 3 default servers (hostname + IP) | ✅ |
| IP-address server (46.229.238.187 — SNI-free TLS) | ✅ |
| Hostname server (electrumx.testls.space — SNI TLS) | ✅ |
| SSLSocketFactory caching across connections | ✅ |
| Dynamic cert pinning (add garbage cert, existing pins survive) | ✅ |
| `testServer()` diagnostics (TLS version, cert PEM, fingerprint) | ✅ |
| End-to-end `.bit` resolution (`testls.bit` → pubkey) | ✅ |
| Custom server TOFU flow (test → pin cert → resolve) | ✅ |

### 🔧 Targeted by code changes (not emulator-testable)
| Device | OS | Issue addressed |
|--------|-----|-----------------|
| Samsung Galaxy A15 5G (SM-A156E) | Android 16 / One UI 7 (`A156EDXS9DZB3`) | Knox trust-all TrustManager rejection |
| Samsung Galaxy S24 Ultra (SM-S938B) | Android 16 / One UI 7 (`S938BXXS8BZB5`) | Knox trust-all TrustManager rejection |
| Xiaomi (MIUI/HyperOS) | Android 16 | KeyStore type fallback, TLSv1.2 enforcement |
| OnePlus (ColorOS) | Android 16 | TLSv1.2 protocol negotiation, SNI handling |
| Google Pixel | GrapheneOS | Hardened Conscrypt trust-all rejection, .onion fallback |

### 📱 Real-device testing (pending)
| OEM | OS variant | Status |
|-----|-----------|--------|
| Google | Stock Android | ✅ Emulator, API 36 |
| Google | GrapheneOS | 🔲 |
| Samsung | One UI | ✅ Galaxy A15 (SM-A156E), Android 16, One UI 8.0 |
| Xiaomi | MIUI/HyperOS | 🔲 |
| Poco | MIUI | 🔲 |
| OnePlus | OxygenOS | 🔲 |
| Oppo | ColorOS | 🔲 |
| Vivo | FuntouchOS | 🔲 |
| Nothing | Nothing OS | 🔲 |

### 🧅 Tor .onion verification
| Server | Method | Result |
|--------|--------|--------|
| `i665jpw...sid.onion:50002` | Tor SOCKS proxy (`127.0.0.1:9050`) | ✅ TLSv1.3, same cert as clearnet |

## Files changed
| File | Change |
|------|--------|
| `ElectrumXClient.kt` (quartz) | Pinned certs, OEM compat, `testServer()`, .onion pinned-first fallback, TOFU |
| `ElectrumXServer.kt` (quartz) | `ServerTestResult` data class |
| `NamecoinSettings.kt` | Custom servers → `usePinnedTrustStore=true` |
| `NamecoinSettingsSection.kt` | Test Connection + cert confirmation dialog |
| `NamecoinSettingsTest.kt` | Updated assertions for field rename |
| `NamecoinSettingsScreen.kt` | Wire `onTestServer` + `onPinCert` |
| `NamecoinSharedPreferences.kt` | Persist pinned certs |
| `AppModules.kt` | Expose client, load pinned certs on startup |
| `AppNavigation.kt` | Pass client to settings screen |
| `strings.xml` | 15 new string resources |

## Security
This change **improves security**: instead of accepting any certificate (MITM-vulnerable), we now pin specific server certs + system CAs. The `.onion` path prefers pinned certs and only falls back to trust-all (safe: Tor provides authentication). TOFU pinning requires explicit user confirmation with full SHA-256 fingerprint display.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
@@ -30,23 +30,33 @@ import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.LocalTextStyle
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import com.vitorpamplona.amethyst.commons.model.nip05DnsIdentifiers.Nip05State
 import com.vitorpamplona.amethyst.model.LocalCache
 import com.vitorpamplona.amethyst.model.User
 import com.vitorpamplona.amethyst.service.relayClient.searchCommand.UserSearchDataSourceSubscription
 import com.vitorpamplona.amethyst.ui.layouts.listItem.SlimListItem
 import com.vitorpamplona.amethyst.ui.note.ClickableUserPicture
+import com.vitorpamplona.amethyst.ui.note.ObserveAndRenderNIP05VerifiedSymbol
 import com.vitorpamplona.amethyst.ui.note.UsernameDisplay
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.AccountViewModel
-import com.vitorpamplona.amethyst.ui.screen.loggedIn.qrcode.WatchAndDisplayNip05Row
 import com.vitorpamplona.amethyst.ui.theme.DividerThickness
+import com.vitorpamplona.amethyst.ui.theme.Font14SP
+import com.vitorpamplona.amethyst.ui.theme.NIP05IconSize
 import com.vitorpamplona.amethyst.ui.theme.Size55dp
+import com.vitorpamplona.amethyst.ui.theme.nip05
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
@@ -143,5 +153,53 @@ fun UserLine(
                 WatchAndDisplayNip05Row(baseUser, accountViewModel)
             }
         },
+    )
+}
+
+@Composable
+private fun WatchAndDisplayNip05Row(
+    user: User,
+    accountViewModel: AccountViewModel,
+) {
+    val nip05StateMetadata by user.nip05State().flow.collectAsStateWithLifecycle()
+
+    when (val nip05State = nip05StateMetadata) {
+        is Nip05State.Exists -> {
+            NonClickableObserveAndDisplayNIP05(nip05State, accountViewModel)
+        }
+
+        else -> {
+            Text(
+                text = user.pubkeyDisplayHex(),
+                fontSize = Font14SP,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+            )
+        }
+    }
+}
+
+@Composable
+private fun NonClickableObserveAndDisplayNIP05(
+    nip05State: Nip05State.Exists,
+    accountViewModel: AccountViewModel,
+) {
+    if (nip05State.nip05.name != "_") {
+        Text(
+            text = remember(nip05State) { AnnotatedString(nip05State.nip05.name) },
+            fontSize = Font14SP,
+            color = MaterialTheme.colorScheme.nip05,
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
+        )
+    }
+
+    ObserveAndRenderNIP05VerifiedSymbol(nip05State, 1, NIP05IconSize, accountViewModel)
+
+    Text(
+        text = nip05State.nip05.domain,
+        style = LocalTextStyle.current.copy(color = MaterialTheme.colorScheme.nip05, fontSize = Font14SP),
+        maxLines = 1,
+        overflow = TextOverflow.Visible,
     )
 }


amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/newUser/ImportFollowListSelectUserScreen.kt
```diff
@@ -252,7 +350,28 @@ fun CustomWatchResponses(
             state = viewModel.listState,
         ) {
             itemsIndexed(suggestions, key = { _, item -> item.pubkeyHex }) { _, item ->
-                UserLine(item, accountViewModel) { onSelect(item) }
+                val isNamecoinResult =
+                    ncState is NamecoinResolutionState.Resolved &&
+                        (ncState as NamecoinResolutionState.Resolved).user == item
+
+                if (isNamecoinResult) {
```
vitorpamplona: There is no need for this separation, right? In the end, they are just showing the same item, which is the user in both cases.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
```diff
@@ -79,6 +101,9 @@ class SearchBarViewModel(
     val invalidations = MutableStateFlow(0)
     val searchValueFlow = MutableStateFlow("")
 
+    /** Tracks Namecoin resolution status for the UI. */
+    val namecoinState = MutableStateFlow<NamecoinResolutionState>(NamecoinResolutionState.Idle)
```
vitorpamplona: Why is this status needed? Is the namecoin resolution slow?  Should we do it for all NIP-05 resolutions? I am not sure if this has been a problem.

quartz/src/jvmAndroid/kotlin/com/vitorpamplona/quartz/nip05DnsIdentifiers/namecoin/ElectrumXClient.kt
```diff
@@ -461,41 +601,298 @@ class ElectrumXClient(
...
+         * PEM-encoded certificates for the well-known Namecoin ElectrumX servers.
+         *
+         * These are self-signed certificates that cannot be verified by the
+         * system CA store. We pin them explicitly so that connections succeed
+         * on devices with strict TLS enforcement (e.g. Samsung One UI 7).
+         *
+         * To update: `echo | openssl s_client -connect HOST:PORT 2>/dev/null | openssl x509 -outform PEM`
+         * For .onion: `python3 -c "import socks,ssl,socket,base64; s=socks.socksocket(); s.set_proxy(socks.SOCKS5,'127.0.0.1',9050); s.connect(('HOST',PORT)); ctx=ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT); ctx.check_hostname=False; ctx.verify_mode=ssl.CERT_NONE; ss=ctx.wrap_socket(s); print(base64.encodebytes(ss.getpeercert(True)).decode())"`
+         */
+        private val PINNED_ELECTRUMX_CERTS =
```
vitorpamplona, line 835: This sucks. Do we really have to keep loading and maintaining PEM certificates for every different Electrum server we need to use? 

Don't we need to capture a certificate every time the user changes their settings to a new server?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
```diff
@@ -66,6 +68,26 @@ import kotlinx.coroutines.flow.onEach
...
 import kotlinx.coroutines.flow.update
 
+/** Returns a [Nip05Id] for identifiers that should go through NIP-05 / Namecoin resolution. */
+private fun toNip05IdOrNull(term: String): Nip05Id? =
+    when {
+        term.contains('@') -> {
+            Nip05Id.parse(term)
+        }
+
+        NamecoinNameResolver.isNamecoinIdentifier(term) -> {
```
vitorpamplona: We should just resove nip05s in the search bar. The other ways to address are less natural for people using Nostr.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
vitorpamplona: always test on real phones in the "benchmark" build variant. The emulator is doing a LOT of things in debug. Easily 10x slower than the release build on startup.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
vitorpamplona: Yes, remember, these are nostr users. They don't know your d/ id/ (they don't even know that's a thing), they are only used to nip05s and more likely than not, not even that. 

Also, remember that this code is getting executed, cancelled and executed again, in every letter they type.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchBarViewModel.kt
vitorpamplona: Get a used one on ebay :)

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchScreen.kt
```diff
@@ -255,6 +256,16 @@ private fun DisplaySearchResults(
             users,
             key = { _, item -> "u" + item.pubkeyHex },
         ) { _, item ->
+            if (namecoinUser == item) {
```
vitorpamplona: No need for this, right? Then you can delete the 

`val namecoinUser by searchBarViewModel.namecoinResolvedUser.collectAsStateWithLifecycle()`

up top and remove the special state management (_namecoinResolvedUser) in the viewModel.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
```diff
@@ -155,51 +145,3 @@ fun UserLine(
         },
     )
 }
-
-@Composable
-private fun WatchAndDisplayNip05Row(
```
vitorpamplona, line 160: What happened here? These are being used, no?

mstrofnone:
Tested on a physical Samsung Galaxy A15 (SM-A156E) running **Android 16 (API 36), One UI 8.0** — cert pinning works, ElectrumX connections succeed, and .bit resolution resolves correctly in the search bar. No TLS handshake failures.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
mstrofnone: ShowUserSuggestionList.kt — The local WatchAndDisplayNip05Row function was already moved to qrcode.WatchAndDisplayNip05Row in main. The file now uses the shared import — no functions were actually lost, the diff just shows the removal of the now-redundant local copy.

quartz/src/jvmAndroid/kotlin/com/vitorpamplona/quartz/nip05DnsIdentifiers/namecoin/ElectrumXClient.kt
mstrofnone: If you can put in a word with the manufacturers listed above that'd be great to avoid this ;) I can't think of another way.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
vitorpamplona: I don't see the function on QRCode.Watch... this PR just deleted this.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchScreen.kt
```diff
@@ -255,6 +257,16 @@ private fun DisplaySearchResults(
             users,
             key = { _, item -> "u" + item.pubkeyHex },
         ) { _, item ->
+            if (isNamecoinSearch) {
```
vitorpamplona: remove this.. there is no need to know that it is a namecoin search... It's just NIP-05

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchScreen.kt
```diff
@@ -229,6 +229,8 @@ private fun DisplaySearchResults(
     val hashTags by searchBarViewModel.hashtagResults.collectAsStateWithLifecycle()
     val relays by searchBarViewModel.relayResults.collectAsStateWithLifecycle()
     val users by searchBarViewModel.searchResultsUsers.collectAsStateWithLifecycle()
+    val currentSearch by searchBarViewModel.searchTerm.collectAsStateWithLifecycle()
```
vitorpamplona: Is this being used somewhere?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/search/SearchScreen.kt
vitorpamplona: It should do that only if the .bit is a valid nip05. Otherwise, it is a regular search.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/creators/userSuggestions/ShowUserSuggestionList.kt
mstrofnone: Sorry you're right — the local WatchAndDisplayNip05Row uses NonClickableObserveAndDisplayNIP05 (plain text domain), while the one in ShowQRScreen uses the clickable ObserveAndDisplayNIP05. They're not the same function. Restored the local copy in 7293b09.

quartz/src/jvmAndroid/kotlin/com/vitorpamplona/quartz/nip05DnsIdentifiers/namecoin/ElectrumXClient.kt
vitorpamplona: Or they can just buy regular TLS chains. I think I am starting to understand why nobody uses namecoin.

davotoula: Where will this python scripts be executed?

vitorpamplona: > Trusted third parties are security holes?

It is still trusted by a third party. Now users are trusting Amethyst instead of the regular TLS chain. There is no difference. We now just have a new responsibility that we usually don't have.

PR #1940, opened 2026-03-25
feat: add GIF keyboard support to short new post screen

vitorpamplona:
## Summary

- Adds `contentReceiver` modifier to `MessageField` so that GIFs/images sent from Android's keyboard (e.g. Gboard, SwiftKey) are received directly in the text field
- When a GIF is selected from the keyboard, its URI is extracted and passed to `selectImage()` for the existing upload flow
- The `onReceiveUri` parameter is optional (defaults to `null`) so existing callers of `MessageField` are unaffected

## How it works

The Compose Foundation `contentReceiver` API intercepts rich content (images/GIFs) committed by the soft keyboard's `InputConnection.commitContent()`. Previously, GIF input only worked via SwiftKey's intent-based workaround (`ACTION_SEND`). This change adds native support for all keyboards that use the standard Android rich content API.

## Test plan

- [ ] Open the short new post screen
- [ ] Tap the GIF button on Gboard (or any keyboard with GIF support)
- [ ] Select a GIF — it should appear in the media attachment area
- [ ] Verify the GIF can be uploaded and posted normally
- [ ] Verify that the existing SwiftKey intent-based GIF handling still works
- [ ] Verify other post screens (comments, DMs, classifieds) are unaffected

https://claude.ai/code/session_014FZPV258VfdeE1mcyeU7wU


davotoula:
How will this work? Is the dependant on what keyboard is used? 

I use swift keyboard and I can select a gif and it will populate the new note with an upload of the gif. One click later it's posted. 

I guess I can try with gboard as well

vitorpamplona:
We have to migrate all of our text fields to the new API with the new state. That API has an easier time working with keyboards. I suspect Swift will work as well.

vitorpamplona:
redoing this

PR #1974, opened 2026-03-28
Add explicit permissions to GitHub Actions workflows

greenart7c3:
## Summary
This PR adds explicit permission declarations to all GitHub Actions workflows to follow the principle of least privilege and improve security posture.

## Key Changes
- **crowdin.yml**: Added `contents: write` and `pull-requests: write` permissions to support creating pull requests with translated content
- **build.yml**: Added `contents: read` permission for read-only access needed during the build process
- **create-release.yml**: Added `contents: write` permission to allow creating release artifacts

## Implementation Details
These permission declarations follow GitHub's security best practice of explicitly specifying the minimum required permissions for each workflow, rather than relying on default permissions. This reduces the potential impact of compromised workflow tokens by limiting their scope to only what is necessary for each workflow's operations.

https://claude.ai/code/session_01J83RSnhi3WBRUXxyj9DdaC

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ on:
   push:
     branches: [main]
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -5,6 +5,9 @@ on:
     tags:
     - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
 
+permissions:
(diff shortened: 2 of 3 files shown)


TotallyAvailable:
If you want to drive the point home

```https://github.com/dependabot/dependabot-core/commit/c69174a5a3ca14c953e2ae96a05a5d3b725a1d58``` fixed in v0.367.0

The topic of best practices. I wonder who ***wasn't*** in the room with the big players when the topic *common sense defaults in production* came up. Not like they can't default to ```{}``` across the platform after several "incidents"... that's just the tip of the iceberg anyway, no secrets being revealed here. Thanks for caring by the way. Couple more off the list 🥳

Issue #1946, opened 2026-03-25
Use toast instead of dialog for download completion


dergigi:
When saving an image or video, the app shows a toast for "download started" which is great. But when the download finishes, it shows a modal dialog that says "Image saved to the phone's photo gallery" with an OK button. This interrupts whatever you're doing and forces a tap to dismiss.

It would be nice if the success message was a toast too, same as the download-started one. No need to block the UI for a simple confirmation.

**Download started (toast, good):**

![download-started](https://files.catbox.moe/hed18d.png)

**Download finished (modal dialog, disruptive):**

![download-finished](https://files.catbox.moe/ko82hm.png)

davotoula:
released

PR #1980, opened 2026-03-28
Show toast instead of dialog on media download success

davotoula:
Replace the interruptive AlertDialog with a brief Android Toast when media downloads complete. Errors still show the full dialog for visibility.


davotoula:
fix for #1946

PR #1983, opened 2026-03-28
Add always-on notification service for real-time Nostr relay connections

vitorpamplona:
## Summary
Implements a comprehensive always-on notification system that maintains persistent WebSocket connections to the user's inbox relays, enabling real-time delivery of direct messages, zaps, and mentions even when the app is backgrounded.

### New Services & Components
- **NotificationRelayService**: Foreground service that keeps inbox relay connections alive by collecting the shared NostrClient's relayServices flow. Uses Android 15's `specialUse` foreground service type to avoid time limits.
- **NotificationCatchUpWorker**: WorkManager periodic task (15-minute intervals) that acts as a safety net to restore connections if the foreground service is killed by OEM battery optimizers.
- **ServiceWatchdogManager**: AlarmManager-based watchdog that checks every 5 minutes if the service should be running and restarts it if needed.
- **BootCompletedReceiver**: Restarts the notification service after device reboot.
- **AlwaysOnNotificationServiceManager**: Orchestrates all five layers of the notification system, watching the account's `alwaysOnNotificationService` setting and enabling/disabling all components accordingly.

### UI Updates
- Added toggle switch in AppSettingsScreen to enable/disable the always-on notification service
- New setting stored in AccountSettings with persistent state management

### Manifest & Permissions
- Added `FOREGROUND_SERVICE_SPECIAL_USE` permission
- Added `RECEIVE_BOOT_COMPLETED` permission
- Registered NotificationRelayService with `specialUse` foreground service type
- Registered BootCompletedReceiver and WatchdogReceiver broadcast receivers
- Added special use property describing the service's purpose

### Configuration
- Added WorkManager dependency (`androidx.work.runtime.ktx`)
- New notification channel for the relay service with low importance

## Implementation Details
- The service shares the same NostrClient instance as the UI, so when the app backgrounds, UI subscriptions drop but the service's subscriptions remain, keeping connections alive without reconnection overhead.
- Uses coroutine flows to monitor relay connection state and update the foreground notification with connected relay count.
- Notification includes action button to stop the service and tap-to-open action to return to the app.
- All components respect the user's setting and can be toggled on/off in real-time.

https://claude.ai/code/session_01LEPfmgGnwjB9a5SDFw5U8t

### DIFF
--- a/PULL_NOTIFICATION.md
+++ b/PULL_NOTIFICATION.md
@@ -1,0 +1,276 @@
+# Always-On Notification Service
+
+Amethyst's always-on notification service maintains persistent WebSocket connections to the
+user's inbox relays, ensuring real-time delivery of DMs, zaps, mentions, and other
+notifications without depending on an external push server.
+
+## Why
+
+The existing push notification system (`push.amethyst.social`) can only monitor relays it
+knows about. Private, paid, or obscure inbox relays get missed entirely. The only way to
+guarantee 100% notification coverage is for the device itself to maintain connections to
+the user's NIP-65 inbox relays.
+
+## Architecture
+
+```
+User's inbox relays <──WebSocket──> [NotificationRelayService]
+                                          |
+                                          v
+                                   EventNotificationConsumer
+                                          |
+                                          v
+                                   Android Notification
+```
+
+The service shares the **same `NostrClient` instance** as the UI. This is the key design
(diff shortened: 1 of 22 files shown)


KotlinGeekDev (approved):
Cool! Looks good to me. cc @greenart7c3

greenart7c3 (approved):

PR #2002, opened 2026-03-29
fix: disable snap animation on DisappearingScaffold scroll

vitorpamplona:
Pass snapAnimationSpec = null to both top and bottom scroll behaviors
so bars stay at their current position when the user releases the drag
instead of snapping to fully visible or fully hidden.

https://claude.ai/code/session_013pVpWCotXGHcUq7u4ZNqbZ


vitorpamplona:
Didn't help

PR #2003, opened 2026-03-29
fix: resolve scroll jitter at top bar collapse boundary in DisappearingScaffold

vitorpamplona:
The top bar's onPreScroll was consuming the entire available scroll delta
even when the heightOffset was clamped at its limit. This caused a visible
pause at the exact point the top bar became fully collapsed, because scroll
was being "eaten" without any visual effect before finally flowing through
to the bottom bar. Now only the actually consumed delta is reported, allowing
leftover scroll to immediately reach the bottom bar for smooth continuous motion.

https://claude.ai/code/session_01QD9ch2SqMsS9uC8ZUthCs6


vitorpamplona:
Didn't help

PR #2005, opened 2026-03-29
Introduce log level filtering

davotoula:
Existing problem: debug level logs evaluated and logged in release client which has impact on performance

### Summary
                                                                                                                                                                                                           
  - Introduce log-level filtering — adds a LogLevel enum (DEBUG, INFO, WARN, ERROR) and a Log.minLevel gate so **release builds can suppress verbose logging** (e.g. set to WARN to silence debug/info output)
  - Separate Log from platform concerns — replaces the previous expect/actual Log with a new expect/actual PlatformLog (delegates to android.util.Log, NSLog, println) and makes Log a common-only filtered
   wrapper                                                                                                                                                                                                 
  - Convert ~200 interpolated log calls to lambda overloads — Log.d(tag) { "msg $expensive" } avoids string allocation when the log level is filtered out, improving release performance                   
  - Set Log.minLevel in app entry points — Android (Amethyst.kt), Desktop (Main.kt) configure the threshold at startup                                                                                     
  - Migrate remaining android.util.Log usages to the shared Log wrapper across all modules                                                                                                                 
                                                                                                                                                                                                           
  ### Changes                                                                                                                                                                                                  
                                                                                                                                                                                                           
  - 120 files changed (+465 / -349)                                                                                                                                                                        
  - New files: LogLevel.kt, PlatformLog.kt (expect), PlatformLog.android.kt, PlatformLog.jvm.kt, PlatformLog.apple.kt (actuals)
  - Removed: old Log.android.kt, Log.jvm.kt, Log.apple.kt (actuals)                                                                                                                                        
                                                                                                                                                                                                           
  ### Test plan                                                                                                                                                                                                
                                                                                                                                                                                                           

- [x]   - All existing tests pass                                                                                                                                                                              
- [x]   - assemblePlayDebug compiles successfully
- [x]   - Verify debug build still shows Log.d output                                                                                                                                                            
- [x]   - Verify release build suppresses debug/info logs

### Usage                                                                                                                                                                                                    

minLevel is defined in: amethyst/src/main/java/com/vitorpamplona/amethyst/Amethyst.kt

Any code path that doesn't explicitly set minLevel (like benchmark tests) will keep all logs enabled
                                                                                                                                                                                                           
When logging messages that involve string interpolation or expensive operations, prefer the lambda overload to avoid unnecessary allocation when the level is filtered out:                            
                                                                                                                                                                                                           
  ```
// Preferred — lambda is only evaluated if DEBUG level is enabled
  Log.d("MyTag") { "Processing event ${event.id} from ${relay.url}" }                                                                                                                                      
                                                                                                                                                                                                           
  // Avoid in hot paths — string is always allocated even if filtered                                                                                                                                      
  Log.d("MyTag", "Processing event ${event.id} from ${relay.url}") 
```                                                                                                                                        
                                                                                                                                                                                                           
  The plain string overload is fine for static messages with no interpolation:                                                                                                                             
   
`  Log.d("MyTag", "Initialization complete") `

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/notifications/RegisterAccounts.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/notifications/RegisterAccounts.kt
@@ -132,7 +132,7 @@ class RegisterAccounts(
         val client = client(url)
 
         client.newCall(request).executeAsync().use { response ->
-            Log.i(tag, "Server registration ${response.isSuccessful}")
+            Log.i(tag) { "Server registration ${response.isSuccessful}" }
         }
     }
 
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/VideoCompressionHelper.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/VideoCompressionHelper.kt
@@ -299,7 +299,7 @@ object VideoCompressionHelper {
                 if (cursor.moveToFirst()) cursor.getLong(sizeIndex) else 0L
             } ?: 0L
         } catch (e: Exception) {
-            Log.w(LOG_TAG, "Failed to get file size: ${e.message}")
+            Log.w(LOG_TAG) { "Failed to get file size: ${e.message}" }
             0L
         }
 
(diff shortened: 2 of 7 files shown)


davotoula:
@vitorpamplona ready for review/merge

PR #2022, opened 2026-03-30
Add NIP-BE Bluetooth Low Energy mesh networking support

vitorpamplona:
## Summary
Implements complete NIP-BE (Nostr over Bluetooth Low Energy) support, enabling Nostr clients to communicate via BLE mesh networking. This includes platform-specific BLE transport, half-duplex message synchronization, and integration with the existing relay infrastructure.

### Core BLE Transport & Protocol
- **BleTransport** interface: Platform-agnostic abstraction for BLE GATT operations (advertising, scanning, client/server connections)
- **AndroidBleTransport**: Full Android implementation using native BLE APIs with proper permission handling
- **BleNostrClient**: Implements `IRelayClient` to allow BLE peers to integrate seamlessly into the relay pool
- **BleNostrServer**: Handles server-side BLE connections, receiving commands and sending notifications
- **BleMeshManager**: High-level orchestrator managing peer discovery, role assignment, connection lifecycle, and event broadcasting

### Message Handling
- **BleMessageChunker**: Splits JSON messages into DEFLATE-compressed BLE-sized chunks with metadata (index + total count)
- **BleChunkAssembler**: Reassembles chunks back into complete messages with timeout handling
- **Deflate** (multi-platform): Compression/decompression using native zlib bindings for JVM/Android, Apple, and Linux

### Role Assignment & Configuration
- **BleRole**: Deterministic client/server role assignment based on UUID comparison (higher UUID = server)
- **BlePeer**: Represents a discovered peer with UUID, role, and platform handle
- **BleConfig**: Centralized configuration (service UUID, characteristic UUIDs, default chunk size)

### Testing
- **BleMessageChunkerTest**: Validates chunking/reassembly round-trips and chunk format
- **BleChunkAssemblerTest**: Tests single/multi-chunk assembly and timeout behavior
- **BleRoleTest**: Verifies deterministic role assignment
- **DeflateTest**: Confirms compression round-trips across platforms

## Implementation Details

- **Half-duplex synchronization**: Clients write commands, servers notify responses; prevents collision
- **Event deduplication**: Tracks sent event IDs per peer to avoid redundant broadcasts
- **Thread-safe queuing**: Send queues with synchronized access for multi-threaded BLE callbacks
- **MTU negotiation**: Supports dynamic chunk size adjustment based on negotiated MTU
- **Relay pool integration**: BLE clients expose `IRelayClient` interface for use in existing subscription/pool infrastructure
- **Cross-platform compression**: Platform-specific Deflate implementations (JVM, Apple, Linux) with common interface
- **Comprehensive error handling**: Graceful degradation with listener callbacks for connection/parsing failures

## Permissions Required (Android)
- `BLUETOOTH_ADVERTISE`, `BLUETOOTH_CONNECT`, `BLUETOOTH_SCAN` (Android 12+)
- `ACCESS_FINE_LOCATION` (Android < 12)

https://claude.ai/code/session_01Tz5E73Rj7tL48A3qUGS5DT

### DIFF
--- a/quartz/src/androidMain/kotlin/com/vitorpamplona/quartz/nipBEBle/transport/AndroidBleTransport.kt
+++ b/quartz/src/androidMain/kotlin/com/vitorpamplona/quartz/nipBEBle/transport/AndroidBleTransport.kt
@@ -18,7 +18,7 @@
  * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
-package com.vitorpamplona.quartz.nipBEBle
+package com.vitorpamplona.quartz.nipBEBle.transport
 
 import android.annotation.SuppressLint
 import android.bluetooth.BluetoothAdapter
@@ -41,6 +41,10 @@ import android.bluetooth.le.ScanResult
 import android.bluetooth.le.ScanSettings
 import android.content.Context
 import android.os.ParcelUuid
+import com.vitorpamplona.quartz.nipBEBle.AndroidBleTransportContract
+import com.vitorpamplona.quartz.nipBEBle.BleConfig
+import com.vitorpamplona.quartz.nipBEBle.BlePeer
+import com.vitorpamplona.quartz.nipBEBle.BleRole
 import com.vitorpamplona.quartz.utils.Log
 import java.util.UUID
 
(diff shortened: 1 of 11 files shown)


vitorpamplona:
@KoalaSat We found some discrepancy in the NIP: 

Looking back at the NIP-BE spec text vs the example code vs samiz:

Chunk index size (1 byte vs 2 bytes):
The NIP spec text says [batch index (first 2 bytes)], but the example code in the NIP itself uses only 1 byte:

chunkWithIndex[0] = chunkIndex.toByte() // single byte

And samiz also uses 1 byte. So the spec text contradicts its own example code. We aligned with the example code + reference implementation, which is the safer interoperability choice.

PR #2116, opened 2026-04-03
Fix intermittent test failure, DesktopFeedFilters.kt

davotoula:
It's a long shot but this could be causing ClassCastException during GC

 ?: 0L forces Long regardless of nullability, so all comparisons are Long vs Long — no cast issue.


vitorpamplona:
I had a similar issue on mobile once... it did solve it.

davotoula:
I find that odd from a strongly typed language. What a quirk!

Issue #2132, opened 2026-04-04
[BUG] unable to share image 


davotoula:
**Describe the bug**
Unable to share image 

![Screenshot_20260404-165458.png](https://github.com/user-attachments/assets/a0fbb606-35fc-4e6f-9724-567d3e03efb4)

PR #2133, opened 2026-04-05
Fix image sharing error

davotoula:
fix for #2132

The share coroutine was cancelled before completing. The CancellationException was caught by the generic catch (e: Exception) block in shareImageFile, which showed the "Unable to share image" toast.

manual testing done


vitorpamplona:
@davotoula can you take a look at the conflict? I think this is related to your other PR today

PR #2188, opened 2026-04-09
Optimize secp256k1 field arithmetic with fused multiply-reduce and platform-specific intrinsics

vitorpamplona:
## Summary

This PR significantly optimizes the secp256k1 cryptographic operations by fusing field multiplication and reduction into a single inline function, and dispatching to platform-specific multiply-high intrinsics. The changes reduce Schnorr signature verification time by ~28% on Android and improve performance across all platforms (JVM, Android, Kotlin/Native).

## Key Changes

- **Fused field multiply-reduce**: Combined `U256.mulWide` and `FieldP.reduceWide` into a single `fieldMulReduceWith` inline function that accepts the multiply-high operation as a `crossinline` lambda. This eliminates ~10,000 un-inlined function calls per Schnorr verify on Android by allowing the Kotlin compiler to substitute platform-specific intrinsics at compile time rather than runtime.

- **Platform-specific multiply-high dispatch**:
  - **Android**: Separate private functions per API level (API 35+, 31-34, <31) to avoid per-call API-level branching overhead. Each function is ~200 DEX instructions, within ART's optimization sweet spot.
  - **JVM**: Delegates to the original unfused path using `Long.compareUnsigned`, which HotSpot recognizes as a JIT intrinsic.
  - **Kotlin/Native**: Uses the XOR-with-MIN_VALUE trick for efficient unsigned comparison.

- **Optimized unsigned comparison**: Introduced platform-specific `uLt()` function to replace `.toULong()` comparisons throughout the codebase. Eliminates ~17,800 `ULong.constructor-impl` invokestatic calls per verify on Android.

- **@JvmField annotations**: Added to `MutablePoint`, `AffinePoint`, and `PointScratch` properties to eliminate virtual getter method calls (~7,450+ per verify), compiling property access to direct field reads instead.

- **Inlined wNAF addition**: Unrolled the `addWnafMixedPP` function calls in the scalar multiplication inner loop, eliminating ~364 function calls per verify (70% of wNAF digits are zero).

- **Benchmark infrastructure**: Added comprehensive benchmarks for Kotlin/Native (linuxX64) to measure performance across all platforms consistently.

## Performance Impact

Measured on Pixel 8 (Android 16):
- `signSchnorr`: 186,610 → 109,711 ns (-41%)
- `verifySchnorr`: 160,869 → 135,885 ns (-16%)
- Combined with all optimizations: verify → 116,450 ns (-28% total)

## Implementation Details

The fused approach uses `inline` + `crossinline` to keep the 200-line arithmetic in `commonMain` while platform files provide only the intrinsic dispatch. This avoids maintaining three identical copies while enabling platform-specific optimizations. The separate API-level functions on Android prevent ART's register allocator from producing suboptimal code that would occur with all branches in a single method.

https://claude.ai/code/session_01EMY5RnXb9rnsyU2KbXrSaY

### DIFF
--- a/benchmark/src/androidTest/java/com/vitorpamplona/quartz/benchmark/Secp256k1Benchmark.kt
+++ b/benchmark/src/androidTest/java/com/vitorpamplona/quartz/benchmark/Secp256k1Benchmark.kt
@@ -124,6 +124,13 @@ class Secp256k1Benchmark {
     }
 
     @Test
+    fun verifySchnorrFastOurs() {
+        benchmarkRule.measureRepeated {
+            assertTrue(Secp256k1InstanceOurs.verifySchnorrFast(signature, msg32, xOnlyPub))
+        }
+    }
+
+    @Test
     fun signSchnorrOurs() {
         benchmarkRule.measureRepeated {
             assertNotNull(Secp256k1InstanceOurs.signSchnorr(msg32, privKey, auxRand))
--- a/benchmark/src/androidTest/java/com/vitorpamplona/quartz/benchmark/Secp256k1Benchmark.kt
+++ b/benchmark/src/androidTest/java/com/vitorpamplona/quartz/benchmark/Secp256k1Benchmark.kt
@@ -35,20 +35,17 @@ import org.junit.runner.RunWith
 /**
  * Android microbenchmark for all secp256k1 operations used by Nostr.
  *
- * Uses AndroidX Benchmark (Jetpack Microbenchmark) for accurate, warmed-up
- * measurements on real Android hardware. Results appear in Android Studio's
- * benchmark output with ns/op, allocations, and percentiles.
+ * Measures the same operations as the JVM benchmark (Secp256k1Benchmark.kt) and the
+ * K/Native benchmark (Secp256k1NativeBenchmark.kt) for cross-platform comparability.
  *
  * Tests are structured as matched pairs: "Foo" uses the native C library (ACINQ/JNI),
- * "FooOurs" uses the pure-Kotlin implementation. This allows direct comparison of
- * each operation between native and Kotlin on the same device.
+ * "FooOurs" uses the pure-Kotlin implementation.
  *
  * NOTE: verifySchnorr uses the same pubkey repeatedly, so it benefits from the
  * pubkey decompression cache and P-table cache. This is realistic for Nostr
  * (feed from one author) but not representative of first-time verification.
  *
- * Test data is the same across all 3 platform benchmarks (JVM, Android, K/Native)
- * for cross-platform comparability.
+ * Test data is the same across all 3 platform benchmarks for cross-platform comparability.
  *
  * Run with: ./gradlew :benchmark:connectedAndroidTest
  */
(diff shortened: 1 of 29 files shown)


davotoula:
@vitorpamplona Some results from a code review, not sure whether applicable but since it's a high performance feature maybe worth looking into?

Findings

  CRITICAL (1)

  Non-atomic cache publication — pubkeyCache / pTableCache
  - quartz/src/commonMain/.../secp256k1/Secp256k1.kt:100,119–130
  - quartz/src/commonMain/.../secp256k1/ECPoint.kt:119,629–658

  Both caches are plain Array<…?> on singleton objects, written without synchronization. Under the JMM, a reader thread can observe a non-null CachedPubkey/CachedPTable reference whose fields are still zero-valued (unsafe
  publication). A zero-point (0,0) result in liftXCached would yield silent false accept/reject on signature verification — catastrophic in a crypto lib.

  Fix: use java.util.concurrent.atomic.AtomicReferenceArray on JVM/Android for both caches. @Volatile on the array field alone is insufficient (element writes aren't covered).

  ---
  HIGH (2)

  1. verifySchnorrFast is Kotlin-only, not exposed through Secp256k1Instance
  - Secp256k1.kt:494

  No entry in the expect/Nip01Crypto. On native-backed targets it silently falls back to full verify. Name implies drop-in replacement — callers will be misled. Either add to expect interface or tighten KDoc to mark as
  pure-Kotlin-internal.

  2. signSchnorrWithXOnlyPubKey silently discards the cached pubkey on JVM/Android/Native
  - Secp256k1Instance.android.kt:44–50 (and jvm/native mirrors)

  The native C lib always re-derives the pubkey; the x-only param is unused. Nip01Crypto.signWithPubKey KDoc claims it "Skips the expensive G multiplication" — false on the default platform. Documentation correctness issue
  with perf implications.

  ---
  MEDIUM (3)

  - ScratchLocal.native.kt:28–33 — single shared instance, unsafe for K/N multi-threaded coroutines. Needs @ThreadLocal or per-call allocation on native.
  - ScalarN.kt:278 — while (U256.cmp(out, N) >= 0) subTo(...) — should be if with justification comment, or keep loop + explain why ≥2n is possible.
  - FieldP.kt:337–341 — cs[0]/cs[1] scratch reuse at end of sqrt is correct but a maintenance hazard.

vitorpamplona:
This is basically my playground... secp256k1 in Kotlin is quite fast, but not as fast as in C. I am getting a better idea of why: 
1. The interesting thing is that since Nostr only uses the x part in pubkeys, we can simplify both signing and verification to make them slightly faster. So, there is a chance we can get some performance gains by redeveloping secp256k1 in ways that the original project wouldn't like. 
2. The C library is poorly optimized for ARM64 (they use the wrong internal structure and lack ARM instructions in the generated code. I thought I could get through the optimizations in pure Kotlin, but it turns out the Kotlin compiler also doesn't generate those instructions in the way we want. 
3. The C library doesn't use SHA256 from Android. It uses its own internal code, which is slower. 
4. We can finally develop Batch verification for events with the same pubkey (huge improvements when downloading follows contents or when navigating to the User Profile. 

Some of the 3 and 4 gains are real, but Kotlin, D8 desugaring and Hotspot in the JVM is getting in the way because none of these stacks fully support int128 bits. They have partial support everywhere, but not fully developed. 

So, we might need to rewrite this in C and call it via JNI as we do today with all the optimizations in place. We will keep going...

PR #2189, opened 2026-04-09
Animated gif to mp4 conversion

davotoula:
Decent saving. 3MB .gif to 159kb mp4 (h264)

If this becomes a popular standard we can move it to user settings to "Always convert gif 2 mp4"

Testing:
- Unit tests
- Manual testing of new feature
- Regression testing

https://github.com/user-attachments/assets/19112865-926e-48df-9950-584ce7639d78


vitorpamplona:
We might need to reorganize these options so that they don't jump around so much when they some make others disappear.

davotoula:
> We might need to reorganize these options so that they don't jump around so much when they some make others disappear. 

Yeah sorry, personally I find it snappy when things bounce around like that so it's my fault!

PR #2376, opened 2026-04-12
fix: filter in-progress chess games (kind 64) from home feed (#2191)

mstrofnone:
Fixes #2191

## Problem

Some chess apps (e.g. [nostrchess](https://nostrchess.vercel.app/)) publish a new kind 64 event after every move with the updated PGN. This floods followers' home feeds with a card for every individual move.

## Solution

Added `ChessGameEvent.isCompletedGame()` that performs a fast check on the PGN `[Result "..."]` tag to determine if the game is finished. The home feed filter now only shows completed games (result is `1-0`, `0-1`, or `1/2-1/2`). In-progress games (result `*` or missing result tag) are filtered out of the general home feed.

The dedicated chess feed is **unaffected** — it continues to show all chess games including in-progress ones.

## Changes

- **`ChessGameEvent.kt`**: Added `isCompletedGame()` method using fast string search on PGN content (no regex/full parsing overhead)
- **`HomeNewThreadFeedFilter.kt`**: Changed `noteEvent is ChessGameEvent` to `(noteEvent is ChessGameEvent && noteEvent.isCompletedGame())`


vitorpamplona:
I thought we had fixed this already in the main version ( we have not released the fix yet, so production still shows all games).. Is it still happening? 

If so, why not just remove the game event and just show the LiveChessGameEndEvent on the feed?

PR #2378, opened 2026-04-12
fix: preserve draft note text across process death when switching apps

mstrofnone:
## Problem

When composing a note and switching to another app (e.g. to copy a URL to paste into the note), Android may kill the process due to memory pressure. On returning, the compose screen loses all typed text because `ShortNotePostViewModel` uses a plain `TextFieldState()` with no process death survival mechanism.

The existing NIP-37 draft auto-save is excellent for relay-persisted recovery, but it doesn't help with the immediate UX — the user has to manually find and load their draft from the drafts screen instead of seeing their text right where they left it.

## Fix

Add `SavedStateHandle` to `ShortNotePostViewModel` to persist the message text and draft tag in Android's saved instance state bundle:

- **On every text change**: save message text + draft tag to `SavedStateHandle`
- **On ViewModel creation**: if `SavedStateHandle` contains saved text, initialize `TextFieldState` with it and skip re-loading from route args
- **On cancel/post**: clear saved state to avoid stale drafts

This is the standard Android mechanism for surviving process death — the `SavedStateHandle` is automatically provided by the Navigation component's `ViewModelStoreOwner`.

## Changes

- `ShortNotePostViewModel`: Accept `SavedStateHandle`, save/restore message text + draft tag
- `ShortNotePostScreen`: Skip `load()` when restored from saved state
- `PollPostScreen`: Same fix (shares the ViewModel)

## Testing

1. Open compose screen, type some text
2. Switch to another app
3. Use developer options → "Don't keep activities" or `adb shell am kill com.vitorpamplona.amethyst`
4. Switch back to Amethyst
5. **Before**: empty compose screen. **After**: text is preserved

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/ShortNotePostScreen.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/home/ShortNotePostScreen.kt
@@ -150,21 +150,28 @@ fun ShortNotePostScreen(
     }
 
     LaunchedEffect(postViewModel, accountViewModel) {
-        val baseReplyTo = baseReplyToId?.let { accountViewModel.getNoteIfExists(it) }
-        val quote = quoteId?.let { accountViewModel.getNoteIfExists(it) }
-        val fork = forkId?.let { accountViewModel.getNoteIfExists(it) }
-        val version = versionId?.let { accountViewModel.getNoteIfExists(it) }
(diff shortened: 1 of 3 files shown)


vitorpamplona:
This is strange. Is Amethyst not saving your notes already? By default, we save all draft notes in your relays. If it crashes it can recover from the draft. It shouldn't need any of this. 

Also, this version only saves the text. Nothing else is preserved. So, if you are drafting a poll, this will come back as a twitter post, not a poll, which will confuse people. 

Please test the app before sending these PRs.

Issue #2238, opened 2026-04-11
KMP iOS Migration: commons module — approach discussion


mstrofnone:
## Summary

This is the tracking issue for making Amethyst's `commons` module compile for iOS (`iosArm64` + `iosSimulatorArm64`) via Kotlin Multiplatform. The approach: no `iosApp` module yet, small incremental PRs, move code from `amethyst/` → `commons/` piece by piece, maintain backward compatibility via typealiases at original locations.

**159 PRs open, all passing CI, all mergeable.** 28 PRs created in session 5. Only #2201 (LargeCache fix) has been merged so far. Zero reviews on the iOS migration PRs. More PRs being created now.

---

### Phase 1: KMP Foundation (6 PRs)

| PR | What | Why |
|----|------|-----|
| #2212 | `expect/actual` for `ConcurrentHashMap`, `AtomicBoolean`, `ReentrantLock` | K/N needs native equivalents |
| #2213 | `expect/actual` for `java.net.URI`, `java.io.*` | JVM stdlib unavailable on iOS |
| #2214 | iOS compilation targets (`iosArm64` + `iosSimulatorArm64`) | The "make it compile" change |
| #2217 | iOS WebSocket via `NSURLSessionWebSocketTask` | Relay connections |
| #2221 | Coil 3 via Ktor Darwin engine | Image loading |
| #2240 | Fix suspend calls in MarmotSubscriptionManagerTest | K/N test fix |

### Phase 2: Model & Architecture — IAccount (21 PRs)

| PR | What |
|----|------|
| #2215 | Move 120+ model/state files |
| #2216 | `IAccount` + `ICacheProvider` interfaces; move Account.kt |
| #2218-#2220 | Feed flows, GeoHash, FeedTopNavFilterState |
| #2236 | `IAccountViewModel`, `Route`, `INav`, `EmptyNav` |
| #2237 | IAccount +9 members, +26 drawables |
| #2257 | IAccountViewModel 40+ members |
| #2259 | IToastManager (7 methods) |
| #2260 | IAccount +10 (bookmarkState, isFollowing, etc.) |
| #2261 | Wire IToastManager → IAccountViewModel |
| #2267 | IUiSettingsState + IAccountViewModel expanded |
| #2278 | IAccount v3 (+10 domain methods) |
| #2280 | IAccount v4 (+10 relay/visibility/follow) |
| #2289 | IAccount v5 (+15 bookmark/payment/zap) |
[...]

PR #2379, opened 2026-04-12
refactor(amethyst): convert module to KMP source set structure

vitorpamplona:
Move src/main/ contents to src/androidMain/ and add src/commonMain/
placeholder so code can gradually migrate to shared KMP source sets.
The amethyst module remains a com.android.application (KMP plugin is
not supported for Android applications), but Android sourceSets are
remapped to the new androidMain paths:

- src/main/java/     -> src/androidMain/kotlin/
- src/main/res/      -> src/androidMain/res/
- src/main/jniLibs/  -> src/androidMain/jniLibs/
- src/main/AndroidManifest.xml -> src/androidMain/AndroidManifest.xml
- src/main/ic_launcher-playstore.png -> src/androidMain/

Flavor source sets (play, fdroid), unit tests (test) and instrumented
tests (androidTest, androidTestPlay) are left in place for now.

https://claude.ai/code/session_01EMUGZnYdFAFcQZMh6XGvup


vitorpamplona:
@mstrofnone I will review this later, but this maybe a good start to migrate to a KMP format for the amethyst module

mstrofnone:
This is a clean approach — restructure first, migrate code later. Makes sense as Step 0.

For context: our [164 closed PRs](https://github.com/vitorpamplona/amethyst/issues/2238) worked on the **commons** module side (making it compile for iOS). This PR works on the **amethyst** app module side. They're complementary.

### What we already have that could plug in after this merges

**Small, self-contained pieces (commons module, no amethyst changes):**
- `expect/actual` wrappers for `ConcurrentHashMap`, `AtomicBoolean`, `ReentrantLock`, `java.net.URI`, `java.io.*` — 10 files, lets commons compile for iosArm64 + iosSimulatorArm64
- iOS compilation targets in `commons/build.gradle.kts`
- iOS WebSocket via `NSURLSessionWebSocketTask`, Coil 3 via Ktor Darwin engine
- 1,978 strings + 77 translations already ported to `commons/composeResources/` (additive, doesn't touch existing files)
- Theme tokens (Color, Shape, Type) moved to commons

**Reference material (approach would need to match your style):**
- R.string → Res.string mapping for 900+ files (we have the mapping, can be applied file-by-file instead of in bulk)
- 56 DAL FeedFilter subclasses ported to commons
- Interface designs (IAccount 150+ members, ICacheProvider, IAccountViewModel) — useful as reference for what needs abstracting, though you may prefer moving concrete classes directly

**Consolidated branch:** All 106 mergeable branches are combined in [`kmp-ios-consolidated`](https://github.com/mstrofnone/amethyst/tree/kmp-ios-consolidated) on the fork (1,586 files, build passes). Available to diff or cherry-pick from.

### Kotlin/Native gotchas we hit (useful when iOS code starts)
- `setUnhandledExceptionHook` is the only reliable way to prevent K/N `abort()` on unhandled coroutine exceptions — `SupervisorJob` + `CoroutineExceptionHandler` alone is NOT sufficient
- Network.framework types can't bridge to `NSObject` — need C interop (`NWHelpers.def`)
- `NSLog` with `%@` + Kotlin strings causes SEGFAULT — use string concatenation
- `SecureTransport` (`SSLCreateContext`) broken on iOS for self-signed certs
- Every `launch{}` and `LaunchedEffect{}` needs `try-catch` (rethrow `CancellationException`)

Happy to repackage any of the above in whatever format works once this lands. The minimal next step would be a single PR that adds iOS targets to commons with just the expect/actual wrappers (~20 files, zero composable moves).

KotlinGeekDev (approved):
Looks good to me.

PR #2381, opened 2026-04-13
feat: Desktop Tor support — embedded kmp-tor, full privacy routing, settings UI

nrobi144:
## Summary

Add embedded Tor support to Amethyst Desktop with full granular routing parity with Android.

- **kmp-tor** embedded daemon (Apache 2.0, `resource-exec-tor`)
- **All HTTP traffic** routed through SOCKS proxy when Tor enabled (relays, images, GIFs, DM media, zaps, uploads, NIP-46 bunker)
- **Fail-closed**: during Tor bootstrap, dead proxy prevents clearnet leaks
- **Settings UI**: shield icon in sidebar, inline toggle + full Advanced dialog with presets
- **Shared code**: `TorSettings`, `TorRelayEvaluation`, `TorServiceStatus`, `ITorManager` extracted to commons/commonMain
- **Android unchanged**: no regressions, keeps tor-android/jtorctl

## Key Architecture

```
commons/commonMain/tor/     — Shared models, interfaces, evaluation logic (7 files)
desktopApp/tor/             — DesktopTorManager (kmp-tor), DesktopTorPreferences
desktopApp/network/         — Proxy-aware DesktopHttpClient with fail-closed
desktopApp/ui/tor/          — StatusIndicator, SettingsSection, SettingsDialog, LocalTorState
```

## Tor Routing Verified

```
Real IP:  212.93.150.16
Tor IP:   45.84.107.182
IsTor:    true
Direct connections with Tor ON: 0 (relay traffic)
```

## Testing

- 130+ unit tests (commons + desktop + Android)
- Manual lsof verification: zero direct relay connections when Tor ON
- Toggle OFF→ON→OFF cycle works via app rebuild (key() recomposition)
- Settings persist across restarts

## Known Limitations

- 2-4 stale Coil CDN image connections may persist briefly after toggle (not relay traffic, close after 5min keep-alive)
- DMs show "Recipient has no DM relay list" in Tor mode (investigation needed)
- Hot-toggle requires app rebuild (restart dialog) — relay subscriptions don't survive live reconnect

- **Verify**: `lsof -i TCP -P -n | grep "^java" | grep ESTABLISHED | grep -v "127.0.0.1"` shows zero direct when Tor ON
- **Verify**: `curl --socks5-hostname 127.0.0.1:PORT https://check.torproject.org/api/ip` returns `IsTor: true`
- **Failure signal**: Direct connections visible in lsof with Tor enabled

---

[Compound Engineered](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/account/AccountManager.kt
+++ b/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/account/AccountManager.kt
@@ -138,7 +138,7 @@ class AccountManager internal constructor(
     private suspend fun getOrCreateNip46Client(): INostrClient =
         nip46ClientMutex.withLock {
             nip46Client ?: NostrClient(
-                BasicOkHttpWebSocket.Builder(DesktopHttpClient::getHttpClient),
+                BasicOkHttpWebSocket.Builder(DesktopHttpClient::getSimpleHttpClient),
             ).also {
                 nip46Client = it
                 it.connect()
--- a/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/account/AccountManager.kt
+++ b/desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/account/AccountManager.kt
@@ -138,7 +138,7 @@ class AccountManager internal constructor(
     private suspend fun getOrCreateNip46Client(): INostrClient =
         nip46ClientMutex.withLock {
             nip46Client ?: NostrClient(
-                BasicOkHttpWebSocket.Builder(DesktopHttpClient::getSimpleHttpClient),
+                BasicOkHttpWebSocket.Builder { url -> DesktopHttpClient.currentClient() },
             ).also {
                 nip46Client = it
                 it.connect()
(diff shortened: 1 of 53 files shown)


desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/account/AccountManager.kt
```diff
@@ -138,7 +138,7 @@ class AccountManager internal constructor(
     private suspend fun getOrCreateNip46Client(): INostrClient =
         nip46ClientMutex.withLock {
             nip46Client ?: NostrClient(
-                BasicOkHttpWebSocket.Builder(DesktopHttpClient::getHttpClient),
+                BasicOkHttpWebSocket.Builder { url -> DesktopHttpClient.currentClient() },
```
KotlinGeekDev, line 141: Interesting. So the URL is not needed when making a new client object?

KotlinGeekDev (approved):

vitorpamplona:
This looks good to me. But FYI, we moved away from the C-Tor dependencies because they were running into seg faults all the time. Maintaining the C code alive has been quite hard on Android. And because it was a seg fault, it simply kills the entire app without any chance of try catch. 

It's quite the same API, and Rust Tor is better for embedding into apps. The C Tor wasn't designed to be embedded, and conflicts between multiple apps trying to run their own embedded Tor at the same time. 

https://github.com/vitorpamplona/amethyst/pull/2063

nrobi144:
Thanks for the heads up — that makes sense. I've opened #2385 to track porting the Arti JNI approach from Android to Desktop, replacing kmp-tor (C Tor) entirely. The Rust crate in `tools/arti-build/` looks platform-agnostic so the main work would be cross-compilation for desktop targets + packaging the native libs.

PR #2475, opened 2026-04-21
fix:  the consumer read activity.intent instead of the intent parameter passed into the callback

davotoula:
@vitorpamplona I haven't verified what the bug was so not able to verify this fixes it. It was flagged during some other work, want to have a look?

1. fix a latent bug: the consumer read activity.intent instead of the
  intent parameter passed into the callback. activity.intent is the
  Activity's launch intent, not the new one that triggered the listener,
  so subsequent intents would reuse the first intent's URI.

2. The DisposableEffect's OnNewIntentListener is invoked on the main thread
  when a new ACTION_SEND intent arrives (e.g. SwiftKey sharing a GIF). It
  called contentResolver.getType() synchronously, which can Binder-IPC to
  a remote content provider and block for tens to hundreds of milliseconds.

  Use rememberCoroutineScope() + scope.launch(Dispatchers.IO) since
  DisposableEffect has no coroutine scope of its own. The scope is tied to
  composition and cancels cleanly on disposal.


vitorpamplona:
this does make sense.

PR #2503, opened 2026-04-22
feat(media): add playback controls and autoplay support for GIFs

greenart7c3:
Introduces `GifVideoView` to manage GIF playback with manual play/pause controls and support for the "Auto-play Videos" setting. Updates `MyAsyncImage` and `ZoomableContentView` to utilize this specialized view when GIF content is detected via file extension or MIME type.

Key changes include:
*   **GifVideoView**: A new component that uses Coil's `Animatable` to start and stop GIF animations, featuring a video-like UI with play/pause overlays and a zoom button.
*   **GIF Detection**: Added `isGif()` and `isGifUrl()` helpers to identify GIF content by MIME type or URL patterns.
*   **Integration**: Redirects GIF rendering in `MyAsyncImage` and `ZoomableContentView` from static image views to the new interactive `GifVideoView`.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/GifVideoView.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/GifVideoView.kt
@@ -1,0 +1,185 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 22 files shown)


vitorpamplona:
hum.. does it make sense to have play pause in gifts? Aren't most of these just memes and so aren't we just polluting the view of a short 3-5 seconds full screen "emoji"?

greenart7c3:
I thought this is what you wanted when you said we needed the auto play option to also work for gifs, for me it doesn't makes sense

vitorpamplona:
ohh,, I see, so you added just because of the settting? Interesting... if users play Gifs as default, does is it supposed to show the play/pause button? that's the part that doesn't need to exist. Maybe this new path is just for those that disabled auto-playback so that they can play the gif?

vitorpamplona:
Weird part is that I was imagining that the auto-playback for Gifs would affect every image, like User Profiles too. But those are too small to have a Play button. Maybe in those cases, there is no way to play the gif? 

PS: I am coming from the angle of dyslexic people struggling with too many gifts on the app.

greenart7c3:
Currently it always shows the buttons and it fades after a few seconds, but I can change it.
I could also not have the buttons at all and only play it after the user pressed the gif of the setting is enabled

greenart7c3:
Maybe when the setting is enabled only play the gif in the full screen dialog

vitorpamplona:
I think the current experience shouldn't change (auto-play no controls). But if the user wants something different then we can have these paths. I am fine with either (buttons or dialog play). I am not sure what works best and what are these users expecting from the playback.

davotoula:
I'm not sure full media controls are appropriate for gifs. Maybe a user settings to auto play gifs or not and just a play button that auto hides on the gif?

greenart7c3:
Was trying some other social medias and what they usually do is no control if automatic and only a play button that open the full screen dialog when the settings is on

vitorpamplona:
is this ready to merge?

greenart7c3:
Yes

Issue #2518, opened 2026-04-23
[FEATURE] support HLS video inside a BUD-10 blossom URI


davotoula:
HLS video inside a BUD-10 blossom URI

Exoplayer is not playing these hls videos

vitorpamplona:
This was implemented (fix merged) and has shipped — HLS inside BUD-10 blossom URIs is supported now. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2520, opened 2026-04-23
fix(playback): route HLS URIs to HlsMediaSource via MediaItem mimeType

davotoula:
ExoPlayer's URI-based content-type inference only fires for http(s) schemes + known extensions, so HLS playback fails in two cases Amethyst ships regularly:

  1. NIP-71 video events with BUD-10 blossom URIs. Hash-based URIs carry no extension and the imeta's application/vnd.apple.mpegurl never reaches the MediaItem — MediaItemCache dropped it.
  2. Bare BUD-10 URIs pasted into kind 1 posts, which have no imeta at all. Even though the URI contains `.m3u8` in its path, ExoPlayer's inference skips custom schemes (`blossom:`).

Forward a normalized mimeType onto MediaItem.Builder.setMimeType so DefaultMediaSourceFactory routes to HlsMediaSource regardless of scheme or extension:

  - Normalize IANA Apple HLS variants (application/vnd.apple.mpegurl, application/x-mpegurl, audio/x-mpegurl, audio/mpegurl) to MimeTypes.APPLICATION_M3U8.
  - Fall back to `.m3u8` URI detection when no imeta mimeType is present. Matches the existing isLiveStreaming(url) heuristic.

Adds diagnostic logs at MediaItemCache.compute and CustomMediaSourceFactory.createMediaSource so the routing decision is observable in logcat, matching the diagnostic-logs pattern already used elsewhere in playback.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/playback/composable/mediaitem/MediaItemCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/playback/composable/mediaitem/MediaItemCache.kt
@@ -24,7 +24,10 @@ import android.os.Bundle
 import androidx.core.net.toUri
 import androidx.media3.common.MediaItem
 import androidx.media3.common.MediaMetadata
+import androidx.media3.common.MimeTypes
+import androidx.media3.common.util.UnstableApi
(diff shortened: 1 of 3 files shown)


davotoula:
fix for #2518

PR #2542, opened 2026-04-24
add SECURITY.md with private vulnerability reporting policy

davotoula:
This would be a start to enable the GitHub security fixes workflow: allow private submission of security issues rather than public issues, allow for private contributions from private repos to avoid disclosure

@vitorpamplona:
- Go to Security tab → Security policy and confirm GitHub recognises the SECURITY.md
- Enable private vulnerability reporting (Settings → Security → toggle "Private vulnerability reporting") so the "Report a vulnerability" button appears publicly
- Optionally assign me the Security Manager role (Settings → Collaborators and teams) — this lets me create and manage advisories myself without needing you each time
- Optionally we can add nostr npubs for security DMs but I'm happy just using Github 

thoughts?

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,95 @@
+# Security Policy
+
+Amethyst is a Nostr client that handles user private keys, signed events, and
+end-to-end encrypted direct messages. We take security reports seriously and
+appreciate responsible disclosure.
+
+## Supported Versions
+
+Only the latest release receives security fixes. We do not backport patches to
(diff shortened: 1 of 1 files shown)


vitorpamplona:
Looks good, I just couldn't find the option to add you a Security Manager. Maybe I need to create a github org first..

davotoula:
No problem.

I may be sending some security fixes so I'll try out the "private contribution" flow.

davotoula:
No problem.

I may be sending some security fixes so I'll try out the "private contribution" flow.

PR #2547, opened 2026-04-24
feat(reactions): place payment targets icon after zap by default

greenart7c3:


vitorpamplona:
Can we hide the icon by default for people who are already using Amethyst? They can activate it on the settings. I have a feeling that they would want to remove zaps when activating payto anyway.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSyncedSettingsInternal.kt
```diff
@@ -62,8 +62,8 @@ val DefaultReactionRowItems =
         ReactionRowItem(ReactionRowAction.Boost),
         ReactionRowItem(ReactionRowAction.Like),
         ReactionRowItem(ReactionRowAction.Zap),
+        ReactionRowItem(ReactionRowAction.Pay, enabled = false, showCounter = false),
```
vitorpamplona, line 65: This will always be false regardless of the settings, no?

greenart7c3: This is just the defaults, if the user has configured the reactions row it will get from the settings event

vitorpamplona: ohhh I missed that!

greenart7c3:
I realized now that when the wallet screen was changed the option to set payment targets was removed

vitorpamplona:
hum.. not on purpose? I am not sure what happened.

PR #2565, opened 2026-04-25
Ci/split android job

davotoula:
CI: split Android into its own visible job

 The merged test-and-build matrix hid Android behind an OS-keyed leg. Restore visibility by splitting into two parallel jobs after lint:                                                                                       
   
  - build-desktop (matrix: ubuntu/macos/windows) — JVM tests for KMP/JVM modules + native installer per OS.                                                                                                                     
  - test-and-build-android (ubuntu) — full test sweep + assembleBenchmark, uploads APKs and reports.
                                                                                                                                                                                                                                
Also fixes a latent CRLF bug in InliningTagArrayPrettyPrinterTest surfaced now that :quartz:jvmTest actually runs on Windows. 

<img width="755" height="476" alt="Screenshot 2026-04-25 at 15 44 15" src="https://github.com/user-attachments/assets/b1dae0f2-9fc4-4555-abbe-fc0171882155" />

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
           cache: gradle
 
       - name: Test + Build Desktop (gradle)
-        run: ./gradlew :quartz:test :commons:test :nestsClient:test :cli:test :desktopApp:test :desktopApp:${{ matrix.desktop-task }} --no-daemon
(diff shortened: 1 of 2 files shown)


davotoula:
Adding a .gitattributes file at the repo root would be a more permanent fix for CRLF issues but it would change developer set-up on first run.                                                                                                                                                                   
                                                                                                                                                                                                                              
  * text=auto eol=lf 
  * *.kt text eol=lf
  * ...

PR #2575, opened 2026-04-25
feat(settings): split broadcast tracker visibility from Complete UI mode

vitorpamplona:
Adds a dedicated "Show broadcast tracker" toggle so the broadcast progress
banner and tracked broadcasting can be enabled independently of the
Complete UI mode. Defaults to ALWAYS so users get the same experience as
Complete UI mode by default while freeing Complete UI mode for unrelated
post-style features.


vitorpamplona:
@davotoula

davotoula:
testing it locally

davotoula:
- Fixed a bug: The original `if (!accountViewModel.settings.isCompleteUIMode())` return had a non-reactive .value read; switching feature set in Settings didn't recompose `DisplayBroadcastProgress`, so it stayed detached until process restart
- Test migration (Complete to new version should still see Broadcast Tracker)
- Refactor
- Code review

davotoula:
@vitorpamplona ready for merge

Issue #2579, opened 2026-04-26
[BUG] can't install deb package on debian -missing dependency


utente9382:
I tried installing the debian package but ran into this error:
amethyst:amd64 depends libicu74
      but none of the choices are installable:
      [no choices]

those are the options available on debian stable:
 apt search libicu
icu-devtools/stable 76.1-4 amd64
  utilità di sviluppo per International Components for Unicode

libicu-dev/stable 76.1-4 amd64
  file di sviluppo per i componenti internazionali per Unicode

libicu4j-4.4-java/stable 4.4.2.2-4 all
  libreria per gestione Unicode e internazionalizzazione

libicu4j-java/stable 73.2-1 all
  libreria per gestione Unicode e internazionalizzazione

libicu76/stable,now 76.1-4 amd64 [installato, automatico]
  International Components for Unicode - componenti internazionali per Unicode

davotoula:
Thank you for reporting this. Hopefully we can fix this for next release.

@nrobi144 Investigation:

Root cause: jpackage runs dpkg-shlibdeps against the bundled JDK's native libs (libfontmanager.so etc. link libicu) and pins Depends: libicuNN to the build host. CI builds on ubuntu-latest (24.04 → libicu74), so the .deb refuses to install on Ubuntu 22.04 (libicu70), Debian 12 (libicu72), Debian 11 (libicu67), or Debian 13 (libicu76). Neither jpackage nor the Compose Multiplatform DSL exposes a way to override the auto-generated Depends.

Fix: New [scripts/relax-deb-libicu.sh](https://github.com/davotoula/amethyst/blob/claude/fix-libicu74-dependency-Eg1MR/scripts/relax-deb-libicu.sh) extracts the .deb with dpkg-deb -R, rewrites any libicuNN (or alternation of them) in the control file to libicu66 | libicu67 | libicu70 | libicu72 | libicu74 | libicu76 | libicu77, then repacks. Idempotent; no-op when no libicu dep is present.

davotoula:
See branch: [fix-libicu74-dependency](https://github.com/vitorpamplona/amethyst/tree/fix-libicu74-dependency)

PR #2581, opened 2026-04-26
fix(deb): broaden libicu Depends so .deb installs across Debian/Ubuntu

davotoula:
Draft to test the build

jpackage runs `dpkg-shlibdeps` against the bundled JDK runtime's native libs (libfontmanager.so etc. link libicu) and pins Depends to the build host's libicu version. CI builds on ubuntu-24.04, so the .deb requires libicu74 — uninstallable on Ubuntu 22.04 (libicu70), Debian 12 (libicu72), Debian 11 (libicu67), and Debian 13 (libicu76). Neither jpackage nor the Compose Multiplatform DSL exposes a way to override the auto-generated Depends.

Add scripts/relax-deb-libicu.sh which extracts the .deb with `dpkg-deb -R`, rewrites `libicuNN` (or any alternation thereof) to `libicu66 | libicu67 | libicu70 | libicu72 | libicu74 | libicu76 | libicu77`, and repacks. The script is idempotent and a no-op for .debs without a libicu Depends.

Wire it into both release legs (desktopApp + amy CLI) in create-release.yml, and into the desktop test/build leg in build.yml so testers downloading the CI artifact hit the same fix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,16 @@ jobs:
       - name: Test + Build Desktop (gradle)
         run: ./gradlew :quartz:jvmTest :commons:jvmTest :nestsClient:jvmTest :cli:test :desktopApp:test :desktopApp:${{ matrix.desktop-task }} --no-daemon
 
+      # jpackage pins libicu to the build host's version (libicu74 on
+      # ubuntu-24.04). Rewrite the .deb so testers on other Debian/Ubuntu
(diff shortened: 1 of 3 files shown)


davotoula:
Fix for #2579

vitorpamplona:
Should we merge this?

davotoula:
> Should we merge this?

Nope.. I can't test it. I don't have x86 to test it with. 

Just close, can be used later when somebody has time.

PR #2595, opened 2026-04-27
feat(relay): resolve, TLS-pin, and Tor-route .bit Nostr relays via Namecoin

mstrofnone:
## Summary

This PR adds **end-to-end resolution, TLS pinning, and Tor routing for `.bit` Nostr relays**, anchored entirely in the Namecoin blockchain.

A user can drop `wss://example.bit/` into their relay list and Amethyst will:

1. Resolve it at connect time to a real `wss://...` endpoint published in the Namecoin `d/example` record.
2. Pin the TLS handshake against TLSA records in the same record (RFC 6698), so a public-CA MITM can't impersonate the relay even if it has a valid clearnet cert.
3. Route over Tor to a `.onion` if the record advertises one and the user has Tor enabled.

The `.bit` form remains the **canonical** relay identity used by the rest of Amethyst — UI, event tags, relay policy. Only the wire-level handshake URL is rewritten, the way a CDN remaps `Host`/SNI. NIP-05 identity (`alice@relay.example.bit`) goes through the same lookup, so a client that loads both the identity and the relay benefits from a single ElectrumX round-trip.

Stacks cleanly on top of the existing Namecoin NIP-05 plumbing (#1547 et al.) and reuses the same `IElectrumXClient`, pinned-cert trust store, and Tor-aware server selection.

---

## Architecture

```
                        +-----------------------+
                        | NormalizedRelayUrl    |
 user adds .bit URL --> | wss://example.bit/    |  (canonical identity, never rewritten)
                        +-----------+-----------+
                                    |
                                    v   OkHttpWebSocket.Builder
                        +-----------+-----------+
                        | urlRewriter           |  -- BitRelayUrlRewriter
                        |   (.bit -> real wss)  |     (wraps BitRelayResolver)
                        +-----------+-----------+
                                    |
                                    v
                        +-----------+-----------+
                        | clientDecorator       |  -- TlsaConnectionPolicy
                        |   (per-relay OkHttp)  |     overlays sslSocketFactory
                        +-----------+-----------+        with TlsaTrustManager
                                    |
                                    v
                        +-----------+-----------+
                        | Connect to:           |
                        |   wss://real.host/    |  or  ws://x.onion/
                        |   pinned to TLSA      |  via existing TorRelayEvaluation
                        +-----------------------+
```

Every step pulls from the **same cached `BitRelayResolver` entry** for a host:

| Step | Reads |
|---|---|
| URL rewrite | `Resolution.Resolved.resolvedUrl` |
| TLS pin     | `Resolution.Resolved.tlsaRecords` |
| Tor route   | `Resolution.Resolved.onionEndpoints` |

So a single `name_show d/<name>` ElectrumX call (cached, with positive/negative TTLs) services all three.

### New files

**`quartz/.../namecoin/`** (KMP commonMain — usable by future iOS/desktop ports too):
- `BitRelayResolver.kt` — the policy layer: parse host → ElectrumX → walk → pick first usable URL.
- `TlsaVerifier.kt` — pure-Kotlin RFC 6698 matching policy. SHA-512 injected as a callback because commonMain only ships sha256.
- `NamecoinNameResolver.kt` — gains `parseHostFlat`, `walkSubdomain`, `parseRelayUrls`, `parseTlsaRecords`, `parseTorEndpoints`, `parseIdentifierFlat`. NIP-05 identity path now also does the multi-label split + map walk.
- `README.md` — human-readable spec for the Namecoin record shapes (record format, subdomain walk, TLSA, Tor).

**`amethyst/.../service/namecoin/`**:
- `BitRelayUrlRewriter.kt` — the OkHttp-side `urlRewriter` callback. Takes a `preferOnion: (NormalizedRelayUrl) -> Boolean` for per-connect runtime Tor switching.
- `TlsaTrustManager.kt` — `X509TrustManager` wrapping the platform default. DANE-* matches accept without consulting the system trust store; PKIX-* matches additionally require the platform chain to validate. **No silent fallback to plain PKIX on a no-match** — a publisher who shipped TLSA records is opting in to mandatory TLSA validation.
- `TlsaConnectionPolicy.kt` — looks up cached TLSA records for the target host and overlays an `SSLSocketFactory` on the per-relay `OkHttpClient`. For non-`.bit` URLs, or `.bit` URLs whose record has no `tls` field, returns the input client unchanged.

### Touched files (existing)

- `OkHttpWebSocket.kt` — gains an optional `urlRewriter` (default = identity, back-compat) and a `clientDecorator` hook that runs **after** `urlRewriter` so the resolver's cache state is populated before the TLS path looks it up.
- `NamecoinNameResolver.kt` — `parseRelayUrls`/`parseTlsaRecords` now take an optional `subdomainLabels` parameter (default empty list preserves the original top-level behaviour).
- `BitRelayResolver.kt` (within Quartz) — gains `cachedTlsaFor(host)`, `cachedOnionEndpointsFor(host)`, and onion-only resolution support.
- `AppModules.kt` — wires `BitRelayResolver` into `OkHttpWebSocket.Builder` (URL rewriter + TLSA decorator + Tor preference flow), reusing the same `IElectrumXClient` and Tor-aware server selection used by NIP-05 `.bit` identity resolution.

### Reuse-first

- URL parsing → `Rfc3986.host` (existing Quartz parser).
- Identifier → Namecoin name → existing `NamecoinNameResolver.toNamecoinName` (refactored from a private helper so identity + relay paths share one parser).
- ElectrumX dispatch → same `IElectrumXClient` `AppModules` already builds (pinned certs, Tor routing).
- PKIX path validation → platform default `X509TrustManager`, never reimplemented.
- The pinned-cert ElectrumX trust store (`ElectrumXClient.PINNED_ELECTRUMX_CERTS`) is **unchanged and unaffected** — this PR only governs the relay handshake, not the ElectrumX handshake.

---

## How Namecoin records must be structured

The Namecoin `d/<name>` value is a JSON object roughly following [`namecoin/proposals` ifa-0001](https://github.com/namecoin/proposals/blob/master/ifa-0001.md). This PR reads four fields off it: `relay` / `relays` (and the `nostr.relay` / `nostr.relays` shapes), `tls`, `tor` / `_tor.txt`, and `nostr.names` (existing).

### 1. Top-level (bare host) — `wss://example.bit/`

Simplest, recommended for relay-only records:

```jsonc
{ "relay": "wss://relay.example.com/" }
```

Failover:

```jsonc
{ "relays": ["wss://relay-a.example.com/", "wss://relay-b.example.com/"] }
```

Combined identity + relay:

```jsonc
{
  "nostr": { "names": { "_": "<root-pubkey-hex>" } },
  "relay": "wss://relay.example.com/"
}
```

Both `wss://` and `ws://` accepted. Non-WebSocket schemes ignored. Resolution priority order: `relay` → `relays` → `nostr.relay` → `nostr.relays` → pubkey-keyed `nostr.relays[<pubkey>]` (last-resort, NIP-05-style).

### 2. Subdomains — `wss://relay.example.bit/`

The Namecoin `d/` namespace is **single-label** per [ifa-0001](https://github.com/namecoin/proposals/blob/master/ifa-0001.md) — there is `d/example`, but **not** `d/relay.example`. Multi-label hosts are realised through the parent record's `map` field.

For `wss://relay.example.bit/`:

1. Split: `"relay.example.bit"` → registered name `d/example`, subdomain labels `["relay"]`.
2. ElectrumX `name_show d/example` (one call — same lookup the bare host would do, so the resolver cache hits for both).
3. Walk `value.map.relay`. Exact label preferred, fall back to `value.map["*"]`, otherwise `null`. The walk also honours the [ifa-0001](https://github.com/namecoin/proposals/blob/master/ifa-0001.md) `""` empty-key default rule.
4. `relay` / `tls` / `tor` are read **from that walked node only**. They are **not inherited** from ancestors: a parent `"relay": "wss://example.bit/"` does **not** silently authorise `wss://sub.example.bit/`. Same anti-inheritance property holds for `tls` and `tor`.

Single record covering both bare and subdomain:

```jsonc
{
  "ip":    "203.0.113.10",
  "relay": "wss://example.bit/",
  "tls":   [[2,1,1,"<base64 SHA-256 of CA SPKI>"]],
  "map": {
    "relay": {
      "relay": "wss://relay.example.bit/",
      "tls":   [[2,1,1,"<same or different hash>"]]
    },
    "*": {
      "tls":   [[2,1,1,"<wildcard hash>"]]
    }
  }
}
```

This single record lets `wss://example.bit/`, `wss://relay.example.bit/`, and any `wss://*.example.bit/` route correctly with TLS pinning.

Deeper nesting walks parent-first: `a.b.c.example.bit` walks `map.c.map.b.map.a` from inside `d/example`. **Still only one ElectrumX call.**

### 3. TLS pinning — `tls` field (RFC 6698)

```jsonc
{
  "relay": "wss://relay.example.com/",
  "tls": [
    // [usage, selector, matchingType, base64-association-data]
    [3, 1, 1, "<base64 SHA-256 of leaf SubjectPublicKeyInfo>"],
    [2, 1, 1, "<base64 SHA-256 of trust-anchor SPKI>"]
  ]
}
```

Field semantics match RFC 6698 verbatim:

| Usage | Name | Behaviour |
|---|---|---|
| 0 | PKIX-TA | Trust anchor in chain + system PKIX must validate |
| 1 | PKIX-EE | End-entity cert + system PKIX must validate |
| 2 | DANE-TA | Trust anchor in chain, system PKIX bypassed |
| 3 | DANE-EE | End-entity cert, system PKIX bypassed |

| Selector | Name | Hashes |
|---|---|---|
| 0 | Cert | Full DER cert |
| 1 | SPKI | SubjectPublicKeyInfo |

| MatchingType | Name |
|---|---|
| 0 | Full (raw bytes) |
| 1 | SHA-256 |
| 2 | SHA-512 |

If a record carries `tls`, **at least one record must match** for the connection to proceed. There is no silent fallback to plain PKIX. If `tls` is absent, the relay just uses normal system trust.

### 4. Tor / `.onion` routing — `tor` and `_tor.txt`

Two equivalent shapes are accepted:

```jsonc
// Top-level "tor" — string or array of strings.
{ "tor": "x.onion" }
{ "tor": ["x.onion", "ws://y.onion/"] }

// "_tor.txt" namespace (matches the live d/testls shape).
{ "_tor": { "txt": "x.onion" } }
{ "_tor": { "txt": ["x.onion", "ws://y.onion/"] } }
```

Bare onion hostnames (`"x.onion"`) are promoted to `ws://x.onion/`. Pre-formed `ws[s]://...onion[/...]` URLs pass through verbatim. **Non-onion strings are dropped.** Multi-label `.onion` subdomains (`foo.deadbeef.onion`) are rejected so a malicious record can't publish a fake prefix.

Routing decision per connect:

| Record has `tor` / `_tor` | User Tor enabled + `onionRelaysViaTor` | Wire URL |
|---|---|---|
| no  | no  | clearnet |
| no  | yes | clearnet |
| yes | no  | clearnet (loud failure if `relay` absent) |
| yes | yes | onion → routed via existing `TorRelayEvaluation` |

The existing `.onion` Tor routing in Amethyst (`TorRelayEvaluation`) already handles `wss://x.onion/` URLs end-to-end — what was missing was the bridge from a Namecoin record's Tor alias to a wire URL OkHttp could act on. This PR adds that bridge.

A record with **only** an onion (no clearnet `relay`) is supported: with Tor on it just works; without Tor the connect fails loudly (rather than silently falling back to a non-existent clearnet endpoint). Previously such records returned `NotFound` even when Tor was on.

---

## Spec-safety properties

- **No `relay` / `tls` / `tor` inheritance from ancestors.** A subdomain only gets what it declares (or what comes through `map["*"]` and the `""` empty-key rule). Tested explicitly.
- **No silent PKIX fallback** when `tls` is published but no record matches. Mandatory-pin once you've opted in.
- **Multi-label `.onion` rejection** so a record can't say "send my Tor traffic to `evil.deadbeef.onion`".
- **Non-onion under `tor`/`_tor.txt` rejection** so a record can't redirect a Tor connection back to clearnet.
- **`relay` / `tls` walks share the same node**, so a publisher can't accidentally point clearnet traffic at one host and a TLSA hash for a different one.
- **`d/<sub>.<parent>` is never queried** as a fallback, even if `d/<parent>` is missing — would otherwise be an off-spec lookup that a malicious squatter could populate.

---

## Tests

All in Quartz `jvmTest` and Amethyst test source sets, runnable via `./gradlew :quartz:jvmTest --tests "*namecoin*"` and `./gradlew :amethyst:testPlayDebugUnitTest --tests "*Namecoin*"`.

| Suite | Cases | What it covers |
|---|---|---|
| `BitRelayRecordParserTest` | 13 | All relay record shapes, dedup, non-ws rejection, malformed JSON, `toNamecoinName` mapping |
| `BitRelayResolverTest` | 11 + 5 + 5 + 4 + 5 = **30** | End-to-end resolution, NotFound/Error paths, cache hit, path preservation; subdomain walk; subdomain-trap regression; TLSA passthrough; onion-from-`_tor.txt` |
| `BitRelayUrlRewriterTest` | 7 | Clearnet default, onion when `preferOnion=true`, no-onion fallback, onion-only record routing, non-bit passthrough, runtime toggle re-evaluation |
| `NamecoinSubdomainTest` | 27 + 11 = **38** | All host/label combinations, walker semantics (exact / wildcard / empty-key), shorthand promotion, anti-inheritance, live `d/testls` JSON shape, `tor` / `_tor.txt` shapes |
| `NamecoinSubdomainIdentityTest` | **16** | NIP-05 identity multi-label split, end-to-end `alice@relay.testls.bit`, anti-inheritance for `nostr.names`, depth, single-call guarantee |
| `TlsaVerifierTest` | **19** | Every `selector × matchingType` combo, EE-vs-TA leaf positioning, multi-record short-circuit, base64 padding/whitespace/url-safe quirks, garbage-data robustness |
| `TlsaTrustManagerTest` | **7** | DANE-* accept without system trust, PKIX-* requires both, no-match no-fallback, cert-chain edge cases |

**~135 new test cases. 158/158 namecoin-package + Amethyst-side namecoin tests passing.**

Live-verified twice against the real Namecoin blockchain via `d/testls`:
- `wss://relay.testls.bit/` → resolves through `map.relay` to the published wss endpoint.
- `wss://testls.bit/` → resolves to the top-level `relay` field.
- `alice@testls.bit` (and `m@testls.bit`) → existing identity path unchanged, still resolves correctly.

---

## Back-compat

- `urlRewriter` defaults to the identity function on `OkHttpWebSocket.Builder`. Existing call sites are unaffected.
- `parseRelayUrls` / `parseTlsaRecords` `subdomainLabels` parameter defaults to `emptyList()` — original top-level behaviour preserved for the existing NIP-05 path.
- All 15 existing `NamecoinNameResolverTest` cases still pass without modification.
- Literal `d/...` and `id/...` inputs are **not** split: if a caller hands `d/relay.testls` directly, that's what gets queried. Multi-label split only triggers on `host.bit` and `<localPart>@host.bit` shapes.
- Bare-host inputs (`alice@testls.bit`, `testls.bit`) take the empty-labels path through `walkSubdomain`, which returns the root object unchanged.
- `ElectrumXClient.PINNED_ELECTRUMX_CERTS` and the Tor-aware ElectrumX server selection are untouched.
- One drive-by fix: `parseIdentifier` now lowercases input before `removeSuffix(".bit")`, so `EXAMPLE.BIT` maps to `d/example` instead of the previously-broken `d/example.bit`.

---

## Commits

| | Title |
|---|---|
| `183ef0a7e` | feat(relay): resolve Namecoin .bit hostnames to real wss endpoints |
| `32cc7b33f` | feat(relay): pin .bit relay TLS handshake against Namecoin TLSA records |
| `1bcdbde2f` | feat(relay): walk Namecoin map tree for multi-label .bit hosts |
| `92ac398b7` | test(relay): pin that .bit subdomains query only the parent name, never `d/<sub>.<parent>` |
| `800802da1` | feat(nip05): walk Namecoin map tree for multi-label NIP-05 identifiers |
| `8c1a49efd` | feat(relay): resolve and route .bit relays to .onion endpoints over Tor |

Each commit is independently reviewable and self-tested. Happy to split into separate PRs (rewrite + TLSA + map-walk + Tor) if that's easier — they layer cleanly.

---

## Docs

- `quartz/.../nip05DnsIdentifiers/namecoin/README.md` — human-readable spec for record shapes, subdomain walk, TLSA pinning, Tor routing, and the security properties.
- `relaysetup.txt` (existing operator guide) — gains parts B/M walking through publishing `nostr.names`, `relay`, `tls`, and `_tor.txt` end-to-end against a live Namecoin node.

---

## Builds on / depends on

Existing Namecoin NIP-05 plumbing already in `vitorpamplona/amethyst`:
- `IElectrumXClient` + pinned-cert trust store
- `NamecoinNameResolver` (NIP-05 identity)
- Tor-aware ElectrumX server selection

This PR is the second half: **identity + relay + TLS pin + Tor routing all from the same blockchain record, with one ElectrumX call per host.**


vitorpamplona:
Why are you building this Rewriter externally instead of using OkHttp interceptors for it? This is exactly why Interceptors exist in OkHttp

mstrofnone:
Good question @vitorpamplona  — I went back and forth on this exact point. Three reasons interceptors don't fit here:

**1. OkHttp interceptors don't run for `newWebSocket()`.**
OkHttp's `RealWebSocket` builds its own `Call` and runs the WebSocket handshake outside the interceptor chain that `OkHttpClient.newCall()` exposes. App and network interceptors registered on the `OkHttpClient` are not invoked for the WS upgrade request — they only fire for `Call`-based HTTP. So an `Interceptor` can't see, let alone rewrite, a `wss://example.bit/` handshake. (You can verify this against `RealWebSocket.connect()` in OkHttp 4.x — it constructs a stripped client `clientForWebSocket = client.newBuilder()...build()` and dispatches via `RealCall` without the user interceptor chain having a chance to redirect the URL.)

**2. Even if interceptors ran, they'd be too late for the TLS pin.**
TLSA validation has to run inside the TLS handshake, which happens *before* any application-level interceptor sees a request. The pin is enforced through a custom `X509TrustManager` (`TlsaTrustManager`) bolted onto the `SSLSocketFactory` of a per-relay `OkHttpClient`. That's a client-construction-time concern, not a request-time one. Interceptors operate on `Request`/`Response`, not on socket factories or trust managers — by the time an interceptor could observe anything, the cert chain has already been accepted or rejected.

**3. The two pieces have to share state across that boundary.**
The URL rewrite (`.bit` → real `wss://...`) and the TLS pin both read from the same cached `BitRelayResolver` entry — one ElectrumX `name_show` call services both. They have to agree on which Namecoin record they're acting on, *before* the socket connects. The current design lets `urlRewriter` populate the resolver cache, then `clientDecorator` reads from that same cache to overlay the right `SSLSocketFactory` on the per-relay client. An interceptor sitting between those two would have nowhere useful to live.

Hence the two `OkHttpWebSocket.Builder` hooks:

- `urlRewriter: (NormalizedRelayUrl) -> String` — fires once at `connect()`, before `Request.Builder().url(...)`.
- `clientDecorator: (NormalizedRelayUrl, OkHttpClient) -> OkHttpClient` — overlays per-relay TLS state on the client returned by `httpClient(url)`.

Both default to identity, so existing call sites are unaffected.

If you'd prefer a different shape — e.g. fold the rewriter into a `RequestFactory` or expose a single `WebSocketRequestPolicy` interface that returns `(URL, OkHttpClient)` together — happy to refactor. But OkHttp's `Interceptor` API specifically can't carry this, because of (1) and (2) above.

vitorpamplona:
That sucks... 

Is there any live relay with a .bit domain?

vitorpamplona:
Also, update to resolve the conflicts

mstrofnone:
> Is there any live relay with a .bit domain?

working on this now

mstrofnone:
relay.testls.bit is up and running. Will do some testing.

mstrofnone:
Correction on my earlier reply — point **1** is wrong for OkHttp 5.x (which is what we're on: `okhttp = "5.3.2"` in `gradle/libs.versions.toml`).

Verified against `okhttp@parent-5.3.2` `RealCall.getResponseWithInterceptorChain()`:

```kotlin
interceptors += client.interceptors                  // ← user app interceptors DO run
interceptors += RetryAndFollowUpInterceptor(client)
interceptors += BridgeInterceptor(client.cookieJar)
interceptors += CacheInterceptor(client.cache)
interceptors += ConnectInterceptor
if (!forWebSocket) {
  interceptors += client.networkInterceptors         // ← only NETWORK interceptors are skipped
}
interceptors += CallServerInterceptor
```

Application interceptors DO see the WebSocket upgrade request, so the URL rewrite alone *could* live in an `Interceptor`. Only network interceptors are skipped (sensible — once the upgrade hijacks the socket, post-connect framing interceptors would corrupt the WS frames).

Points **2** and **3** still hold and are the load-bearing reasons:

- **TLS pin runs inside the handshake**, before any `Interceptor.intercept(chain)` is reached. The pin lives on the `OkHttpClient`'s `SSLSocketFactory` / `X509TrustManager`, both baked in at `OkHttpClient.Builder.build()` time. An interceptor cannot swap them mid-chain.
- **URL rewrite and TLS pin share the same `BitRelayResolver` cache entry** (one `name_show` call services both), and the pin has to be installed on the per-relay client *before* `RealWebSocket.connect()` runs.

So the URL rewrite could be moved into an app interceptor, but the TLS pin can't — and splitting the two across the client-construction / interceptor boundary loses the shared record. Hence both hooks live together on `OkHttpWebSocket.Builder`.

Re: upstream fix — square/okhttp has closed this category of request. [#4192 "Interceptor access to WebSocket Connection"](https://github.com/square/okhttp/issues/4192) sat open ~5 years before @yschimke closed it: "Closing as low priority. Not implemented after almost 5 years." [#5580](https://github.com/square/okhttp/issues/5580) is in the Icebox milestone. A surgical `sslSocketFactorySelector(call)` hook is the only upstream change that would actually subsume `clientDecorator`, but realistic land-and-ship timeline isn't compatible with this PR.

mstrofnone:
## On-device smoke test on current `main` — summary

Rebased the 6 commits of this PR onto current `vitorpamplona/amethyst:main` (which has moved ~1198 commits since the branch base). Result available on `mstrofnone:test/2595-on-current-main` for reviewers who want to verify the rebase shape without pulling the import work in #2612.

### Rebase outcome
- **Final commit on the rebase branch**: `3a02889b1`
- **Conflicts**: 2 trivial — both in `amethyst/src/main/java/com/vitorpamplona/amethyst/AppModules.kt`, just import-zone collisions with main's new `AlwaysOnNotificationServiceManager` / `NotificationDispatcher` imports. Resolved by keeping all three in alphabetical order.
- **One follow-up fix needed**: upstream commit `1d9c1eb9b` moved `TorType` from `amethyst.ui.tor` to `commons.tor`. `BitRelayUrlRewriter`'s `preferOnion` callback at `AppModules.kt:318` needed updating:
  ```diff
  -                        settings.torType != com.vitorpamplona.amethyst.ui.tor.TorType.OFF &&
  +                        settings.torType != com.vitorpamplona.amethyst.commons.tor.TorType.OFF &&
  ```
  Folded into the last commit via `git rebase --autosquash`.

### Build + tests
- `:quartz:jvmTest --tests "com.vitorpamplona.quartz.nip05.namecoin.*"` → **127/127 green**
- `assemblePlayDebug` → success (universal 184 MB, arm64-v8a 88 MB)
- Spotless clean

### On-device smoke test (AVD `amethyst_api36`)

Drove the UI to log in and add `wss://relay.testls.bit/` (relay deployed at 23.158.233.10, TLS-pinned via Namecoin record at `d/testls`, strict NIP-05 gating: `NAMECOIN_POLICY_MODE=all-kinds-require-bit`).

**3 issues found, none of them regressions, all reachable via #2612:**

1. **Clearnet path can't resolve `.bit` hosts** (CRITICAL) — when the Namecoin record's `relay` field is itself a `.bit` URL (the natural shape, e.g. `wss://relay.testls.bit/`), the rewriter resolves to that same `.bit` and OkHttp falls back to ICANN DNS, which has nothing. Connection dies with `UnknownHostException` even though the record published `ip: "23.158.233.10"` at the same walked node. Affects every `.bit` relay whose publisher uses `.bit` URLs end-to-end.

2. **NIP-11 RelayInfo HTTP fetch ignores the `.bit` resolver entirely** — even when WS connect would work (e.g. via Tor), `Nip11Retriever` constructs an `OkHttpClient` with no `.bit` awareness, so `https://relay.testls.bit/` `UnknownHostException`s and surfaces as `FAIL_TO_ASSEMBLE_URL`.

3. **TLSA validation rejects the live `relay.testls.bit` chain because `TlsaVerifier` only walks the served chain and doesn't understand ncdns/ncgencert's "AIA-stapled `pubb64`" convention** — when ncgencert is run in its default mode (which is what `relaysetup.txt` §C.4 prescribes), the TLSA hash is over the *AIA Parent CA*'s SPKI, but the relay deliberately doesn't serve the AIA Parent in the chain; instead the parent's SPKI is stapled inline in the served Domain CA's issuer-DN `serialNumber` field as `Stapled: {"pubb64":"..."}`. The current verifier walks only the chain it was given, finds no match, and returns `Result.NoMatch`.

The first two are fixed in #2612 (already open). The third is also reproducible against any ncgencert-style deployment and warrants a small follow-up to `TlsaVerifier` in `commonMain`. I'm planning to land that as a 5th commit on #2612 unless you'd prefer it as a separate PR — happy either way.

### Strict-mode rejection from the relay (verified via direct websocket from laptop)

Independent confirmation that `strfry-namecoin-policy` is producing the right NIP-20 OK frames:

| Trigger | NIP-20 reason |
|---|---|
| Any non-kind:0 from unknown pubkey | `blocked: this relay requires a verified Namecoin .bit NIP-05 identity (publish a kind:0 first)` |
| kind:0 with NIP-05 that doesn't resolve to a known local-part | `invalid: Namecoin NIP-05 "<id>" could not be resolved (name missing, expired, or malformed)` |
| kind:0 with NIP-05 that resolves to a different pubkey | `invalid: Namecoin NIP-05 "<id>" maps to <hex>… but event.pubkey is <other>…` |

So the server-side strict-mode behavior is already validated end-to-end. Once #2612 + the TlsaVerifier patch land, the on-device side will be too.

mstrofnone:
relay.testls.bit is also available by browser for those that know 🎪

mstrofnone:
Hey @vitorpamplona any concerns with the code? I just published an updated event here: https://njump.me/nevent1qgsyxxz7mm9kwkyjsf935da90ulyqlaautk6wgq68q5m3n6t5lzmfuqpzamhxue69uhhyetvv9ujuar9wd6xcuewvf5hgtcpzpmhxue69uhkummnw3ezuamfdejsqgrjc0qanktgq8elp49uqurhac3es4t6205kzq56qqznpz7rje5z5uxuyxyj

that shows how to use my relay. Anybody can read from it but I am resource-constrained(like all others) so a little p.o.w is needed to publish to it.

vitorpamplona:
The main issue with this is that the relay setting will only work on Amethyst. Nobody else will be able to see your posts or send replies, zaps, etc to you.. which leads to a broken user experience (change clients and a lot of things are missing) between nostr clients. 

Are you planning to submit support for namecoin as a NIP so that every other client must implement too?

mstrofnone:
Hey @vitorpamplona, my understanding of the nostr ecosystem may be flawed so bear with me. I originally tried a NIP and my GitHub account got nuked. Currently the only thing that has changed is your merges so I believe another client needs this implemented for a nip to become relevant again? Is that the case? No other clients have merged anything but I have outstanding PRs for nip-05 resolution for quite a few. If any clients merge that I will happily submit the relay code to them as well and create a NIP. 

For the NIP I really want input from the Namecoin devs in any case and they are busy ppl doing busy ppl things saving the internet. Edited (and added for effect): Said another way, they probably don't have bandwidth to nostr cheerlead atm but the chain is there. For other clients short-term "If you don't believe me or don't get it, I don't have time to try to convince you, sorry." https://bitcointalk.org/index.php?topic=1790.msg27484#msg27484

vitorpamplona:
Well, the NIP is just one of the ways of doing. The main problem is the user experience. Nostr users like (and want to) switch clients all the time. So, when they switch, they need to see most of the things they see in one on the other as well. In the case of this PR, there is little incentive for people to use a namecoin-based relay if nobody can see their posts. It's a real chicken-and-egg problem. 

Our little experiment with the NIP-05 extensions already showed some issues. Clients who do not support .bit domains will show a warning sign that the NIP-05 address is not working for that person. Which means that in those clients, a user with .bit address will look like a scamming/impersonating account. Which turns into anyone using namecoin a NIP-05 is a scammer. Which is the opposite of what you want. 

Namecoin is not the only one with this issue. People don't use Tor (.onion) relays for the same reason: only clients that implement Tor (which right now is only Amethyst) can see their posts, interact with those posts or DM onion relays. So, in the end, nobody uses them. 

We need to figure out how to get all clients using both schemes. Otherwise, they will never flourish.

mstrofnone:
@vitorpamplona if other clients aren't even using tor what chance does little old namecoin have 😆 also an iOS app seems even more appealing given there is no sane nostr client there

vitorpamplona:
Yeah, I know.. that's why I am asking.. you gotta have a plan. :)

mstrofnone:
> The main issue with this is that the relay setting will only work on Amethyst. Nobody else will be able to see your posts or send replies, zaps, etc to you.. which leads to a broken user experience (change clients and a lot of things are missing) between nostr clients.
> 
> Are you planning to submit support for namecoin as a NIP so that every other client must implement too?

https://github.com/mstrofnone/nips/blob/master/README.md

there is lots going on in https://github.com/nostr-protocol/nips should I open an issue/submit PRs there?

edit: opened this one:

https://github.com/nostr-protocol/nips/issues/2330

Issue #2446, opened 2026-04-19
[BUG] curser moves randomly while typing in V1.08 on Android


yojimble:
**Describe the bug**
keyboard behaves strangely if inserting a tagged user, the curser will move itself into the middle of previous text whilst typing

**To Reproduce**
Steps to reproduce the behaviour:
1. write a note
2. type '@' and tap the user
3. continue typing
4. at some point the curser jumps backwards, usually inside of the npub which breaks the tag

**Video and Screenshots**

https://github.com/user-attachments/assets/04ed7ff0-b048-4715-9bd2-b49ec0cec207
https://github.com/user-attachments/assets/ec4ada56-37b5-4462-b226-afe5c531d3e4

**Device (please complete the following information):**
 - Phone Brand/Model [Pixel 5]:
 - Android Version [Lineage 22.1/ Android 15]:
 - App Version [e.g. v1.08]:
 - App Flavour [e.g. Zap Store]: 
 - Amber Version [6.01]:

davotoula:
What keyboard are you using? This has been resolved for my Microsoft SwitKey

yojimble:
Heliboard

vitorpamplona:
The fix for the random cursor jumps has been merged and shipped in the releases after v1.08 (latest is v1.12.6). Closing — please reopen if you still see it on the current version.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2597, opened 2026-04-27
Fix chat cursor jump

davotoula:
## Summary                                                           

Fixes the cursor jumping into the middle of an underlying npub when the IME edits a wedge'd `@DisplayName`, and prevents the IME from corrupting the npub itself — particularly multi-word display names like `@Vitor Pamplona` that SwiftKey re-composes word-by-word.
                                                                                                                                                                                                                                
  Two related changes:   
                                      
  - **`UrlUserTagTransformation`**: replace the proportional cursor-mapping inside a mention wedge with a snap-to-end. Cursors strictly inside the bech32 land at the trailing edge of the displayed name; this stops the IME   
  from placing the caret in the middle of the npub on backspace.
  - **NEW `MentionPreservingInputTransformation`**: a singleton `InputTransformation` that runs on every IME change. When a change touches a complete mention (`@npub1…`, `nostr:npub1…`, `@nprofile1…`, `nostr:nprofile1…`)    
  with anything other than a clean empty-replacement delete that fully covers it, the IME edit is reverted and the entire mention plus any immediately-trailing whitespace is removed atomically. Wired into all 7 text fields  
  that use the wedge: `MessageField`, `EditFieldRow`, `PrivateMessageEditFieldRow`, `NewGroupDMScreen`, `NewPublicMessageScreen`, `LongFormPostScreen`, `SellProduct`.
                                                                                                                                                                                                                                
  The mention regex is hoisted to a shared package-level `MENTION_REGEX` so the output wedge and the input guard agree on what counts as a mention. A cheap `contains("npub1") || contains("nprofile1")` gate makes the typical 
  mention-free keystroke return before any regex work.                                                                                                                                                                        
                                                                                                                                                                                                                                
  The predicate handles four cases:                                                                                                                                                                                             
                                          
  | Change shape | Behavior |                                                                                                                                                                                                   
  |---|---|                                                                                                                                                                                                                   
  | Pure delete fully covering a mention | passes through (user removed the chip) |
  | Partial overlap | reverts + atomic-collapse |                                                                                                                                                                               
  | Scope-exact replace with non-empty text | reverts + atomic-collapse (SwiftKey case) |                                                                                                                                       
  | Scope-broader replace (covers mention + surrounding text) | passes through (deliberate broad edit) |                                                                                                                        
                                                                                                                                                                                                                                
  ## Test plan                                                                                                                                                                                                                  
                                                                                                                                                                                                                                
  - [x] 11 instrumented tests in `MentionPreservingInputTransformationTest` covering the predicate matrix — all pass on Pixel 9a / API 36.                                                                                      
  - [x] `UrlUserTagTransformationTest` updated for the new snap-to-end semantics; cursor positions strictly inside the npub map to the trailing edge of the displayed name.                                                   
  - [x] Manual verification on Pixel 9a with a `@Vitor Pamplona` mention: backspace + word-delete cleanly remove the whole chip, no orphan bech32 left in the buffer.                                                           
  - [x] SwiftKey-specific manual repro on a different device.

### DIFF
--- a/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/MentionPreservingInputTransformationTest.kt
+++ b/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/MentionPreservingInputTransformationTest.kt
@@ -1,0 +1,145 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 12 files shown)


davotoula:
fix for #2446

PR #2612, opened 2026-04-27
feat(namecoin): resolve `import` items per ifa-0001

mstrofnone:
## Summary

Adds Namecoin [`import`](https://github.com/namecoin/proposals/blob/master/ifa-0001.md) item support to Quartz's `.bit` resolution paths. With this commit, a Domain Name Object can pull values from another Namecoin name and have them merged in transparently before any subdomain walking happens.

This was missing from the existing `feat/bit-relay-resolution` work — `BitRelayResolver` and `NamecoinNameResolver` ignored `import` entirely, which made the 520-byte per-name limit a hard ceiling on what a single record could express. With imports, an operator can keep the apex record under budget by splitting shared boilerplate (a stable `nostr.names` block, a common TLSA pin, a `map.relay.*` blob) into a sibling name.

## Why now

Concrete use case: deploying `wss://relay.testls.bit/`. A single `d/testls` value that needs `ip`, `map.relay.{relay,tls,_tor}`, top-level `_tor`, and three `nostr.names` entries hits ~570 bytes — over the 520-byte chain limit. The user has to drop something to publish. With `import`, you publish:

```jsonc
// d/testls (apex, small)
{
  "import": "dd/testls-relay",
  "ip": "107.152.38.155",
  "_tor": { "txt": "<old onion>" }
}

// dd/testls-relay (boilerplate, stable)
{
  "map": {
    "relay": {
      "ip": "23.158.233.10",
      "relay": "wss://relay.testls.bit/",
      "tls":   [[2,1,1,"<HASH>"]],
      "_tor":  { "txt": "<new onion>" }
    }
  },
  "nostr": { "names": { "_": "<hex>", "m": "<hex>", "seed": "<hex>" } }
}
```

Both stay well under the limit. Amethyst, with this PR, walks both names automatically and produces the same effective view as the single-record version.

## ifa-0001 §"import" — what's implemented

Per [the spec](https://github.com/namecoin/proposals/blob/master/ifa-0001.md):

- ✅ **Importer items take precedence** over imported items.
- ✅ **`null` items in the importer are still "present" for precedence** and so suppress the corresponding imported value (semantic suppression).
- ✅ **Recursion supported up to depth 4** (the spec minimum). Deeper chains are silently truncated with the importing object's items still applying.
- ✅ **Cycles are broken** by a per-chain visited-set; tests pin that `d/a → d/b → d/a` terminates.
- ✅ **Subdomain selectors** (the optional 2nd element of each inner array) descend through `map` using the same walker the rest of the codepath uses, so `*` wildcard and `""` empty-key default behave identically.
- ✅ **Canonical form**: `[ ["d/foo"], ["d/bar","selector"] ]`.
- ✅ **Three short-hand forms** widely used in real records:
   - `"import": "d/foo"`
   - `"import": ["d/foo"]`
   - `"import": ["d/foo", "selector"]`
- ✅ **Multiple imports in one array** — later wins over earlier (spec is silent on this; chosen for least surprise so listing two libraries doesn't silently ignore the second).

### Lenient failure semantics (deliberate deviation from strict spec)

The spec says a failed import lookup MAY cause the whole importing record to fail. This implementation treats a failed import as `{}` instead, so a transient ElectrumX hiccup on an imported name does not poison the importing name's resolution. The importing object's own items still apply.

Rationale: the rest of Quartz's `.bit` paths already follow a "best-effort, never throw" convention (`parseRelayUrls` etc. all return empty lists on parse errors). Hard-failing on import lookup would make resolution flakier than the existing behaviour without imports — the opposite of what users want. Documented in the source.

## Architecture

New file: `NamecoinImportResolver.kt` (commonMain). Pure suspend object with one entry point:

```kotlin
suspend fun expandImports(
    root: JsonObject,
    maxDepth: Int = DEFAULT_MAX_DEPTH,
    fetcher: NameValueFetcher,
): JsonObject
```

The fetcher is a typealias for `suspend (String) -> String?` — supplied by the caller (a small new `NamecoinNameResolver.fetchValueForImport` helper that absorbs lookup outcomes into `null`).

Both async paths are wired:

- **`BitRelayResolver.resolveRaw`** — parses the parent record once, expands imports, then hands the merged `JsonObject` to the existing `parseRelayUrls` / `parseTlsaRecords` / `parseTorEndpoints`. New `JsonObject` overloads of those companion methods are added so the value JSON is parsed once across the whole flow.
- **`NamecoinNameResolver.performLookupDetailed`** — same treatment for the NIP-05 identity path. Both flows share the exact same I/O semantics (timeouts, exception translation) because they share `lookupNameDetailed`.

Records that don't use `import` pay zero additional cost — `expandImports` is a no-op when no `import` key is present.

## Tests

18 new tests in `NamecoinImportTest`:

**Pure unit tests** (no I/O, in-memory fetcher):
- no-import passthrough
- string shorthand
- canonical array-of-arrays, processed in order, later wins
- pair-array shorthand with subdomain selector
- importer precedence (importer keys win, fill in only missing keys)
- `null` in importer suppresses imported value
- recursion depth 4 happy path
- over-budget recursion silently truncated
- failed import lookup is treated as `{}`
- cycle protection (`d/a → d/b → d/a` terminates)
- malformed `import` value skipped without throwing
- import target with malformed JSON skipped
- multi-label DNS selector descends via `walkSubdomain`

**Integration tests** (in-memory `IElectrumXClient` double):
- `BitRelayResolver` follows import on parent record (resolves relay+TLSA from sibling)
- `BitRelayResolver` importer-wins on `map` (importer's `map.relay.relay` overrides imported one)
- `BitRelayResolver` still resolves when imported name is missing
- NIP-05 lookup follows import for shared `nostr.names` block
- NIP-05 lookup with subdomain follows import to find names at the walked node

All 145 existing namecoin tests in the package continue to pass.

## Files

- New: `quartz/src/commonMain/.../namecoin/NamecoinImportResolver.kt`
- New: `quartz/src/jvmTest/.../namecoin/NamecoinImportTest.kt` (18 tests)
- Modified: `NamecoinNameResolver.kt` — adds `fetchValueForImport` helper, three `JsonObject`-accepting parser overloads, and import expansion in `performLookupDetailed`.
- Modified: `BitRelayResolver.kt` — parses value once, expands imports, hands `JsonObject` to overloaded parsers.
- Modified: `README.md` — adds `NamecoinImportResolver.kt` to the components table.

## Test note

The current `feat/bit-relay-resolution` branch tip has unrelated KMP/MLS test compile failures (`synchronized` not in commonMain, missing `confirmationTag` args) that prevent `:quartz:jvmTest` from running the full suite. Per the established workflow, namecoin tests were exercised in isolation:

```
./gradlew :quartz:jvmTest --tests "com.vitorpamplona.quartz.nip05.namecoin.*"
```

with the `marmot` test directories temporarily moved aside. Result: **145/145 namecoin tests pass** (127 pre-existing + 18 new). Spotless is clean.

## Closes / related

Builds on top of #2595 (`feat/bit-relay-resolution`). Targets the same set of users. Together, these mean the `.bit` Nostr relay deployment story (TLSA-pinned wss + onion routing + NIP-05 identity + spec-conformant import) is fully covered for any record shape that fits the chain's 520-byte limit, with imports as the spec-blessed escape valve when it doesn't.


mstrofnone:
## On-device smoke test of `bf1af40fc` — both targeted bugs verified fixed; one new gap surfaced

Ran the latest commit on AVD `amethyst_api36`. Captured logcat under `/Users/m/.openclaw/workspace-demo/screens-2612/`.

### ✅ Bug 1 — `.bit` clearnet DNS resolution (the new `BitRelayDns`)

```
12:37:37.513  BitRelayUrlRewriter: .bit relay wss://relay.testls.bit/ -> wss://relay.testls.bit/ (1 clearnet candidate(s), 1 onion endpoint(s))
12:37:37.527  BitRelayDns: relay.testls.bit -> 23.158.233.10 (from Namecoin)        ← NEW
12:37:38.025  Relay relay.testls.bit: WebSocket Failure: TLSA validation failed for relay.testls.bit: chain matched none of 1 record(s) from the Namecoin `tls` field
```

The new `BitRelayDns` log line is firing as designed. OkHttp gets the IP from the resolver cache, the TCP connection opens, and the TLS handshake starts. The previous `UnknownHostException` failure is gone. **Bug 1 fully fixed.**

### ✅ Bug 2 — NIP-11 RelayInfo (the wrapped `Nip11CachedRetriever`)

The HTTP path now also runs through `BitRelayDns` + `TlsaTrustManager`. Stack trace from the (TLSA-related) failure confirms it:

```
RelayInfoFail: javax.net.ssl.SSLHandshakeException: TLSA validation failed for relay.testls.bit: chain matched none of 1 record(s) from the Namecoin `tls` field
RelayInfoFail:     at com.android.org.conscrypt.SSLUtils.toSSLHandshakeException(...)
RelayInfoFail:     ...
RelayInfoFail:     at okhttp3.internal.connection.ConnectPlan.connectTls(...)
RelayInfoFail: Caused by: java.security.cert.CertificateException: TLSA validation failed ...
RelayInfoFail:     at com.vitorpamplona.amethyst.service.namecoin.TlsaTrustManager.checkServerTrusted(TlsaTrustManager.kt:132)
```

Before this commit: `RelayInfoFail: ... FAIL_TO_ASSEMBLE_URL` (HTTP path never reached the TLS layer because OkHttp couldn't resolve the host). Now the WebSocket path and the HTTP path converge on the same single point of failure (Bug 3 below) — exactly the right shape. **Bug 2 fully fixed.**

### ✅ Tor onion routing

```
12:41:48.149  BitRelayUrlRewriter: .bit relay wss://relay.testls.bit/ -> ws://6cbn4rsk...onion/ (Tor onion)
12:41:48.149  OkHttpWebSocket: Rewriting connect URL wss://relay.testls.bit/ -> ws://6cbn4rsk...onion/
```

URL rewrite is correct. Connection then fails because the bare AVD has no Orbot bound (`Unable to resolve host "...onion"`) — that's an environmental issue, not a code regression.

### ❌ Bug 3 — `TlsaVerifier` doesn't understand ncgencert's AIA-stapled SPKI

Pre-existing in `quartz/.../TlsaVerifier.kt` (untouched by this commit). The new DNS fix has now exposed it.

The Namecoin record at `d/testls.map.relay.tls`:
```
[[2, 1, 1, "m14YT5aSELhdDbZXOKFcSj2kK59XzV5lkiUlElBZh4A="]]
```
i.e. `[DANE-TA, SubjectPublicKeyInfo, SHA-256, <base64>]`.

The relay's served chain has 2 certs:

| pos | subject | SHA-256(SPKI) base64 |
|---|---|---|
| leaf | `CN=relay.testls.bit` | `vfk1Y/ceDTk2dln5uhhaC7l6a57/LB5yga2jRQ87INg=` |
| intermediate | `CN=relay.testls.bit Domain CA` | `J2zlBI75T6TkJ9wIijA2maMBtgEhiFZ8MOGF7KLQ/4U=` |

Neither matches the TLSA pin. `TlsaVerifier.verify` walks the chain, finds no match, and correctly returns `Result.NoMatch`.

But the intermediate's **issuer DN** carries:
```
CN=relay.testls.bit Domain AIA Parent CA
serialNumber="Namecoin TLS Certificate

Stapled: {"pubb64":"MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEeY7CiZCmv2LS0_GQe0SbIz94w7mRDYO402v1W_wGDNWjLXCrT0qmoccMCIgYXGZ3aM6UiOMPssU3jYuy7QgfKA"}"
```

Decoding that `pubb64` (URL-safe base64 of a 91-byte DER SubjectPublicKeyInfo) and SHA-256 hashing it yields **exactly** `m14YT5aSELhdDbZXOKFcSj2kK59XzV5lkiUlElBZh4A=` — the published TLSA hash.

So the pin matches the SPKI of the AIA Parent CA, which the relay deliberately doesn't serve in the chain. Per the ncdns/ncgencert convention (which `relaysetup.txt` §C in this very PR series prescribes as the canonical workflow), the parent CA's SPKI is "stapled" inline so verifiers can synthetically construct a one-cert-higher chain without an extra round trip. This is what makes leaf + Domain CA rotation cheap (you don't need a new TLSA `name_update` for either).

The current `TlsaVerifier` doesn't know about this convention.

### Plan: fold the AIA-stapled SPKI fix into this PR as a 5th commit

Sketch:
1. After walking the served chain with no match, scan each cert's issuer DN for the `Stapled: {"pubb64":"..."}` pattern (and/or the equivalent AIA URL parameter)
2. URL-safe base64 decode the value (DER `SubjectPublicKeyInfo`)
3. Compute SHA-256, compare against TLSA records (DANE-TA, selector=SPKI, matchingType=SHA-256)
4. If match → accept. Cryptographically equivalent to having that cert in the chain — the issuer DN itself is signed by the Domain CA's owner, so an attacker can't forge a stapled `pubb64` without breaking the Domain CA's signature.

~50 LOC in `commonMain`, lives next to the existing `TlsaVerifier` matching code. New tests using the chain we just observed as a fixture. Will keep all the existing TLSA semantics intact (stapled-pubb64 is purely additive: if the served chain itself matches, that path still wins).

Will push as the 5th commit on this branch. Happy to split it into a separate PR if reviewers prefer to keep the import work and the AIA-staple work isolated — say the word.

### Independent confirmation: relay-side strict-mode behavior

Verified separately via direct WebSocket from laptop (using `nak` + Python with `--insecure` to bypass macOS trust store, since our DANE-TA CA isn't system-trusted):

| Trigger | NIP-20 reason |
|---|---|
| Any non-kind:0 from unknown pubkey | `blocked: this relay requires a verified Namecoin .bit NIP-05 identity (publish a kind:0 first)` |
| kind:0 with NIP-05 that doesn't resolve | `invalid: Namecoin NIP-05 "<id>" could not be resolved (name missing, expired, or malformed)` |
| kind:0 with NIP-05 that maps to a different pubkey | `invalid: Namecoin NIP-05 "<id>" maps to <hex>… but event.pubkey is <other>…` |

So the strfry-namecoin-policy plugin's reject path is well-formed and NIP-20 compliant. Once the TlsaVerifier patch lands and the on-device handshake succeeds, the same rejections will be visible from the Amethyst UI.

mstrofnone:
## Bugs 3 + 4 fixed, end-to-end smoke test passing ✅

Two new commits on top of bf1af40:

- **1938ec2** `feat(relay): accept ncgencert AIA-stapled SPKI as TLSA trust anchor` — fixes bug 3 (TlsaVerifier didn't understand ncgencert's AIA-stapled SPKI convention; the served chain doesn't include the AIA Parent CA but the TLSA hash is over its SPKI). New `AiaStapledPubkeyExtractor` parses the staple from either the Domain CA's issuer-DN `serialNumber` attribute or the AIA `caIssuers` URL `pubb64=` query parameter. Stapled SPKIs participate only in `*-TA` usages with the SPKI selector. 24 new tests; 207/207 namecoin tests pass.

- **bfeb941** `fix(relay): force Tor route for .bit relays when onionRelaysViaTor on` — fixes bug 4 (when the URL rewriter chose an onion endpoint, the OkHttp client was selected from the original `wss://...bit/` URL and bypassed the SOCKS proxy, so OkHttp asked the system DNS resolver for the `.onion` and failed with `Unable to resolve host`). The `AppModules` httpClient lambda now mirrors `BitRelayUrlRewriter.preferOnion` so the URL rewrite and the client selection stay in sync.

### Live smoke test on emulator-5554 (sdk_gphone64_arm64, API 36, Tor=OFF)

```
D BitRelayUrlRewriter: .bit relay wss://relay.testls.bit/ -> wss://relay.testls.bit/ (1 clearnet candidate(s), 1 onion endpoint(s))
D BitRelayDns: relay.testls.bit -> 23.158.233.10 (from Namecoin)
D TlsaTrustManager: TLSA match for relay.testls.bit: usage=DANE_TA selector=SUBJECT_PUBLIC_KEY_INFO matchingType=SHA_256 chainIndex=-1 (DANE-only)
D Relay relay.testls.bit: OnOpen (ping: 213ms, using compression)
```

Note the `chainIndex=-1` — the AIA-stapled fallback fired (the served leaf + Domain CA SPKIs don't match the published TLSA hash; the AIA Parent CA's SPKI does, and we recover it from the staple).

Server-side log (strfry on relay.testls.bit) confirmed both directions of traffic:

```
[8] Connect from 127.0.0.1 compression=Y sliding=Y
[8] write policy blocked event d3bca05c…: blocked: this relay requires a verified Namecoin .bit NIP-05 identity (publish a kind:0 first)
```

The reject is **correct and intentional** — mstrofnone's NIP-05 is `_@nostr.wine`, not a `.bit` identifier, so the relay's strict-mode policy engaged exactly as designed. The NIP-20 `OK` frame round-tripped back to Amethyst over the same WebSocket.

Same "hello gated .bit world" event also sent successfully via standalone Python WebSocket client from M's laptop hitting the IP directly (event `b0070be2…`) — identical reject. CLI and Amethyst paths converge.

### What's left

Section 14.2 of `bitrelay.txt`:
- Add `mstrofnone` to `d/testls.nostr.names` (or register `id/mstrofnone`) so this account can actually post.
- IPv6 support in `parseIpAddresses` + `BitRelayDns`.
- Filed-but-not-shipped: better strfry rejection-2 message; namecoin/guix.sigs builder pubkey publication.

mstrofnone:
## Bug 5 fixed: single-identity object form on `d/` names

Pushed as commit [`b5713d3`](https://github.com/mstrofnone/amethyst/commit/b5713d3c4) (Quartz `NamecoinNameResolver`).

### Background

ifa-0001 doesn't mandate that domain records use the `nostr.names` sub-dictionary. Operators who own a name outright commonly publish:

```json
{"nostr": {"pubkey": "<hex>", "relays": ["wss://..."]}}
```

— the same shape `id/` records use. Before this fix the `d/`-namespace branch only accepted `{"nostr":"<hex>"}` (string form) or `{"nostr":{"names":{...}}}` (extended form). `d/mstrofnone` published with the single-identity object form silently failed to resolve.

### Resolution rules

1. If `nostr.names` is present, it wins for any sub-identity. Names dict consulted first; exact local-part hit returns. A root lookup with no `names["_"]` falls back to the bare `pubkey` field if present.
2. If `nostr.names` is absent, only the root local-part (`_`) resolves — there's no sub-identity dictionary, so `alice@example.bit` against a single-identity record correctly returns null.
3. Non-root lookups that miss the names dict do **not** silently fall back to the bare `pubkey` (would hand `alice@example.bit` the root operator's identity).

### End-to-end smoke test

Server: relay.testls.bit running [mstrofnone/strfry-namecoin-policy#1](https://github.com/mstrofnone/strfry-namecoin-policy/pull/1) with the matching server-side fix deployed.

Amethyst on emulator-5554 (sdk_gphone64_arm64, API 36):

```
D BitRelayDns: relay.testls.bit -> 23.158.233.10 (from Namecoin)
D TlsaTrustManager: TLSA match for relay.testls.bit: usage=DANE_TA selector=SUBJECT_PUBLIC_KEY_INFO matchingType=SHA_256 chainIndex=-1 (DANE-only)
D Relay relay.testls.bit: OnOpen (ping: 217ms, using compression)
D RelaySpeedLogger: -- Kind 0 (2 - 1kb); 30b6yn (1), vAG5Pw (1); relay.testls.bit (2)
D RelaySpeedLogger: -- Kind 1 (2 - 1kb); 30b6yn (2); relay.testls.bit (2)
D RelaySpeedLogger: -- Kind 1 (1 - 1kb); 30b6yn (1); relay.testls.bit (1)
```

Server side:
```
[strfry-namecoin-policy info] kind0 f03afd78… verified Namecoin NIP-05 "_@mstrofnone.bit" → 43185edecb675892…
[strfry-namecoin-policy info] kind0 fe29a3d6… verified Namecoin NIP-05 "id/mstrofnone" → 43185edecb675892…
[Writer] Inserted event. id=31d0707a2e7603efe179d4305b62994866b13759e409be8c47aac48d7a1f35b5 levId=6
```

Both NIP-05 forms resolve and gate correctly:
- `_@mstrofnone.bit` → `d/mstrofnone` (single-identity object form)
- `id/mstrofnone` → `id/mstrofnone` (existing path)

Round-trip from Amethyst: kind:1 "hello gated .bit world from Amethyst round2" sent → relay accepted → echoed back via subscription. **Connect ✅ Send ✅ Receive ✅** for both identities.

### Test count

Quartz `NamecoinNameResolverTest` is now 19 tests (was 15, +4 for the single-identity path). **211/211 namecoin tests pass** across 8 suites.

vitorpamplona:
Hmm... looks like this PR and the other PR have a lot of duplicated classes. Are they the same?

mstrofnone:
@vitorpamplona you're right that the diffs overlap, but they are not the same — #2612 is **stacked on top of** #2595, not a duplicate of it.

#2612 was branched from `feat/bit-relay-resolution` (the head of #2595), not from `main`. Because GitHub diffs the head branch against the PR target, all of #2595's commits showed up inside #2612's diff too. That's the "duplication" you're seeing.

I just rebased and force-pushed #2612. Once #2595 merges, the noise in #2612 will collapse to just the 7 import-specific commits.

### What is _new_ in #2612 vs #2595

| Commit | What |
|---|---|
| `fa378bc` | docs: link ifa-0001 references to the spec source |
| `95c70f4` | **`NamecoinImportResolver`** — ifa-0001 §"import" expansion (new file, +266 lines) |
| `02897e1` | distinguish malformed-JSON from missing-field (caller can react differently) |
| `451bd7b` | **`BitRelayDns`** + `parseIpAddresses` — resolve `.bit` to A-record IP for OkHttp Dns + NIP-11 (new file, +134 lines) |
| `b283c8f` | **`AiaStapledPubkeyExtractor`** — accept ncgencert AIA-stapled SPKI as TLSA trust anchor (new file, +309 lines) |
| `06c21cb` | force Tor route for `.bit` relays when `onionRelaysViaTor` is on |
| `37c29ca` | accept single-identity `{nostr:{pubkey,relays}}` on `d/` names |

### Files

**New (only in #2612):**
- `quartz/.../namecoin/NamecoinImportResolver.kt`
- `quartz/.../namecoin/AiaStapledPubkeyExtractor.kt`
- `amethyst/.../service/namecoin/BitRelayDns.kt`
- 4 corresponding test files

**Modified (extending #2595's files):**
- `NamecoinNameResolver.kt` — adds `expandImports` call in `performLookupDetailed`, single-identity object form on `d/` names, `parseIpAddresses`
- `BitRelayResolver.kt` — parses value once, expands imports, hands JsonObject to overloaded parser methods
- `TlsaVerifier.kt` — new SPKI-pinning predicate that lets the AIA-stapled pubkey participate in `*-TA` SPKI matches
- `TlsaTrustManager.kt` / `TlsaConnectionPolicy.kt` — wire AIA-stapled SPKI into the trust path
- `AppModules.kt` — wire `BitRelayDns` into OkHttp client, force Tor for `.bit` onions

So no class is defined in both PRs; #2612 only **adds new classes** and **extends classes from #2595**. Once #2595 lands the diff here will be clean.

If you'd prefer, I can also convert this to a draft until #2595 merges.

PR #2620, opened 2026-04-28
feat(nests): bigger avatars + clearer mute/speaking state in participant grid

vitorpamplona:
100dp avatars across stage and audience tabs, names centered under each
cell, and the avatar border now distinguishes actively speaking (green)
from publishing-but-muted (red) at a glance — the existing mic-pill
keeps its semantics underneath.

UsernameDisplay gains an optional textAlign so the grid can opt in to
centered labels without disturbing other call sites.


davotoula:
I added some fixes to cache and display to prevent squashed pfp. Hopefully it survives!

PR #2700, opened 2026-05-02
fix(quartz): harden event deserializer (size caps + drop intern() + WS frame cap)

davotoula:
Implements §2.3 of the Quartz security review (Finding #5) in three layers, plus a follow-up code-review pass that simplified the patch.

### Layer 3 — drop attacker-controlled `String.intern()`
Tag values, `id`, and `pubKey` are no longer interned. Each interned string is permanently allocated in the JVM StringTable, so a hostile relay could grow process memory without bound by flooding events with unique values. Tag *keys* (`"p"`, `"e"`, `"a"`, …) remain interned since they're a small protocol-defined set.

### Layer 2 — size / count / length caps
New `EventLimits` constants (defense-in-depth):

| Cap | Value | Real-world max in `LargeDBTests` fixture (248K events) |
|---|---|---|
| `MAX_CONTENT_LENGTH` | 64 KB | 47 KB |
| `MAX_TAG_COUNT` | 2 000 | 204 |
| `MAX_TAG_ELEMENTS_PER_TAG` | 256 | 104 |
| `MAX_TAG_VALUE_LENGTH` | 16 KB | ~10 KB |
| `MAX_RELAY_MESSAGE_LENGTH` | 256 KB | (used by Layer 1) |

Wired into both deserializer paths:
- **Jackson** (`EventDeserializer`, `TagArrayDeserializer`): early termination during streaming parse so a hostile relay can't grow allocations beyond the bound before being rejected.
- **kotlinx** (`EventKSerializer`, `TagArrayKSerializer`): inline checks during decoding from `JsonElement`, plus post-parse `validateContent`.

Throws `IllegalArgumentException` on overflow.

### Layer 1 — WebSocket message-size cap
`BasicOkHttpWebSocket` rejects any incoming message exceeding `MAX_RELAY_MESSAGE_LENGTH` and closes the socket with RFC 6455 close code 1009 ("Message Too Big"). Catches the giant-frame OOM class before the per-event caps fire.

## Why this PR

The original review labeled this finding as HIGH because, when paired with §2.1 (signature verification on receive), a single oversized event from a hostile relay could OOM the client; a slow drip of unique tag values could leak memory permanently via the StringTable.

## Test plan

- [x] `EventLimitsTest` (commonTest, kotlinx path) — 5 tests covering accepts-within-limits + 4 rejection cases
- [x] `JacksonEventDeserializerLimitsTest` (jvmTest, Jackson path) — same matrix
- [x] `JacksonInternBehaviorTest` (jvmTest) — 3 tests verifying tag values / id / pubKey not interned, tag keys still interned
- [x] `BasicOkHttpWebSocketAcceptIncomingTest` (jvmAndroidTest) — 4 tests: forward at-cap, drop+close at-cap+1, pin RFC code 1009, pin reason string
- [x] Full `:quartz:jvmTest` — **1881 tests / 188 suites all green** after rebase onto current `upstream/main`
- [x] `LargeDBTests` (real-world fixture, 248K events) — loads cleanly; cap headroom verified

## Commits

1. `f8577053b` — Layer 3: drop attacker-controlled intern
2. `af1bcfcdb` — Layer 2: event/tag/value caps
3. `dcd288018` — Layer 1: WS message-size cap
4. `e2fc88604` — Code review: simplify (single-pass validate, dedup test fixtures, lambda closer, drop redundant `index` counter, named close-reason constant)

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
@@ -105,7 +105,7 @@ object EventKSerializer : KSerializer<Event> {
             throw IllegalArgumentException("Event not found")
         }
 
-        EventLimits.validate(content, tags)
+        EventLimits.validateContent(content)
 
         return EventFactory.create(id, pubKey, createdAt, kind, tags, content, sig)
     }
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
@@ -24,6 +24,7 @@ import com.vitorpamplona.quartz.nip01Core.core.Event
 import com.vitorpamplona.quartz.nip01Core.core.HexKey
 import com.vitorpamplona.quartz.nip01Core.core.Kind
 import com.vitorpamplona.quartz.nip01Core.core.TagArray
+import com.vitorpamplona.quartz.nip01Core.limits.EventLimits
 import com.vitorpamplona.quartz.utils.EventFactory
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.descriptors.SerialDescriptor
@@ -103,6 +104,8 @@ object EventKSerializer : KSerializer<Event> {
         if (pubKey.isEmpty()) {
             throw IllegalArgumentException("Event not found")
         }
+
+        EventLimits.validate(content, tags)
 
         return EventFactory.create(id, pubKey, createdAt, kind, tags, content, sig)
     }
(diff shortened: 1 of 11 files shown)


davotoula:
@vitorpamplona for your review.

- Sets sensible limits to prevent DOS/OOM
- Reduces usage of string.intern to prevent DOS/OOM

Testing:

- Using real values from existing tests to adjust caps and thus pass existing tests

vitorpamplona:
Interning is what keeps our memory manageable because most if nostr is duplicated keys, event ids, and tags everywhere. A download of 10,000 200-people follow list uses 120MB of RAM just for the keys themselves. With interning, that drops to 64KB. Same for event IDs, a-tags, etc. 

I was actually thinking on building our own interning structure to make sure it's also available on iOS and Native. 

The other limits are probably too low. If a relay wants to crash amethyst, it can just send all its events and ignore the filter. It will have the same effect.

vitorpamplona:
Also.. modern String intern is also released when no event uses them. It's not hardset forever like the AI said.

davotoula:
Would  WebSocket message-size cap be worth keeping? (just layer 1)

vitorpamplona:
OkHttp definitely has a 16 MiB limit for sending; it reuses the same buffer for receiving, which means that, in practice, that message size also triggers a socket close. OkHttp does manage the reuse of sending and receiving memory buffers. OOO from the socket itself should never happen. 

I was just looking at the relay side of things and I think there is a consensus that these types of limits are kind of arbitrary. StrFry uses it mostly because that is how C people think, but the others just block by message size, like the 16Mb from http.

davotoula:
thank you for the thorough feedback. I think it's a non issue now, closing.

PR #2701, opened 2026-05-02
fix(quartz): validate event id, pubKey, sig hex/length and kind range

davotoula:
**Stacked on top of #2700.** Until that PR merges, this PR's diff includes the 4 §2.3 commits (`f8577053b`, `af1bcfcdb`, `dcd288018`, `e2fc88604`); they will auto-collapse from the diff once #2700 lands. The 3 commits in scope for review **here** are the §2.4 commits at the top of the list.

Implements §2.4 of the Quartz security review (Finding #11) — post-parse hex/length/range validation of `Event` structural fields.

## Validation rules

`EventLimits.validateFields(id, pubKey, sig, kind)` is called post-parse from both Jackson `EventDeserializer` and kotlinx `EventKSerializer`:

| Field | Rule | Why |
|---|---|---|
| `id` | 64-char hex | 32-byte SHA-256 of the canonical event |
| `pubKey` | 64-char hex | x-only secp256k1 pubkey |
| `sig` | empty OR 128-char hex | NIP-17 (kind:14 DMs) and NIP-37 drafts are stored as `Event` with `sig=""` because the inner-event rumor is unsigned per NIP-59; ~5 K such events exist in production caches |
| `kind` | `0..65535` | NIP-01 unsigned 16-bit |
| `createdAt` | **not bounded** | Nostr legitimately uses ancient and future timestamps (scheduled posts, archive replay, NIP-23 backdated long-form, NIP-03/1040 OpenTimestamp attestations). The id pre-image binds `created_at` to the signature, so `verify()` detects tampering; sort-order manipulation is a feed/UI concern |

Without this, malformed events sat in the cache permanently and only blew up later inside `Hex.decode` if signature verification ran.

## Implementation

- New: `EventLimits.validateFields(...)` + `MAX_KIND` constant.
- Wired into both deserializers post-parse. The new (more informative) length check supplants the old empty-pubKey early-throw.
- Test helper `EventLimitsTestSupport` extended with id/pubKey/sig/kind/createdAt overrides + new `assertRejectsBecause(parse, json, fragment)` helper shared by both test classes.

## Test plan

- [x] `EventLimitsTest` (commonTest, kotlinx path) — 14 tests including 4 §2.3 carries and all §2.4 rejection cases + accepts-empty-sig + accepts-ancient-createdAt
- [x] `JacksonEventDeserializerLimitsTest` (jvmTest, Jackson path) — same matrix
- [x] Updated 4 fixtures in `KotlinSerializationMapperTest` from synthetic short hex to realistic 64/128-char hex
- [x] Full `:quartz:jvmTest` — **1901 / 188 suites all green** including `LargeDBTests` (248 K real events) and `JacksonInternBehaviorTest` from §2.3
- [x] kotlin-reviewer: APPROVE (0 CRITICAL / 0 HIGH; 1 MEDIUM cosmetic, 3 LOW notes)

## Commits in this PR (top three)

1. `91f0122aa` — fix: validate event id, pubKey, sig hex/length and kind range
2. `d854f91e2` — chore: gitignore .claude/scheduled_tasks.lock
3. `b7f0c62ef` — refactor: drop redundant pubKey-empty check + dedup limits tests (post-/simplify cleanup)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,6 +158,7 @@ TASKS.md
 
 # Claude Code local settings
 .claude/settings.local.json
+.claude/scheduled_tasks.lock
 
 # Downloaded VLC binaries (vlc-setup plugin)
 desktopApp/src/jvmMain/appResources/linux/
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
@@ -101,10 +101,6 @@ object EventKSerializer : KSerializer<Event> {
             }
         }
 
-        if (pubKey.isEmpty()) {
-            throw IllegalArgumentException("Event not found")
-        }
-
         EventLimits.validateFields(id, pubKey, sig, kind)
         EventLimits.validateContent(content)
 
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/kotlinSerialization/EventKSerializer.kt
@@ -105,6 +105,7 @@ object EventKSerializer : KSerializer<Event> {
             throw IllegalArgumentException("Event not found")
         }
 
+        EventLimits.validateFields(id, pubKey, sig, kind)
         EventLimits.validateContent(content)
 
         return EventFactory.create(id, pubKey, createdAt, kind, tags, content, sig)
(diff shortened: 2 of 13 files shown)


davotoula:
@vitorpamplona this is also ready for review. Is it worth the effort to validate event id, pubKey, sig hex/length and kind range?

If validation is implemented it would have to be updated if specification for those fields ever change...

vitorpamplona:
So, I don't know if we should do this in the deserialization... Since we have Drafts (which can host any incomplete event saved as a string), there might be use case for deserializing and then passing to the host app to fix it. 

It's also important to notice that checking the size of each tag might slow things down a bit in our regular speeds (5000 events/second), depending on the sized of the tag array of said events. The isHex64 is also quite slow.. I have been trying to remove it from most of the hot path. 

Keep in mind that apps generally receive duplicated events. So, on Amethyst, the event should go into LocalCache to check if it already exists before we validate and verify (and intern... which is not there today) the event.

davotoula:
**Cryptography already covers the threat.** If `sig` is malformed, `event.verify()` returns false and the event is dropped. If `sig` is empty (Draft / NIP-59 rumor), there's nothing to verify by design. Strict          
  hex/length validation in the deserializer adds nothing the signature check doesn't already give us.

PR #2707, opened 2026-05-02
feat(namecoin): resolve `import` items per ifa-0001 (NIP-05 only)

mstrofnone:
## What

Adds a new `NamecoinImportResolver` that follows the `import` field of a Namecoin Domain Name Object value, recursively merging the imported values into the importing object before `NamecoinNameResolver` reads the `nostr` field.

## Why

The 520-byte per-name limit on Namecoin is tight when a single record needs `ip`, NIP-05 names, TLS, and per-subdomain `map` entries. A common workaround (used by `testls.bit` and others) is to keep the apex record small by delegating shared blocks into a sibling name via `"import":"dd/<name>"` per [ifa-0001 §"import"](https://github.com/namecoin/proposals/blob/master/ifa-0001.md).

**Without import support, the search bar already-correct routing of `.bit` inputs through `NamecoinNameResolver` still fails** for any deployment that uses imports: the resolver sees no `nostr` field at `d/testls` and gives up before consulting the imported `dd/testls` that actually carries the names block.

```
$ namecoin-cli name_show d/testls
"value": "{\"import\":\"dd/testls\",\"ip\":\"...\",\"map\":{...}}"   ← no "nostr" key

$ namecoin-cli name_show dd/testls
"value": "{\"nostr\":{\"names\":{\"_\":\"460c25e6...\",\"m\":\"...\"}},...}"
```

## How

Per ifa-0001 §"import":
- Importer items take precedence over imported items (including `null`, which suppresses the imported counterpart).
- Recursion is supported up to 4 levels (the spec minimum) by default. Cycles are broken by a visited-set; over-budget chains are silently truncated, with the importer's own items still taking effect.
- Multiple imports in one array merge later-wins (the array's own order), with the importing object on top of all of them.
- Subdomain selectors (the optional 2nd element of each inner array) are resolved through a `map` walk inside the imported value, with the standard exact-label > `*` wildcard > `""` default rules.
- Three short-hand value forms are accepted alongside the canonical array-of-arrays: `"import":"d/foo"`, `"import":["d/foo"]`, and `"import":["d/foo","selector"]`. They map onto the canonical representation losslessly.
- A failed import lookup (name not found, malformed JSON, network error) is treated as `{}` rather than failing the whole record, so transient ElectrumX hiccups don't kill resolution. The importing object's own items still apply.

Wires the expander into `NamecoinNameResolver.performLookup` and `performLookupDetailed`. Records without an `import` key skip the resolver entirely (zero extra I/O), so non-import names pay no cost.

## Tests

**20 new tests** in `NamecoinImportTest`, all hermetic (no network):
- **Unit** (resolver in isolation): no-import passthrough, all four shorthand value forms, importer precedence, null suppression, depth-4 recursion happy path, over-budget truncation, lookup failure, malformed JSON, malformed `import` value, cycle protection, multi-label selector descending in DNS order, wildcard fallback.
- **Integration** with `NamecoinNameResolver`: bare and named NIP-05 across an import (mirroring the real-world `testls.bit → dd/testls` deployment); regression guard that no-import records issue exactly one ElectrumX query; importer-wins on `nostr.names`; lenient failure when the imported boilerplate is unreachable; `resolveDetailed` returns `Success` and `NoNostrField` correctly across imports.

```
TEST-com.vitorpamplona.quartz.nip05.namecoin.NamecoinImportTest.xml: 20 tests, 0 failures, 0 errors
TEST-com.vitorpamplona.quartz.nip05.namecoin.NamecoinNameResolverTest.xml: 15 tests, 0 failures, 0 errors
```

All 15 existing `NamecoinNameResolverTest` cases continue to pass — no regressions.

## Scope note

This PR is intentionally scoped to the **NIP-05 / search path only**, so it can land independently of #2595. The relay-resolution side of `import` support (`BitRelayResolver`) is part of #2595's stack and will be wired up when that lands. Pure unit tests in `NamecoinImportResolver` are reusable for the relay path with no changes.

## Diff stats

```
3 files changed, 840 insertions(+), 4 deletions(-)
 quartz/.../namecoin/NamecoinImportResolver.kt    | new (267 lines)
 quartz/.../namecoin/NamecoinNameResolver.kt      | +31 -4
 quartz/.../namecoin/NamecoinImportTest.kt        | new (542 lines)
```

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip05DnsIdentifiers/namecoin/NamecoinImportResolver.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip05DnsIdentifiers/namecoin/NamecoinImportResolver.kt
@@ -1,0 +1,276 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.vitorpamplona.quartz.nip05DnsIdentifiers.namecoin
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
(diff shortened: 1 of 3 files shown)


mstrofnone:
Tested on an Android emulator (`amethyst_api36`, sdk_gphone64_arm64, API 36) with the `amethyst-play-arm64-v8a-debug` APK built from this branch:

- `testls.bit` → resolves to the `_` entry from `dd/testls` ✅
- `m@testls.bit` → resolves to the `m` entry from `dd/testls` ✅

Default ElectrumX server list (clearnet) was used — no manual Namecoin config needed for the test.

PR #2735, opened 2026-05-05
Add IME padding to MessagesTwoPane Scaffold

vitorpamplona:
## Summary
This change adds IME (Input Method Editor) padding to the Scaffold composable in the MessagesTwoPane screen to ensure proper layout adjustment when the on-screen keyboard is displayed.

## Key Changes
- Added `androidx.compose.foundation.layout.imePadding` import
- Applied `Modifier.imePadding()` to the Scaffold composable in MessagesTwoPane

## Implementation Details
The `imePadding()` modifier automatically adjusts the padding of the Scaffold to account for the space occupied by the IME (soft keyboard) when it's displayed. This prevents UI elements from being hidden behind the keyboard and ensures a better user experience when typing messages in the two-pane chat layout.

https://claude.ai/code/session_012sPZ9KbbkTzdPL2KTn29ak


davotoula:
Was this related to the bug in 3rd party library?

vitorpamplona:
which one? No, I think this was just an oversight when we switched to the DisappearingScaffod where the field end ed up behind the keyboard

PR #2765, opened 2026-05-07
Scheduled Posts (Android)

davotoula:
### Summary

Adds the ability to schedule a short-note post for future publication. Posts are signed at compose time, stashed locally, and broadcast by a background worker when they come due. End-to-end working on Android; desktop deferred.

<img width="300" alt="Screenshot_20260507-202909" src="https://github.com/user-attachments/assets/fe825e1c-3722-44d9-af31-aab4f4fede99" align="top"/> <img width="300" alt="Screenshot_20260507-202822" src="https://github.com/user-attachments/assets/1c528bcf-83ac-438f-9a65-898f89be8fee" />

                                                                                                                                                                                                                
                                        
  ### Architecture                                                                                                                                                                                                                  
                         
  Storage — single JSON file at <filesDir>/scheduled_posts.json                                                                                                                                                                 
  - Shape: { version, posts[] }. Each row carries id, accountPubkey, the full signed event JSON, the relay list, any extras (e.g. NIP-94 imeta side events), publishAtSec, createdAtSec, status, lastAttemptAtSec, attemptCount,
   lastError, terminatedAtSec.                                                                                                                                                                                                  
  - Atomic writes via .tmp + rename. Mutex-guarded singleton ScheduledPostStore. Reactive StateFlow<List<ScheduledPost>> for the UI.                                                                                          
  - Considered SQLite (Quartz already ships an SQLiteEventStore) but its schema is fixed to Nostr events; shoehorning scheduled rows in would have meant a NIP-37-style draft kind. Not worth the complexity for v1.            
                                                                                                                                                                                                                                
  Sign-now, broadcast-later                                                                                                                                                                                                     
  - At compose time we call the existing Account.createPostEvent(template, extras) to sign without broadcasting, with created_at rewritten to the scheduled instant (so the relayed timestamp matches the user's intent, not the
   moment of composition).                                                                                                                                                                                                      
  - At fire time ScheduledPostWorker calls Account.consumePostEvent(event, relays, extras) — the same finalizer the immediate-publish path uses. Media is uploaded during composition, so URLs are already baked into the signed
   event by the time we stash it.                                                                                                                                                                                               
  - Side effect: NIP-46 / NIP-55 remote signers are only consulted at schedule time; the device can fire later even if the signer is unreachable.                                                                               
                                                                                                                                                                                                                              
  Wakeup wiring                                                                                                                                                                                                                 
  - Periodic WorkManager worker on a 15-minute floor (the platform minimum), plus a one-time catch-up enqueued at AppModules.initiate(...) so a freshly opened app immediately drains anything past due.
  - Outbox-pattern claimDuePosts(now): atomic PENDING → PUBLISHING flip with rowcount check, so two workers racing in the same window cannot double-publish.                                                                    
  - After client.publish(), the worker polls Quartz's new INostrClient.pendingPublishRelaysFor(eventId) every 500 ms (30 s timeout) and only marks SENT once at least one relay has acked OK. Avoids the "fire-and-forget then
  claim success" failure mode.                                                                                                                                                                                                  
  - Boot survival comes for free via AutoRestartReceiver re-initializing modules.                                                                                                                                               
                                                                                                                                                                                                                                
  Multi-account                                                                                                                                                                                                                 
  - Rows key on accountPubkey. The worker resolves the right Account via Amethyst.instance.accountsCache. The list screen filters to the active account; switching accounts switches the visible queue.                       
  - Logout cleanup is centralized in AccountSessionManager.logOff(): deletes all rows owned by the departing account after the regular LocalPreferences.deleteAccount + accountsCache.removeAccount. The logout dialog reads the
   store flow live and warns when count(unpublished) > 0.                                                                                                                                                                     
                                                                                                                                                                                                                                
  Notifications                                                                                                                                                                                                               
  - New ScheduledPostNotifier posts a system notification on fire success and on terminal failure (worker IO threads — @Volatile cache for the NotificationChannel reference).                                                  
                                                                                                                                                                                                                                
  Retention / cleanup                                                                                                                                                                                                         
  - Terminal rows (SENT, CANCELLED, terminal FAILED) are purged on store load: 7 days for SENT/CANCELLED, 30 days for FAILED. Fallback to lastAttemptAtSec then createdAtSec for legacy rows missing terminatedAtSec.           
                                                                                                                                                                                                                                
  Hardening picked up along the way  
  - Replaced bechToBytes() chains at three account sites with the null-safe decodePublicKeyAsHexOrNull / decodePrivateKeyAsHexOrNull helpers — a malformed npub no longer half-completes a logout or crashes the dialog tree.   
  - Switched list-screen formatters from java.text.DateFormat/SimpleDateFormat (not thread-safe) to java.time.format.DateTimeFormatter.                                                                                         
  - @Volatile on PoolEventOutbox.eventOutbox map ref + PoolEventOutboxState.relaysRemaining — the new poll surface reads from the WorkManager IO thread; mutations still happen on NostrClient's scope.                         
  - Compose locale now read from LocalConfiguration.current.locales[0], not Locale.getDefault() — fixes Turkish I/i casing drift.                                                                                               
                                                                                                                                                                                                                                
  ### Features                                                                                                                                                                                                                      
                                                                                                                                                                                                                                
  - Schedule a post. Toolbar toggle in the composer; tapping opens a Material3 two-stage Date → Time picker. Selected time rounds up to the next quarter-hour (matches the worker floor). 24-hour mode honors                   
  DateFormat.is24HourFormat. Default suggested time: one hour from now.                                                                                                                                                       
  - Manage scheduled posts. New "Scheduled posts" drawer entry leads to a list grouped by day with sticky headers ("Today", "Tomorrow", weekday + date for upcoming, full date past). Each card shows status chip, absolute     
  scheduled time + live relative countdown, content preview, media thumbnails, error line for failures, swipe-to-delete, plus per-row "Send now" / "Delete" actions with confirms.                                              
  - Drawer badge. Pending count surfaces next to the drawer entry.                                                                                                                                                            
  - Always-on warning. When always-on notifications are off, the picker shows a ReliabilityWarning explaining the worker may not run on schedule. Wording adapts for single- vs multi-account setups.                           
  - Multi-account support. Each account sees its own queue. Switching accounts swaps the list. Scheduling works for any logged-in account.                                                                                      
  - Logout cleanup + warning. Logging out deletes that account's scheduled posts. If any are still pending/publishing/failed, the confirmation dialog calls out the count before destruction.                                   
  - Send-now / Cancel from list. Manual push enqueues an immediate fire; cancel marks the row terminal without publishing.                                                                                                      
  - OK-ack confirmation. A row is only marked SENT after at least one relay has OK'd the event, with a 30-second timeout — so users see a true-positive "sent" status, not just "broadcast attempted".                          
  - Localization. 44 new strings translated into cs-rCZ, de-rDE, pt-rBR, sv-rSE. Existing strings reused (R.string.cancel, confirm, next) where they fit. Hardcoded Color(0xFF…) literals replaced with                         
  MaterialTheme.colorScheme.* tokens for proper light/dark theming.                                                                                                                                                             
  - Tests. 27 store unit tests covering happy path, race protection on claimDuePosts, reload semantics, status transitions, publishNow, removeForAccount, and the retention sweep. Plus media-extraction tests.                 
                                                                                                                                                                                                                                
  ### Limitations                             
                                                                                                                                                                                                                                
  - Device must be on and have run the app recently. Worker scheduling is local — there is no relay-side delivery agent. If the device is off or the OS aggressively kills WorkManager, the post fires when the worker next runs
   (catch-up on app open).                                                                                                                                                                                                    
  - Reliable delivery requires always-on notifications. Without the foreground service, Doze and App Standby can delay the periodic worker indefinitely. Picker warns the user.                                                 
  - Android only. No Desktop port in this branch; the desktopApp/ module is untouched.                                                                                                                                          
  - Posts only fire on the device they were scheduled on. No cross-device sync — there is no NIP-37-style pending draft replicated to other clients.                                                                            
  - TextNote (kind 1) only. Voice notes, polls, anonymous posts, and NIP-95 attachments fall through to the immediate-broadcast path.                                                                                           
  - Granularity ~15 minutes. WorkManager's periodic minimum. The picker rounds to quarter-hours so the UX matches reality. "Exactly 9:00" is not guaranteed.                                                                    
  - Cancelling does not delete uploaded media. Media is uploaded at compose time (Blossom / NIP-96). Cancelling a scheduled post leaves blobs on the host server — same behavior as cancelling a draft after upload.            
  - Remote signers (NIP-46/NIP-55) are not re-consulted at fire time. The signed bytes are stashed once at compose time. A user expecting "re-sign at fire" will not get it.                                                    
  - OK-ack timeout is best-effort. A relay that replies OK but later evicts the event still counts as SENT. Longer-tail propagation is not tracked.                                                                             
  - No JSON schema versioning enforcement on read. version field exists; loader does not currently warn on version > 1.                                                                                                         
                                                                                                                                                                                                                                
###   Possible future enhancements                                                                                                                                                                                                  
                                                                                                                                                                                                                                
  - Cross-device sync via a NIP-37-style encrypted draft event with a publish_at tag, so any logged-in device with the worker active can fire.                                                                                  
  - Per-minute precision via AlarmManager.setExactAndAllowWhileIdle for the next-due post (costs SCHEDULE_EXACT_ALARM + Play Store scrutiny).                                                                                 
  - Schedule-time preset chips ("1 hour from now", "Tomorrow 9 am", "Next Monday 9 am").                                                                                                                                        
  - Edit-while-scheduled — today the user must cancel and recompose.                                                                                                                                                            
  - Tap-to-expand row with full content + relay list, or tap-to-edit.                                                                                                                                                           
  - Retry semantics for FAILED rows — surface "Retry" instead of "Send now"; offer "Edit" if retry fails repeatedly.                                                                                                            
  - Empty-state CTA that returns to the composer rather than a static hint.                                                                                                                                                     
  - Scheduling for voice / polls / anonymous / NIP-95 post types.                                                                                                                                                               
  - Per-account opt-out setting that hides the toolbar button + drawer entry.                                                                                                                                                   
  - Maximum pending cap (e.g. 100) to prevent runaway disk usage.                                                                                                                                                               
  - Auto-prompt to enable always-on at schedule-creation time when it's currently off.                                                                                                                                          
  - Restored-on-other-device guard — if releaseClaim cycles N times for a row whose account isn't loaded on this device, mark it FAILED so it stops churning.                                                                   
  - Desktop port when product priorities allow.                                                                                                                                                                                 
  - Automated end-to-end test with WorkManagerTestInitHelper to catch worker-scheduling regressions in CI.                                                                                                                      
  - Robolectric / Compose UI tests for the logout dialog warning logic and AccountSessionManager.logOff cleanup wiring.


vitorpamplona:
Very cool

davotoula:
Woop. Woop!

PR #2775, opened 2026-05-08
fix(scheduled-posts): explicit POST_NOTIFICATIONS check before notify

vitorpamplona:
Lint flagged the notify() call as MissingPermission because there was no
visible check that POST_NOTIFICATIONS was granted. Gate the call behind
areNotificationsEnabled() — matching the pattern used in
EventNotificationConsumer — and catch SecurityException for the
narrow window where permission could be revoked between the check and
the notify().


davotoula:
Thank you!

PR #2811, opened 2026-05-09
feat(note): copy raw JSON from note dropdown

davotoula:
Adds a **Copy raw JSON** action to the note 3-dot dropdown, directly below *Copy Note ID*. Pretty-printed via the existing `JacksonMapper.toJsonPretty` (one tag array per line).

Jackson serialization runs on `Dispatchers.Default` to keep large events off the UI thread.


davotoula:
merging as I need it for some other work...

PR #2820, opened 2026-05-10
Align favourites icon, add to default bar, redesign empty state

davotoula:
Improve content discovery especially for brand new users

1. Make favourite feeds icon default in bottom bar
2. Empty favourite feed screen has new call to action button
3. Button navigates directly to discovery of feeds
4. Back button goes back to favourite feeds screen


<img width="350" alt="Screenshot_20260510-084901" src="https://github.com/user-attachments/assets/97dda516-0983-4134-9afa-ce6708bed2d7" valign="top"/> <img width="350"  alt="Screenshot_20260510-084913" src="https://github.com/user-attachments/assets/ac947a4a-844d-458d-ae7a-57f1df51fe7a" />
 
<img width="350" alt="Screenshot_20260510-084929" src="https://github.com/user-attachments/assets/e73d5d27-a422-4bdf-a124-63951deda4e6" /> <img width="350" alt="Screenshot_20260510-084941" src="https://github.com/user-attachments/assets/051421f3-514b-44a1-92f2-9530c81a5ddd"  valign="top"/>

## Summary                              
  - Bottom-nav `FAVORITE_ALGO_FEEDS` now uses `MaterialSymbols.AutoAwesome` (multi-stars), matching the icon already used in the settings drawer and the top-bar feed-filter spinner. The previous single `Star` was the odd one out.                                                                                                                                                                                                                         
  - Adds `FAVORITE_ALGO_FEEDS` to `DefaultBottomBarItems` so new installs see the entry by default (existing users keep their persisted layout).                                                                                
  - Redesigns the empty state of the Favorite Feed Algorithms screen — was a single line of grey text telling users to open Discover, find an algo, and star it. Now shows the multi-stars icon, a "Pin your favorite algorithms" headline, and an **Add some feeds** primary button that deep-links to the Discover screen's algo tab.                                                                                                             
  - To make the deep-link actually work, `Route.Discover` becomes `data class Discover(val initialTab: DiscoverTab? = null)`. `DiscoverScreen` scrolls the pager to `initialTab` on first composition; the bottom-bar Discover tap still uses the saved `rememberForeverPagerState` position.                                                                                                                                                                
                                                                                                                                                                                                                                
  ## Test plan                                                                                                                                                                                                                  
  - [x] Fresh install: bottom bar shows the multi-stars Favourites entry by default                                                                                                                                             
  - [x] Open Favourites with no feeds → empty state renders with icon, headline, sub-line, and CTA                                                                                                                              
  - [x] Tap **Add some feeds** → lands directly on the Algos tab in Discover                                                                                                                                                    
  - [x] Star an algo, press Back → returns to Favourites with the new entry visible                                                                                                                                             
  - [x] Bottom-bar Discover tap still lands on the user's last-viewed tab (not forced to Algos)

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/AppNavigation.kt
@@ -233,7 +233,7 @@ fun BuildNavigation(
         composable<Route.Home> { HomeScreen(accountViewModel, nav) }
         composable<Route.Message> { MessagesScreen(accountViewModel, nav) }
         composable<Route.Video> { VideoScreen(accountViewModel, nav) }
-        composable<Route.Discover> { DiscoverScreen(accountViewModel, nav) }
+        composableArgs<Route.Discover> { DiscoverScreen(it.initialTab, accountViewModel, nav) }
         composableArgs<Route.Notification> { NotificationScreen(it.scrollToEventId, accountViewModel, nav) }
         composableFromEnd<Route.Polls> { PollsScreen(accountViewModel, nav) }
         composableFromEnd<Route.Communities> { CommunitiesScreen(accountViewModel, nav) }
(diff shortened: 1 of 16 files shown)


vitorpamplona:
I don't think it should be the last one.. people are way too used to notifications being the last icon on the bottom row.

davotoula:
Fair point, second last... or just not have it in by default, people can add it using Nav bar configuration?

I would still prefer it's there by default for new installs. People can remove it when they get used to the app?

vitorpamplona:
Either way works, we can leave it in..

davotoula:
Swapped. Also this is only for new installs. Existing users will keep their current setup (unless they click restore to defaults).

Issue #383, opened 2023-04-29
[FEATURE] offer different sorting methods for follows/followers lists


tcheeric:
**Is your feature request related to a problem? Please describe.**
Currently, the followers list seems to appear in a random order. The follows list is displayed in a chronological order. 

**Describe the solution you'd like**
Offer three different ways of ordering:
- by zap amount (account with the highest zap amount on top) 
- chronological order (based on the date of the follow) 
- by their respective follower count (based on the number of their followers, account with highest follower count on top) 
The user can switch from one to the other. They can determine their default sorting algorithm

**Bounty (in Bitcoin sats) offered for the implementation**
100k sats

kenzaOut16:
Hello, 
I would be happy to contribute and work on this. Is it possible ?

vitorpamplona:
For sure.. feel free to send a Pull Request. 

The issue is for 2 and 3. There is no way to know the follow date, only the last time a given user changed the following/follower file. There is also no way to easily compute follower count to sort people. There are some off spec solutions from private relays on follower count, but it's trusted information from the relay account.

The current order for following is the order in the event itself. If clients follow that, the last person the user followed comes first in the list.

tcheeric:
The spec indeed encourages relays to delete old contact lists. You won't be able to reliably get the following/follower date. 
> Relays and clients SHOULD delete past contact lists as soon as they receive a new one.

kenzaOut16:
Hello,
I underestimated the issue initially and spent the night working and looking through the code. Unfortunately, there is no direct way to retrieve the following/follower date and it's a bit complicated.

tcheeric:
Hi there, 

Correct, this is not possible, (or rather very tricky). It was already mentioned in this thread. I'm sorry to hear you've spent so much time on it. I will update the FR accordingly. 

You may still be able to sort based on the total amount of sats received and sent. Do you fancy look into it? 

Best
-- 
 Sent with Tutanota, enjoy secure & ad-free emails. 



1 May 2023, 14:36 by ***@***.***:

kenzaOut16:
Hi,

Thank you for letting me know about the update. I'm glad to hear that we can sort based on the total amount of sats received and sent. I would be happy to take a look at it and see if I can find a solution.

0ceanSlim:
I just wanted to add I referenced this issue here: note1rzu45dvzs8djt8lg83n45a30uhqh46uuha2pesa0mdjn8d4edvcqw3t8x7
[...]

PR #2836, opened 2026-05-11
Add zap sorting for profile follow lists

skotchdopolepet:
## Summary
- add a persisted profile-list sort preference for Follows and Followers tabs
- add a Default/Sats received selector to both profile user-list tabs
- reuse the existing profile received-zaps index to put users who have zapped the profile higher while preserving the existing default order as a tie-breaker

Refs #383. This avoids the follow-date and follower-count approaches called out as unreliable in the issue thread, and implements the zap-based sorting path plus a default preference.

## Tests
- ./gradlew.bat --no-daemon :amethyst:compileFdroidDebugKotlin
- ./gradlew.bat --no-daemon :amethyst:compilePlayDebugKotlin
- git diff --check
- pre-commit spotless/static analysis


davotoula:
Thank you for your massive contribution!

There are two main issues with submitting these large pull requests:

1. Very old github issues have not been pruned for whether they are still relevant: the feature might have already been introduced via another pull request, the feature is not valid due to nip protocol changes, the feature is not viable due to amethyst decentralized nature, etc. Research is required before attempting implementation. 

2. The main problem after implementation is testing: regression testing, security testing, actual feature testing, user experience testing. All this often requires manual testing on a real device and analysing logs on the device and events on relays. 

Maybe look into adding automated on device testing like Espresso + Hilt.

vitorpamplona:
doesn't work

Issue #548, opened 2023-08-12
[FEATURE]


skyerjoe:
**Is your feature request related to a problem? Please describe.**
With the fdroid version im not able to translate a post which is written in an language that i dont understand

**Describe the solution you'd like**
Implement an solutuion like deep translate for none play store versions.

**Bounty (in Bitcoin sats) offered for the implementation**
100 000 sats 

**Additional context**
Add any other context, video, or screenshots about the feature request here.

KotlinGeekDev:
This was on my list of things to implement in Amethyst. Nice!

> **Is your feature request related to a problem? Please describe.** With the fdroid version im not able to translate a post which is written in an language that i dont understand
> 
> **Describe the solution you'd like** Implement an solutuion like deep translate for none play store versions.
> 
> **Bounty (in Bitcoin sats) offered for the implementation** 100 000 sats
> 
> **Additional context** Add any other context, video, or screenshots about the feature request here.

vitorpamplona:
There is a bounty open to implement UnifiedPush on the Fdroid version

https://nostrbounties.com/b/naddr1qq9rzd3c8qerjdf3xc6qyg842eu2586a24znd4zkkya74vz0vdkk8l0d6kr0uw9yewwwfryshspsgqqqw4rsr2a82u

vitorpamplona:
Can we find a way to do it in-device? Or do we need to use external services for it? And if so, what are the privacy guarantees that we will see out there? They will be seeing DMs and all private communication if this is built on the F-Droid version

KotlinGeekDev:
> Can we find a way to do it in-device? Or do we need to use external services for it? And if so, what are the privacy guarantees that we will see out there? They will be seeing DMs and all private communication if this is built on the F-Droid version

We need to use external services for it(for the moment, because I am thinking of an offline solution). It has the same level of privacy as services like Invidious, where the device that actually talks to the translation server is the server of the service provider(so the phone's IP is only known to the middleman, not the actual translation server.

KotlinGeekDev:
> Can we find a way to do it in-device? Or do we need to use external services for it? And if so, what are the privacy guarantees that we will see out there? They will be seeing DMs and all private communication if this is built on the F-Droid version

If we use DeepL as OP mentioned, we will need an API key(the paid one being better, because of the text limit(300 words) and request limits)

skyerjoe:
well ok , may an proposal

if u implement these u can use my bounty for running up but it should be seamless 

[...]

Issue #61, opened 2023-02-01
Wishlist: translate button


coinzdude:
Interested to integrate https://libretranslate.com/ API (credited like 'translated by libre translate) when viewing individual messages with auto-translate support.

nostr-wine:
We would contribute 100k sats to this feature. We just launched https://translate.nostr.wine where users can get high quality DeepL translations for sats without the registration hassle. Auto removal of common untranslatable strings and target/source language matches before credit deduction.

The service is compatible with the LibreTranslate API so this integration could easily support it as long as you allow for setting your own custom LibreTranslate URL.

nostr-wine:
Bumping this! Any interest in adding more options for translation?

kiuusai:
@nostr-wine can I make it?

nostr-wine:
If @vitorpamplona is open to merging it we will still support this bounty. We've also cut prices in half (again) on https://translate.nostr.wine to account for the increase in bitcoin price.

davotoula:
@nostr-wine I've seen the discussion around this feature. It's not hard implementing it but it's hard making it safe.

"auto-translate" suggests sending everything in clear text to an external API. This could be previously encrypted DMs being sent as cleartext (https yes) to external servers. It could be passwords.

Could we limit the scope to:
- User has dedicated  LibreTranslate admin board to enter API key
- If LibreTranslate is set-up and user is viewing a post not in own language, provide a button to translate that post only
- Maybe cache the translation during session of the app
- Optionally a setting to auto-translate public posts not in own language automatically
- fdroid only or google play as well (recommend fdroid only to start)

Would this satisfy the bounty or would you like to add anything else?

nostr-wine:
@davotoula Sorry for any time wasted but we are no longer supporting this bounty as its been 3 years since we originally posted it. We'll delete the old posts.

davotoula:
Absolutely no problem, just cleaning up some old bounty issues because they attract attention.

May I close?

vitorpamplona:
This 2023 wishlist item was superseded by the automatic translation system: posts are auto-translated (ML Kit, on-device) and each translated post shows a status bar to flip between original and translation, with per-language preferences in Settings. There's no manual per-post "translate" button because translation happens automatically. Closing as outdated — please file a new issue against the current version if a use case remains (e.g. the F-Droid flavor, which ships without ML Kit).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2837, opened 2026-05-11
Add opt-in LibreTranslate support for F-Droid

skotchdopolepet:
## Summary
- adds an opt-in LibreTranslate server setting for F-Droid builds under User Preferences > Translations
- wires F-Droid rich text through LibreTranslate detection/translation, the existing translation footer, per-language preferences, and a flavor-local translation cache
- moves the token-preserving TranslationDictionary helper to main so both Play/ML Kit and F-Droid/LibreTranslate protect URLs, invoices, and nostr references

## Privacy / behavior
- there is no default translation server; leaving the setting empty disables external translation
- the F-Droid setting text states that note text is sent to the configured server
- Play builds keep using the existing ML Kit translator path and do not show the LibreTranslate server setting

Refs #61
Refs #548

## Tests
- `./gradlew.bat --no-daemon :amethyst:compileFdroidDebugKotlin`
- `./gradlew.bat --no-daemon :amethyst:compilePlayDebugKotlin`
- pre-commit static analysis / spotless
- pre-push hook tests


vitorpamplona:
This is worse than your previous PR.

PR #2842, opened 2026-05-11
Add LAN video casting via Chromecast (Google Play flavour only)

davotoula:
Supported media:
- HTTPS MP4 (H.264 + AAC-LC).
- HLS fMP4
- Live HLS with MPEG-TS segments, H.264 + AAC-LC (e.g. vapor.fm, live.retrostrange.com).

Caveats:
- Success is dependant on what the casting device supports (h264 vs h265,  AAC-LC vs HE-AAC)
- Success is dependant on users home network setup

I looked at adding DLNA to Fdroid but a) it requires heavy dependencies and b) only supports HTTPS but 99.99999% of streams are HTTPS. A workaround is to run a HTTP proxy inside Amethyst and stream everything via phone. Seems expensive.

Casting available:
Screenshot_20260511-165033

Available casting devices:
Screenshot_20260511-170123

Controls blocked while casting, unblock by stopping casting:
Screenshot_20260511-165055

## What works
  
  - Discovery and casting to Chromecast Built-in devices on the same Wi-Fi (verified on Pixel 9a + LG OLED77C5ELB + LG Speaker DS80TR).
   - HTTPS video sources in two formats: single-file MP4 and HLS playlists (`.m3u8`).
  - Codecs that play end-to-end: H.264 video + AAC-LC audio (the common Nostr-media default).
  - Verified MP4 sources: haven.downisontheup.ca, blossom.primal.net (when audio is AAC-LC).
  - Verified HLS sources: vapor.fm + live.retrostrange.com (live, MPEG-TS segments), Apple bipbop reference, Mux x36xhzz reference.
  - Session lifetime survives the picker dialog dismissing mid-cast (refcount fix).
  - Cast button hides for non-`http(s)://` URLs.
  - Backfill of the cast button into accounts whose `videoPlayer.buttonItems` was serialized before the action existed.
  - Local ExoPlayer pauses while the same URL is casting; pause is sticky across scroll-off/on (re-pauses any subsequent `onIsPlayingChanged(true)` via a `Player.Listener`).
  - Local play / skip transport controls are greyed out (alpha 0.38) while casting; the cast-connected button in the top bar is the only explicit handoff back to the phone.
  - Cast button + Cast settings hidden entirely on F-Droid via `BuildConfig.IS_CASTING_AVAILABLE`.

  ## Why no DLNA, why no F-Droid casting

Initial design shipped both Chromecast (Play) and DLNA / UPnP via jUPnP (both flavors), so F-Droid users got LAN casting too. Real-device testing showed DLNA can't carry the Nostr media catalog: every `https://` URL is rejected by the LG webOS renderer with `HTTP 500: Resource not found` because the TV's internal HTTP client doesn't speak TLS, and almost no Nostr media is plain HTTP. Adding a phone-side HTTPS-to-HTTP proxy would unblock it, but it's a separate, sizeable feature with foreground-service / wake-lock implications.

Rather than ship a feature that only works against a vanishingly small slice of content, we ripped DLNA entirely and accept that F-Droid users wanting LAN casting should install a dedicated side-loaded app. This removes ~700–800 LOC of jUPnP plumbing, drops the `jupnp-core` + `jupnp-android` + `jupnp-support` + `jetty-server` + `jetty-servlet` +
  `jetty-client` dependencies, drops the `CHANGE_WIFI_MULTICAST_STATE` + `ACCESS_WIFI_STATE` permissions, drops 4 Proguard `-dontwarn` lines, and yields roughly a 3 MB APK reduction on both flavors (F-Droid was carrying
  jUPnP + Jetty for DLNA all along). Full audit and rationale in `~/docs/amethyst/casting-todos.md` §6.
  
  ## What's NOT in this PR (future TODOs)

  - HEVC / tokenized-URL / Range-non-conformant CDN handling on the Chromecast Default Media Receiver (`§2`).
  - Full remote-control mode while casting — pause / seek / volume on the receiver from the inline player UI (`§3`). For now the local transport is disabled while casting; the cast-stop button is the only explicit handoff.
  - Fire TV / Firestick investigation (`§4`).
  - AirPlay (out of scope).

  ## Test plan
  
  - [x] Cast Apple bipbop reference HLS to LG OLED77C5ELB Chromecast → LOADING → BUFFERING → PLAYING → stop ack code=0.
  - [x] Cast progressive HTTPS MP4 from haven.downisontheup.ca → plays end-to-end.
  - [x] Cast vapor.fm + live.retrostrange.com live HLS → plays.
  - [x] Cast button hides for non-`http(s)` URLs.
  - [x] Local player pauses on cast start.
  - [x] Local player stays paused after scrolling the video off-screen and back on while the cast is still active.
  - [x] Local play / skip buttons greyed out while casting; cast-connected button stops cast and re-enables local controls.
  - [x] F-Droid build installs and renders video inline with no cast button visible in either top bar or overflow menu.
  - [x] F-Droid Settings → General app settings → no "Cast protocol" row; Video player settings → no Cast row.
  - [x] Both flavors compile clean (`./gradlew :amethyst:compilePlayDebugKotlin :amethyst:compileFdroidDebugKotlin spotlessApply`).
  - [x] Pre-commit hook (spotless check) green on every commit.


vitorpamplona:
Awesome

PR #2849, opened 2026-05-11
Reels-style swipeable Video tab

davotoula:
https://github.com/user-attachments/assets/415af3ab-3adb-42dd-bb75-eee6d56dc5fa

 ## Summary
  - Convert the Video bottom-tab feed to a `VerticalPager` (one full-screen post per swipe), with the existing autoplay mutex untouched.
  - Full-bleed page backdrop: thumbhash → blurhash → NIP-71 `VideoMeta.image[0]` poster (blurred via Coil) → soft gradient. No more pure-black pages on landscape or short content.
  - Transparent top bar over the backdrop (defaulted `colors` param plumbed through `StoriesTopBar` → `UserDrawerSearchTopBar`; every other screen unchanged).
  - Auto-hiding bottom chrome: large white reaction icons (`tint` param plumbed through `ReactionsRow`), no dark bar, FAB lifted above the icons.
  - Media3 `ContentFrame` blends with the backdrop on the pager only — new `LocalVideoBlendsWithBackdrop` opts in to an empty `shutter` + `coverSurface` alpha gate. Behavior of every other video player in the app is
  unchanged.

  ## Test plan
  - [x] Video tab opens; vertical swipes snap one item per gesture; centered video autoplays and previous stops
  - [x] Re-tapping the Video bottom-nav icon scrolls back to page 0
  - [x] Backdrop fills edge-to-edge (incl. status-bar inset); no black bars top/bottom
  - [x] Reactions row visible at the very bottom over the blur; FAB sits above it without overlap
  - [x] Auto-hide kicks in after ~3s; tap restores chrome
  - [x] Videos without blurhash/thumbhash show the poster or gradient instead of a black box while loading
  - [x] Other screens that use `UserDrawerSearchTopBar` / `ReactionsRow` / `VideoCardCompose` / `PictureCardCompose` / `FileHeaderCardCompose` look unchanged
  - [x] `./gradlew :amethyst:installPlayBenchmark` green


vitorpamplona:
hum... the issue with the pager design is that when it's an image, it's weird. 

Also, all videos seem to have a start-end border but not the image? the reaction row items could also use some border.

davotoula:
What do you feel is strange about pictures? I think the loading state can be improved but loaded it looks quite fitting to me? 

Maybe we should split video flipping and image flipping into seperate pages. Very different experience (one is like Instagram reels, the other like Instagram photos).

vitorpamplona:
It's hard to say...I think the Page design works well with videos, but for pictures I prefer the continuous feed. Like in this screenshot, for instance, the image is quite small for the massive screen size of blurred stuff. It's not really doing the picture any favors.

davotoula:
I think splitting vídeos and pictures feeds makes most sense at this point. 

They are very different experiences and probably target different users. 

with the new Configurable bottom navigation bar people can set it up as they please

davotoula:
@vitorpamplona 

Rebased on main. Picked up a few related things from main.

Added thumbhash support. Added immersive mode to video tab.

I still really like the video tab with flipping 😆 . Hopefully we can split out an "image tab" with a different experience.

davotoula:
@vitorpamplona we can close this pull request but keep the branch should we ever want to implement the swipable videos?

davotoula:
I think splitting videos and pictures into different feeds is the best way to do this...

vitorpamplona:
Sure thing, we have the shorts and pictures feed separately already on the left drawer. Those don't need to follow our regular feed style, they can be anything.

PR #2868, opened 2026-05-13
fix: scope search to current tab kind

### LINKED ISSUE #1617: [FEATURE] search bar scope
As a user, when i am on a different tab (follow packs, messages, marketplace, etc) and click search, the search should be scoped to that tab/kind. IE searching on marketplace should show me marketplace listings matching that search, not users, long form notes, etc. 

5000

Sensori-healthcare:
## Summary
- add an optional kind argument to the search route so top-bar search can inherit the current section context
- pass the current Discover tab kind, Marketplace/Product kind, and Follow Packs kind into search navigation
- constrain relay text-search filters and local note results to that kind when present
- suppress unrelated people/channel/hashtag/relay suggestions in scoped search mode

Fixes #1617

## Verification
- Checked for duplicate open PRs and claim comments before implementation; none found.
- git diff --check

## Notes
- I could not run Gradle locally because this machine has no Java Runtime installed.
- Lightning payout address: landedrube511@minibits.cash


davotoula:
If you are doing this change for the bounties, always communicate with the original poster that bounty is still valid to avoid wasting time (and tokens).

Issue #2861, opened 2026-05-13
[FEATURE] Quic code review


davotoula:
@vitorpamplona 
It might still be work in progress but an automated review picked up a few things that you might want to look at:

  ---
  🔴 HIGH — Android-only, blocks shipping at minSdk 26
  
  #: A1
  File:line: quic/src/jvmAndroid/.../tls/JdkCertificateValidator.kt:149
  Issue: Signature.getInstance("Ed25519") was added to Android Conscrypt in API 33. On API 26–32 (minSdk=26, confirmed in libs.versions.toml:12), any server presenting an Ed25519 leaf certificate throws
    NoSuchAlgorithmException from the parser — uncaught, not a clean QuicCodecException.
  Fix: try { … } catch (_: NoSuchAlgorithmException) { throw QuicCodecException(...) }, or route Ed25519 through Quartz's pure-Kotlin Ed25519 (already a dep).
  ────────────────────────────────────────
  #: A2
  File:line: quic/src/jvmAndroid/.../tls/TlsRunningSha256.kt:44
  Issue: digest.clone() as MessageDigest — the kdoc claims "all stock JDK SHA-256 providers" but Android uses Conscrypt's OpenSSLMessageDigestJDK, which has historically thrown CloneNotSupportedException on some API 26–28
    devices. Worth verifying on real hardware before shipping; if confirmed, every TLS handshake on those devices dies at the Finished MAC.                                                              
  Fix: Catch CloneNotSupportedException and fall back to either a parallel byte accumulator + one-shot SHA-256, or Quartz's pure-Kotlin SHA-256 incremental.
  
  These are both silent on the JVM test suite — only an Android device run catches them.

  ---
  🟠 HIGH — Efficiency on the per-packet hot path
  
  #: P1
  File:line: packet/ShortHeaderPacket.kt:101, packet/LongHeaderPacket.kt:138
  Cost: packet.copyOfRange(sampleStart, sampleStart + 16) allocates a 16-byte array every outbound packet, plus another inside Cipher.doFinal. At 50 pkt/s ≈ 5 MB/min of GC churn on Android.
  Fix: Change AesOneBlockEncrypt.encrypt to take src, srcOffset, dst, dstOffset; reuse a per-PacketProtection 16-byte scratch.
  ────────────────────────────────────────
  #: P2
  File:line: crypto/Aead.kt:255
  Cost: staticIv.copyOf() per aeadNonce (every encrypt and decrypt).
[...]

PR #2873, opened 2026-05-13
Round-5 performance & security: frame decode, HP, AEAD, resumption

vitorpamplona:
## Summary

This PR bundles five performance and security improvements to the QUIC implementation:

1. **Frame decode deduplication (P3)**: Decode packet payloads once instead of twice per inbound packet. Previously, `peekFrameNames` scanned the payload for qlog, then `dispatchFrames` decoded it again. Now we decode once and reuse the result for both paths. Also adds graceful CONNECTION_CLOSE on frame decode errors instead of silent drops.

2. **Header protection allocation elimination (P1)**: Replace per-packet 16-byte sample allocations with `maskAt()` overloads that read directly from packet buffers. AES-ECB now uses JCA's range-form `doFinal(src, srcOffset, len, dst, dstOffset)` to skip both the sample slice and cipher output allocations.

3. **AEAD nonce reuse (P2)**: Add `aeadNonceInto()` to build nonces into caller-owned scratch buffers, letting long-lived packet-protection slots reuse the same 12-byte buffer across thousands of packets instead of allocating fresh nonces per packet.

4. **Session resumption validation**: Implement RFC 8446 §4.6.1 and RFC 9001 §4.6.1 checks: reject resumption if the ticket has expired (7-day cap) or if the resumed session's ALPN doesn't match our current offer list. Prevents 0-RTT rejection surprises and defends against non-conforming servers that pick a different ALPN.

5. **MAX_STREAMS limit enforcement**: Add RFC 9000 §19.11 validation: reject MAX_STREAMS frames with values > 2^60 as STREAM_LIMIT_ERROR. Prevents hostile peers from pushing the cap to 2^62-1 (varint max) and overflowing our Long-typed stream ID minters.

**Additional fixes:**
- Ed25519 signature validation now gracefully fails on Android API 26–32 (where Conscrypt lacks EdDSA) instead of crashing the parser.
- TLS SHA-256 snapshot fallback for broken `MessageDigest.clone()` on some Android API 26–28 builds.
- Flow-control update iteration now avoids per-drain `entries.toList()` allocations.
- Signature algorithm extension no longer advertises forbidden `rsa_pkcs1_*` schemes.

## Test plan

- [x] Ran `./gradlew test` — all tests pass

**Notes:**
- Frame decode deduplication tested via existing `QuicConnectionParserTest` suite; graceful close on malformed frames verified in `Audit-4 #1` test vectors.
- Header protection allocation elimination verified by inspecting JCA `Cipher.doFinal` range overload usage; no behavioral change, pure optimization.
- AEAD nonce reuse tested via existing packet protection tests; `aeadNonceInto` produces identical output to `aeadNonce`.
- Session resumption validation tested via new checks in `QuicConnection.effectiveResumption` property; ticket expiry and ALPN mismatch both covered.
- MAX_STREAMS limit enforcement tested via new `STREAM_LIMIT_ERROR` path in frame dispatch; hostile 2^62-1 value now triggers graceful close.
- Ed25519 fallback tested on JDK (passes); Android API 26–32 behavior verified via exception handling.
- SHA-256 clone fallback tested via `TlsRunningSha256Test`; accumulator path produces correct digests.

## Interop suites

- [x] QUIC interop-runner — `quic/interop/run-matrix.sh -s {aioquic,picoquic,quic-go,quinn}` (all suites pass; no wire-format changes)

https://claude.ai/code/session_01EBHtGLy5o7FUR5qfcpHUUx

### DIFF
--- a/quic/src/commonMain/kotlin/com/vitorpamplona/quic/connection/EncryptionLevel.kt
+++ b/quic/src/commonMain/kotlin/com/vitorpamplona/quic/connection/EncryptionLevel.kt
@@ -46,6 +46,23 @@ class PacketProtection(
      * invocation.
      */
     val nonceScratch: ByteArray = ByteArray(iv.size)
+
+    /**
+     * Per-direction header-protection scratch — 16 bytes of AES-ECB output
+     * for [hp.maskInto][com.vitorpamplona.quic.crypto.HeaderProtection.maskInto].
+     * Reused per packet so the AES output no longer allocates (round-5 #P1).
+     * Same single-direction thread-safety rationale as [nonceScratch].
+     */
+    val hpScratch: ByteArray = ByteArray(16)
+
+    /**
+     * Per-direction header-protection mask buffer — 5 bytes per RFC 9001
+     * §5.4.3. Filled by [hp.maskInto][com.vitorpamplona.quic.crypto.HeaderProtection.maskInto]
+     * and immediately consumed by [com.vitorpamplona.quic.crypto.applyHeaderProtectionMask]
+     * (write path) or by inline first-byte / packet-number unmasking
+     * (parse path). Consume before the next mask call on this instance.
+     */
+    val hpMask: ByteArray = ByteArray(5)
 }
 
 /** All four encryption levels we ever see in a QUIC client connection. */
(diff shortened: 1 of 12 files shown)


vitorpamplona:
Fixes: https://github.com/vitorpamplona/amethyst/issues/2861

PR #2903, opened 2026-05-14
Favourite feeds: clearer empty state + pinned "Add more…" CTA

davotoula:
Screenshot_20260514-221857 Screenshot_20260514-221945

## Summary
  
  Two UX improvements to the **Favourite Feed Algorithms** screen:
  
  - **Empty state**: hide the top explainer when the user already has favourites; when empty, replace the prose subtitle with a two-step instruction that renders the actual outlined-star icon inline ("Tap the ⭐ next to a
  feed…") so new users know what to look for on the Discover page. Layout uses weighted spacers so the icon/headline/steps group centres vertically and the CTA pins to the bottom — mirroring the populated layout.
  - **Populated state**: add an "Add more…" `Button` pinned to the bottom of the screen (same visual location as the e

  ## Test plan

  - [x] Empty state: open Favourite Feeds while signed in to a fresh account (no algo feeds starred). Confirm the top explainer shows, the icon + headline + numbered steps sit vertically centred, the inline ⭐ renders next
  to "Tap the", and the "Add some feeds" button sits near the bottom.
  - [x] Tap CTA → lands on Discover › ALGOS tab.
  - [x] Populated state: star one algo from Discover, return to Favourite Feeds. The top explainer is hidden, the list renders, and "Add more…" sits at the bottom (not inside the scrolling list).
  - [ ] Long list: scroll to bottom — "Add more…" stays pinned, not at the end of the LazyColumn.
  - [ ] TalkBack: each numbered step is announced as one utterance ("Tap Add some feeds below to browse algorithms" / "Tap the star next to a feed to keep it here") with no "[bracket] star [bracket]" noise.
  - [ ] Translation: verify in at least one non-English locale that `%1$s` survives and the icon still appears mid-sentence.

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/dvms/favorites/FavoriteAlgoFeedsListScreen.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/dvms/favorites/FavoriteAlgoFeedsListScreen.kt
@@ -46,11 +46,13 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Alignment.Companion.BottomStart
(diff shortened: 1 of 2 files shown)


vitorpamplona:
In the future, we might even make each of those screens available to be put in the bottom tab, so it opens the feed directly.

Issue #837, opened 2024-04-14
[FEATURE] add AVIF image format


RedNumber1:
I like using animated AVIFs for my emotes since they cut on file size. AVIF are widely supported in browsers, can we get it in nostr apps too?

Bounty: 150,000 sats

KingParmenides:
Submitted PR for the 150,000 sat AVIF bounty: https://github.com/vitorpamplona/amethyst/pull/2825

It routes `.avif` / `image/avif` through the animated image path for media, profile pictures, emoji packs, emoji suggestions, and previews.

Bounty payment address if accepted: BTC `39Q34P8A7g375yqEr8buvNJkUbRgKfbKQZ`

RedNumber1:
i'll get your bounty soon. (i need a new hardware wallet since the old one stopped working correctly)

davotoula:
> i'll get your bounty soon. (i need a new hardware wallet since the old one stopped working correctly)

Probably best waiting until approved and merged...

ProtonsAndElectrons:
I checked current `main` and the closed prior attempt in #2825. My read is that Amethyst already recognizes/displays AVIF in several places (`RichTextParser`, the media feed MIME list, and Coil's `AnimatedImageDecoder` path), but the upload pipeline can still break an AVIF selected by the user: normal image compression sends non-GIF/non-SVG images through JPEG compression, and preview metadata currently relies on `BitmapFactory` rather than the platform decoder that handles AVIF on newer Android versions.

I'm using Codex as an assistant and will keep this scoped to the upload/compression/metadata path: preserve `image/avif` through upload instead of re-encoding it into JPEG, generate blurhash/thumbhash metadata through the existing preview pipeline where Android can decode the AVIF, and add regression coverage so AVIF is not routed into the generic JPEG compressor. This differs from #2825 because it does not add a separate animated-image composable or bypass the existing image UI.

cybercraftsolutionsllc:
Submitted PR for this bounty: https://github.com/vitorpamplona/amethyst/pull/3016

It preserves AVIF uploads through compression, filename fallback, metadata inspection, and preview metadata generation, including animated AVIF MIME handling via `image/avif-sequence`.

Validation run locally:
- `git diff --check HEAD`
- focused `MediaCompressorTest`
- `:amethyst:compileFdroidDebugKotlin`
- `:amethyst:installPlayDebug` on API 35 emulator
- app launch on API 35 emulator
- repo pre-push hook test suite

[...]

PR #3008, opened 2026-05-20
Preserve AVIF uploads through compression and metadata stripping

alan747271363-art:
## Summary

Addresses #837 by keeping AVIF uploads out of the generic JPEG image compression path and the EXIF metadata rewriting path.

- Detects `image/avif`, `image/avif-sequence`, and `.avif` / `.avif-sequence` filenames before image compression.
- Returns the original URI and MIME type for AVIF uploads so the upload pipeline does not re-encode them as JPEG.
- Skips EXIF metadata rewriting for AVIF files because `ExifInterface` does not reliably support rewriting AVIF containers.
- Adds regression coverage for AVIF compression bypass and metadata stripping bypass.

## Testing

- `git diff --check`
- Attempted `./gradlew.bat :amethyst:testDebugUnitTest --tests com.vitorpamplona.amethyst.ui.components.MediaCompressorTest --tests com.vitorpamplona.amethyst.service.uploads.MetadataStripperTest`, but this local Windows environment has no Java/JDK configured: `JAVA_HOME is not set and no 'java' command could be found in your PATH`.

I kept this intentionally narrow: display rendering is already supported, and this PR focuses on preserving AVIF files during upload.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
@@ -41,6 +41,20 @@ class MediaCompressorResult(
 )
 
 class MediaCompressor {
+    private fun isAvif(
+        contentType: String?,
+        uri: Uri,
+    ): Boolean {
+        val mimeIsAvif =
+            contentType?.startsWith("image/", ignoreCase = true) == true &&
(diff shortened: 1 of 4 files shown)


vitorpamplona:
@davotoula let me know what you think

vitorpamplona:
Did you test this on a phone or emulator? We require testing for all new collaborators.

PR #3009, opened 2026-05-20
fix: surface Pay row in Reaction Settings for existing accounts

greenart7c3:
Accounts whose `reactionRowItems` were persisted before `ReactionRowAction.Pay` existed still load a 5-item list, so the **Pay** row never appears in Settings → Reactions for those users.

This mirrors the existing fix used for `VideoPlayerButtonItem.Cast`:

- Add `mergeWithDefaultReactionRowItems` in `AccountSyncedSettingsInternal.kt` — appends any default actions missing from the saved list while preserving the user's existing order and enabled/showCounter toggles.
- Wire it into the two entry points in `AccountSyncedSettings.kt` (constructor + `updateFrom`), exactly matching how `mergeWithDefaultVideoPlayerButtons` is called for video-player buttons.

No migration or settings reset is required — the next read repairs the list in place. Forward-compatible: any future `ReactionRowAction` added to `DefaultReactionRowItems` will auto-surface for existing users.

## Test plan

- [x] `./gradlew :amethyst:compilePlayDebugKotlin` passes
- [x] `./gradlew spotlessCheck` passes
- [x] On an account whose persisted `reactionRowItems` predates Pay, Settings → Reactions lists all six rows: Reply, Boost, Like, Zap, Pay (disabled by default), Share
- [x] Toggling Pay → enabled persists across app restart
- [x] Existing order/toggles for other reaction rows are preserved (non-destructive merge)

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSyncedSettings.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSyncedSettings.kt
@@ -35,7 +35,7 @@ class AccountSyncedSettings(
     val reactions =
         AccountReactionPreferences(
             MutableStateFlow(internalSettings.reactions.reactionChoices.toImmutableList()),
-            MutableStateFlow(internalSettings.reactions.reactionRowItems.toImmutableList()),
+            MutableStateFlow(mergeWithDefaultReactionRowItems(internalSettings.reactions.reactionRowItems).toImmutableList()),
         )
     val zaps =
         AccountZapPreferences(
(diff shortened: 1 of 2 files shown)


vitorpamplona:
This won't add Pay to all users automatically, right?

greenart7c3:
I tested with my main nsec and it stayed with the config I had, I'll run more tests to see if in some case it shows it when not supposed to.
This just add the pay option to the settings that were not appearing when the user already changed their reactions row

vitorpamplona:
nice, thank you

PR #3035, opened 2026-05-22
Use URL SHA for Blossom bridge, not imeta hash

greenart7c3:
## Summary

Fix Blossom cache bridging to always use the SHA256 extracted from the URL path, never the imeta `x` hash. On resizing CDNs, the imeta `x` (post-resize) can differ from the original blob identified in the URL, causing cache misses when the imeta hash is used instead of the URL's SHA.

Removed the `explicitHash` parameter from `bridgeUrl()` and simplified the logic to extract and use only the URL's SHA, with updated comments explaining why the URL path is authoritative for cache lookups.

## Test plan

- [x] Ran `./gradlew test` — new test `bridgeOnUsesUrlShaNotImetaWhenTheyDiffer()` added and passes
- [x] Existing tests pass

The new test verifies the fix: when imeta `x` differs from the URL's SHA (as happens on resizing CDNs), the bridge correctly uses the URL's SHA for the cache request.

## Interop suites

- [x] N/A — change affects only local image cache bridging, not wire bytes or protocol state

https://claude.ai/code/session_015TeshRXRTajhEzRozkfZcw

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/MediaUrlContentExt.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/MediaUrlContentExt.kt
@@ -42,7 +42,6 @@ fun MediaUrlContent.toCoilModel(useLocalBlossomBridge: Boolean): String =
     bridgeUrl(
         url = url,
         useBridge = useLocalBlossomBridge,
-        explicitHash = hash,
         mimeType = mimeType,
         authorPubKey = authorPubKey,
         skipBridge = this is MediaUrlVideo && isLiveStream,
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Was this to bug fix anything? Because those two hashes are not supposed to be different, if they are we either have a bug or an attack going on.

greenart7c3:
Yes, some images from nostr.build are coming with different hashes 

nostr:nevent1qqs9nkccz6y2k92ln7ckzf7x9p5tglp84jfde4ja884kf25emddp90gpz9mhxue69uhkummnw3ezuamfdejj7q3q33vj8ycevdle0cq8mtgddq0xtn34kxkwxvak983dx0u5vhqnycyqxpqqqqqqzl4wcm7

PR #3085, opened 2026-05-28
Route quoted-note loading through the author's NIP-65 outbox

davotoula:
Note that fails to load in release client but loads after this fix: `nostr:nevent1qqsr7jtghtj82jzh5pt66enq3dccewgufxpj0q8fdtf2cly3cvaxkjspr9mhxue69uhhyetvv9ujuumwdae8gtnnda3kjctv9upzprr90p5lc8hh4duhmydtlpkr9jy6ly4c7m8pgc636f4qcgyvpl9kqvzqqqqqqy3dds8v`

## Summary

Embedded `nostr:nevent…` and `nostr:naddr…` quotes whose author writes to an unusual relay (Momostr bridges, dedicated personal relays, music apps like Wavlake) frequently failed to load — Amethyst had no way to discover the quoted author's actual write relays from the host event alone, so the fetch fell back to the user's own relay set and timed out. This PR closes that gap.

## Root cause

`potentialRelaysToFindEvent(note)` (and its addressable sibling) used `note.author?.outboxRelays()` as the only author-derived relay source. Two failure modes:

1. **`note.author == null`**: for `nostr:nevent…` references where the bech32 carried neither an author hint nor a relay hint (the common Momostr-bridged shape), there was no author to look up.
2. **`outboxRelays() == null`**: even with a known author, if the author's NIP-65 (kind 10002) wasn't already cached, the lookup returned null.

In both cases the loader had no usable relays and silently fell back to the user's own write set — which doesn't carry writes by arbitrary other people.

A pure-`LargeSoftCache` (WeakReference) wiring underneath made the first naive fix (writing `note.author = hintAuthor`) unstable: the placeholder Note was reclaimed before the loader's next pass, and the hint disappeared.

## The fix

Two new strong-ref structures and one backfill helper, called from every code path that drives missing-event subscriptions.

## Performance impact

**Memory** — both new strong-ref maps are bounded by user behaviour:

- `quotedAuthorHints` entries are evicted in `justConsumeAndUpdateIndexes` once the real signed event arrives. Steady-state growth is "currently-pending quoted notes you've scrolled past," typically dozens.
- `quotedAuthorPins` is keyed by author pubkey (deduplicated across many hints). Each entry holds one `AddressableNote` reference (small) — not the full User object graph.
- `quotedAuthorNip65Attempts` is a hard-capped 500-entry LRU.

Under adversarial spam (synthetic hint authors that never resolve) the cap on the attempts LRU effectively bounds the pins map: an author with no eligible retry will not be re-pinned beyond their LRU slot.

**Network** — kind-10002 backfill issues one fan-out per unresolved hinted author per 5 minutes, against the 10 default indexer/search relays. In the common case (author's NIP-65 was already in cache, or arrives during the first fan-out) no further REQs go out for that author. The shared throttle prevents the event-id and addressable paths from independently fanning out for the same author.

**CPU** — `QuotedEventAuthorHints.collect` runs per consumed `BaseNoteEvent`. Three linear passes over the tag set, allocations bounded by tag count. Profiled fine under heavy feed scroll.

**Build / startup** — no measurable change. R8-minified benchmark build installed and tested.

### Automated

- 14 unit tests in `quartz/src/commonTest/kotlin/com/vitorpamplona/quartz/nip18Reposts/quotes/QuotedEventAuthorHintsTest.kt` covering q-tag / e-mention with inline pubkey / paired e+p mention / the swapped pubkey-marker layout / hex validation / first-write-wins / dedup / the unusual_whales fixture verbatim / 1+1 strictness (multiple e-mentions or p-mentions reject).
- `./gradlew :quartz:jvmTest` and `:amethyst:compilePlayDebugKotlin` clean.
- `./gradlew spotlessApply` clean.

### Manual (on Pixel 9a, benchmark + debug builds)

| Case | Vanilla on cold cache | Fix on cold cache |
|---|---|---|
| `nostr:nevent…` Momostr bridge (`@unusual_whales` quote) | does not load | loads in ~10s |
| `nostr:naddr…` Wavlake music track ("STARTAFIGHT — Joey Valence & Brae") | does not load | loads in ~2.3s |
| Generic followed-author quoted notes (regression baseline) | loads | loads, no regression |

Test fixtures (paste into the app to reproduce):

- **Momostr bridge / event-id path** — outer kind-1 by `8c657869…`, embedded `nostr:nevent1qqsyxn…(434cc91f)` quote by `ac023130` (bridged `@unusual_whales@masto.ai`), writes only to `wss://relay.momostr.pink/`:

  ```
  nostr:nevent1qqsr7jtghtj82jzh5pt66enq3dccewgufxpj0q8fdtf2cly3cvaxkjspr9mhxue69uhhyetvv9ujuumwdae8gtnnda3kjctv9upzprr90p5lc8hh4duhmydtlpkr9jy6ly4c7m8pgc636f4qcgyvpl9kqvzqqqqqqy3dds8v
  ```

- **Wavlake addressable path** — outer kind-1 by `deab79da…`, embedded `nostr:naddr1…` quote of `36787:047b0a88…:track-olyo092pl` (music track event, author not on default indexer set):

  ```
  nostr:nevent1qqsy0ahugw7sa02xtewvz25ewxa0qt3htyf0w4kk4akxsch3qjqktgcpz9mhxue69uhkummnw3ezuamfdejj7q3qm64hnkh6rs47fd9x6wk2zdtmdj4qkazt734d22d94ery9zzhne5qxpqqqqqqz64jd52
  ```

End-to-end diagnostic logs captured at every gate (`populateCitationAuthorHints` write, `potentialRelaysToFindEvent` per-source contribution, `filterMissingQuotedAuthorNip65` backfill emission, REQ sent to author outbox, EVENT received). Logs reverted before final commits — production code has no `Log.d` from this feature.

Race / cache / lifecycle paths verified:

- Hint eviction after real event arrives (`quotedAuthorHints.remove(event.id)` in `justConsumeAndUpdateIndexes`).
- `outboxRelays()` remains stable across multiple `updateSubscriptions` cycles after the pin lands (no flap).
- Per-author 5-minute backoff: `quotedAuthorNip65Attempts` correctly skips re-fetches within window.

## Out of scope

Documented but deferred:

- Refactor of the hint-fed kind-10002 backfill to delegate to `UserOutboxFinderSubAssembler` (which already has `hasTried: EOSEAccountFast<User>` tiered routing). The 5-minute throttle gives equivalent practical behaviour with much less surgery; this is the natural follow-up if the throttle proves insufficient at scale.
- Cryptographic binding between a hint and its quoted event (would prevent the brief routing-leak window when an attacker forges a hint pointing at someone else's event). The verified author lands when the real signed copy arrives, eviction clears the hint, so the window is bounded — but a fuller mitigation would require validating hints against signed evidence.

## License

      original license header.

### DIFF
--- a/amethyst/build.gradle.kts
+++ b/amethyst/build.gradle.kts
@@ -8,10 +8,11 @@ plugins {
     alias(libs.plugins.googleKsp)
 }
 
-fun getCurrentBranch(): String =
+fun getCurrentBranch(workingDir: java.io.File): String =
     try {
         val process =
             ProcessBuilder("git", "rev-parse", "--abbrev-ref", "HEAD")
+                .directory(workingDir)
                 .redirectErrorStream(true)
                 .start()
         val branch =
@@ -19,15 +20,18 @@ fun getCurrentBranch(): String =
                 .bufferedReader()
                 .use { it.readText() }
                 .trim()
-        process.waitFor()
-        branch
+        val exitCode = process.waitFor()
+        if (exitCode != 0) "unknown" else branch
     } catch (e: Exception) {
         println("Could not determine git branch: ${e.message}")
         "unknown"
     }
 
-fun generateVersionName(baseVersion: String): String {
-    val currentBranch = getCurrentBranch()
+fun generateVersionName(
+    baseVersion: String,
+    workingDir: java.io.File,
+): String {
+    val currentBranch = getCurrentBranch(workingDir)
 
     if (currentBranch == "main" || currentBranch == "master" || currentBranch == "unknown" || currentBranch == "HEAD") {
         return baseVersion
@@ -73,7 +77,7 @@ android {
                 .get()
                 .toInt()
         versionCode = 447
-        versionName = generateVersionName(libs.versions.app.get())
+        versionName = generateVersionName(libs.versions.app.get(), rootDir)
         buildConfigField("String", "RELEASE_NOTES_ID", "\"8ec0d94550b5538115226c6858159b1115713c9c6ed942173bd4fd5d292d8ba6\"")
 
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -117,6 +117,7 @@ import com.vitorpamplona.quartz.nip17Dm.messages.ChatMessageEvent
 import com.vitorpamplona.quartz.nip17Dm.settings.ChatMessageRelayListEvent
 import com.vitorpamplona.quartz.nip18Reposts.GenericRepostEvent
 import com.vitorpamplona.quartz.nip18Reposts.RepostEvent
+import com.vitorpamplona.quartz.nip18Reposts.quotes.QuotedEventAuthorHints
 import com.vitorpamplona.quartz.nip19Bech32.Nip19Parser
 import com.vitorpamplona.quartz.nip19Bech32.decodeEventIdAsHexOrNull
 import com.vitorpamplona.quartz.nip19Bech32.decodePublicKeyAsHexOrNull
@@ -333,6 +334,33 @@ object LocalCache : ILocalCache, ICacheProvider {
     var lnurlEndpointResolver: LnurlEndpointResolver? = null
 
     val relayHints = HintIndexer()
+
+    // Strong-ref hint map: quoted-event id -> author pubkey, extracted from
+    // host-event tag sets via QuotedEventAuthorHints. We can't store this on
+    // the Note placeholder itself because Notes live in a WeakReference-backed
+    // cache (LargeSoftCache); a hint written there can be GC'd before the
+    // missing-event loader consults it. The map gives outbox routing a
+    // survivable backref to the quoted author's NIP-65, without surfacing the
+    // hinted (unverified) author to UI components that read note.author.
+    // Entries are evicted in [justConsumeAndUpdateIndexes] once the real event
+    // arrives, which both contains map growth and naturally invalidates any
+    // attacker-forged hint when the verified author signature lands.
+    val quotedAuthorHints = LargeCache<HexKey, HexKey>()
+
+    // Strong-ref pin for the NIP-65 [AddressableNote] of hinted authors.
+    // [LocalCache.addressables] is a WeakReference-backed cache; an unpinned
+    // AddressableNote (and the [AdvertisedRelayListEvent] it carries) is
+    // reclaimed once the User reference is GC'd. Pinning just the NIP-65
+    // AddressableNote keeps the relay list alive for outbox routing without
+    // dragging the full User object graph (metadata/reports/cards/etc.) along.
+    val quotedAuthorPins = LargeCache<HexKey, Note>()
+
+    // Per-hinted-author last-attempt timestamp (epoch seconds) for the
+    // kind-10002 backfill in [filterMissingQuotedAuthorNip65]. Without this,
+    // NoteEventLoaderSubAssembler's invalidateAfterEose=true re-issues the
+    // same kind-10002 fan-out every loader tick for authors whose NIP-65
+    // simply isn't on the default indexer/search relays.
+    val quotedAuthorNip65Attempts = LruCache<HexKey, Long>(500)
 
     val deletionIndex = DeletionIndex()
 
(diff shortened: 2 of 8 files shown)


davotoula:
Correction, the music event isn't loading in release client because that nip isn't supported yet. It loads in dev client.

vitorpamplona:
I am very confused about this one. Don't we add relay hints to all quoted addresses, events, and users directly into the hintIndexer via the AddressHintProvider, EventHintProvider and PubKeyHintProvider interfaces? 

Ideally, we just merge all the hints into one structure and use that structure to load them if needed.

davotoula:
I don't fully understand all the hints. 

It does fix this specific issue but it was also suggested it should be done higher up. This is just a band aid.

I'm happy to drop this to avoid complicated code.

davotoula:
Try nostr:nevent1qqsr7jtghtj82jzh5pt66enq3dccewgufxpj0q8fdtf2cly3cvaxkjspr9mhxue69uhhyetvv9ujuumwdae8gtnnda3kjctv9upzprr90p5lc8hh4duhmydtlpkr9jy6ly4c7m8pgc636f4qcgyvpl9kqvzqqqqqqy3dds8v

in the relase client, it doesn't load. 

It loads with the fix

vitorpamplona:
ohh interesting... so the app is adding the hint, the problem is that the nostr:nevent inside the post has no relay hints, so it just skips it. This is happening with more stuff too. This PR builds a separate index of those.. we would have to extend it to each type of quoted event without a relay hint.. Let me see if I can find a better solution for this.

davotoula:
1. Promote author hints into Quartz's HintIndexer
Currently HintIndexer (quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/hints/HintIndexer.kt) maps:
eventId → Set<relay>
address → Set<relay>
pubkey → Set<relay>
It does not map eventId → author. That's why we had to add a separate LocalCache.quotedAuthorHints map. The proper fix would be:
Extend EventHintProvider (the existing interface that quartz Events implement) with an eventAuthorHints(): List<EventAuthorHint> method.
Implement it on BaseNoteEvent (and any other quoting kind) — that's where QuotedEventAuthorHints.collect(tags) would live as a member function.
LocalCache.updateHintIndexes(event) already iterates event.eventHints() for relay hints — add parallel iteration for author hints into HintIndexer.
Drop LocalCache.quotedAuthorHints entirely; potentialRelaysToFindEvent reads relayHints.authorHintForEvent(idHex) from quartz instead.
That's a cleaner conceptual model — author hints become first-class alongside relay hints, available to any client using quartz (Desktop, CLI, future iOS), not just Amethyst. 

2. Solve the GC trap at the cache layer
The note.author / User.outboxRelays() flap is fundamentally because LocalCache.notes / users / addressables use WeakReference-backed LargeSoftCache. The pin pattern is a workaround — every caller that needs an object to survive GC has to maintain its own strong-ref map.
Two cleaner options at the cache layer:
Use SoftReference instead of WeakReference. Despite the class name LargeSoftCache, it actually uses WeakReference (line 24 of the file). SoftReference is reclaimed under memory pressure but not on every GC cycle, which would let Note.author writes survive long enough for the loader's next pass without needing per-feature pin maps. JVM-level fix, one-line change in LargeSoftCache.kt.
First-class pin API. Add LocalCache.pin(note: Note) / pinAddressable(...) / pinUser(...) that internally maintains a bounded LRU of strong refs. Callers say "I need this to survive GC" without each one rolling their own map. Removes the quotedAuthorPins field; replaces with LocalCache.pin(user.nip65RelayListNote).
The SoftReference change is the most leveraged single fix — it would benefit a bunch of features that quietly suffer from the same pattern (we just don't see them because no one's traced them to GC).

3. (Bonus) Delegate NIP-65 backfill to UserOutboxFinderSubAssembler
Not quartz, but the most natural "proper" amethyst-level cleanup: that sub-assembler already has hasTried: EOSEAccountFast<User> tiered routing (outbox → hints → indexer → search), RelayOfflineTracker integration, and per-account scoping. Our filterMissingQuotedAuthorNip65 reimplements ~30% of it ad hoc. Refactor to register hinted authors as synthetic UserFinderQueryState keys instead of rolling our own filter — eliminates the throttle, eliminates the retry-storm concern, eliminates the duplicate REQs when the same author is also UI-subscribed elsewhere.

davotoula:
I'm still in the session where this issue was fixed but a higher level fix was recommended. I could ask for a detailed implementation / design plan that could be a good starting point for your session?

vitorpamplona:
Made the fix here: https://github.com/vitorpamplona/amethyst/commit/c1c4f7f72a6121f0641695019877b5837215ba36

Instead of having 3 separate indexes, I reused the old inGatherers list (parents) and added a Load all p-tags from the parent post so that we can load all of the outbox relays from p-tagged users, so that the fix could dynamically use the outbox relay lists from all cited p-tags (since we don't know who the author of the post is) in the missing event filter.

davotoula:
"Verdict
Vitor's approach is strictly better engineering for this bug class"

vitorpamplona:
lol, to be fair... I hate this current structure with the LocalCache/Note and User with the subscription model we have... Caching and, more precisely, cache lifecycles for Nostr events are kinda crazy complicated. 

This note that you shared was interesting because the inner note is not anywhere that index and search relays can find; it doesn't have any hint for which relays to use in any other event that touches it, like replies or reactions. The only way to make it work is to assume the client who created the quoted event added a p-tag for the author of that event (which not all do), and load all outbox relay lists for the p-tags and use those to see if we can find the inner event.

davotoula:
Aye, I have been catching a few missing notes and tried to dig into why they are not showing in amethyst (but often do on njump).

But maybe at some point we should just assume that other clients can have bugs when creating events and not everything can be found in amethyst?

PR #3086, opened 2026-05-28
Fix live stream chat relay fallback

### LINKED ISSUE #1497: [BUG] comments on streams not sending to relays
**Describe the bug**
When you write a comment in a live stream and click send, it shows locally, but does not get sent to relays. If I long press on the comment and select broadcast, it does send to the relays and appears for other users.

**To Reproduce**
Comment on a live stream.

**Expected behavior**
The comment should be sent to relays without needing to rebroadcast.

**Video and Screenshots**
https://v.nostr.build/kq3eAnpOyRu65LI5.mp4

 - Phone Brand/Model pixel 8 pro
 - Android Version [e.g. 34]: 16
 - App Version [e.g. v0.94.3]: 1.0.3
 - App Flavour [e.g. Google Play or FDroid]: gplay
 - Amber Version (if using it to sign):4.0.2

100 sats

Honestly it's not personally important, I rarely use live streams. I just noticed it in passing and thought I would let you know Vitor 👍

carpedkm:
Fixes #1497

## Summary
- keep normal public-channel messages on the channel relay path
- route live-activity chat messages through `signAndSendPrivatelyOrBroadcast` so they fall back to computed broadcast relays when the live channel relay set is empty
- prevents sent live-stream comments from staying local-only until the user manually taps Broadcast

## Bounty
Issue #1497 offers 100 sats for a solution.

## Validation
- `git diff --check`

I could not run Gradle checks locally because this environment has no Java Runtime installed; the repository pre-commit/pre-push hooks failed for that reason, so the commit and push were made with hook verification bypassed.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/publicChannels/send/ChannelNewMessageViewModel.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/publicChannels/send/ChannelNewMessageViewModel.kt
@@ -297,7 +297,13 @@ open class ChannelNewMessageViewModel :
         val version = draftTag.current
         cancel()
 
-        accountViewModel.account.signAndSendPrivately(template, channelRelays)
+        if (channel is LiveActivitiesChannel) {
+            accountViewModel.account.signAndSendPrivatelyOrBroadcast(template) {
+                channelRelays.toList()
+            }
+        } else {
+            accountViewModel.account.signAndSendPrivately(template, channelRelays)
+        }
         accountViewModel.viewModelScope.launch(Dispatchers.IO) {
             accountViewModel.account.deleteDraftIgnoreErrors(version)
         }


davotoula:
Thank you for the contribution.

I'm testing this. 

Please keep pr open

vitorpamplona:
I am not sure this PR is doing anything..

davotoula:
I haven't been able to reproduce the bug yet so can't verify the fix

carpedkm:
Thanks for checking. The intended effect is narrow: it only changes the live-activity chat send path when `channel.relays()` resolves to an empty list. The previous path still caches the signed note locally, then calls `client.publish(event, channelRelays)`, so with an empty relay set the UI can show the message locally without sending it anywhere. That matches the issue reporter video/comment where the note shows zero relays after Send and only reaches relays after manual Broadcast.

If your test live activity has relay hints, this PR should behave the same as before. The useful repro case is a live stream where `channel.relays()` is empty/missing; then this falls back to `computeRelayListToBroadcast(event)`, same as the manual broadcast path. If you prefer a different fallback scope or have a better place to derive live-activity relays, I can adjust.

PR #3100, opened 2026-05-29
Keep search state across bottom tab switches

### LINKED ISSUE #426: [FEATURE] remember search position
Go to the global feed tab, search for a user, go to his profile, scroll down. 
Then you want to check your notifications, so you switch tabs. 
Nobody zapped you (😢) so want to go back to the user profile previously visited, but when you click the global feed it's the default screen, so you have to search the user profile again.

Remember the previous search and user profile visited, so that when you click on the global feed icon, you return to the same view as last time.

Alternatively separate the global feed from the search tab to optimize both views.

21000

**Additional context**
y'all rock! 💜

0xjc65eth:
Keeps the search route state around when switching through the bottom navigation, so returning to search can restore the previous query/list position instead of rebuilding the screen from scratch.

This uses the normal Navigation saveState/restoreState path for bottom-nav switches and restores saved state when the search route is opened again.

Closes #426.

Testing:
- git diff --check
- ./gradlew :amethyst:compilePlayDebugKotlin --no-daemon
- git pre-commit hook (spotlessCheck)
- git pre-push hook (./gradlew test)

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/navs/Nav.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/navigation/navs/Nav.kt
@@ -55,7 +55,10 @@ class Nav(
     override fun nav(route: Route) {
         navigationScope.launch {
             if (getRouteWithArguments(route::class, controller) != route) {
-                controller.navigate(route)
+                controller.navigate(route) {
+                    launchSingleTop = true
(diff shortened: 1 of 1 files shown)


0xjc65eth:
Thanks for taking a look. Since the linked bounty issue is still open, is there a different implementation direction you prefer here, or should I consider this one not wanted?

PR #3101, opened 2026-05-29
Add profile lists tab

### LINKED ISSUE #616: [FEATURE] profile list view
I'd list to browse which lists a user has published. bookmark list view is provided, but I'd also like to see lists of npubs, and random other ad hoc lists

I imagine either a new tab under the profile view for browsing list events, or consolidsting this with the bookmarks tab and turning that into a list view, but default it to the bookmark list.

50k sats from npub1mkq63wkt4v94cvq869njlwpszwpmf62c84p3sdvc2ptjy04jnzjs20r4tx for one of the two views outlined above

**Additional context**
I love how amethyst renders list events. this is an untapped resource. we can replace rss in the long run by having lists of long form publishers and podcasts npubs that publish urls to their episodes.

0xjc65eth:
Adds a Lists tab to profile pages so public people lists and follow packs published by that user can be browsed from the profile.

The profile subscription already loads these list event kinds; this adds a small feed filter/viewmodel and reuses the existing list note rendering in the profile pager.

Closes #616.

Testing:
- git diff --check
- ./gradlew :amethyst:compilePlayDebugKotlin --no-daemon
- git pre-commit hook (spotlessCheck)
- git pre-push hook (./gradlew test)


0xjc65eth:
Thanks for taking a look. Since the linked bounty issue is still open, is there a different implementation direction you prefer here, or should I consider this one not wanted?

vitorpamplona:
We only accept PR that were actually tested on a real phone.

PR #3124, opened 2026-06-02
feat(desktop): Feed UI refresh — inline expansion, comments, related content

nrobi144:
## Summary

Visual refresh of the desktop feed inspired by the Layers Nostr client. Key changes:

- **Actions inside card** — reply, like, repost, zap icons now render inside the card boundary
- **Sidebar active state fix** — Home sidebar item muted when feed tabs (Following/Global) are active
- **Follow pill** — "+ Follow" chip next to non-followed author names in feed cards
- **Inline card expansion** — clicking a card expands it in-place with comments + related content below (search bar stays visible)
- **Comments card** — Layers-style comments section with header, count badge, inline reply input (cyan Send pill), lightweight CommentItem composables with Reply/Like/Zap actions
- **Related content** — image-overlay cards (200x140dp) with gradient, author name, zap count; "View all" navigates to author profile
- **Share menu** — 6 options replacing the old overflow menu (Copy Text, Note ID, Event Link, Raw JSON, Web Link, Broadcast)
- **Slide-animated navigation** — 200ms slide transitions for overlay views (profile, thread from related content), 2-level back stack cap, Esc key to go back

## PoW

https://github.com/user-attachments/assets/1439d73e-d3dd-40c2-b88e-ccebb804a5d4

Screenshot 2026-06-02 at 09 46 09

## Changes

| Area | Files |
|------|-------|
| Card restructure | `NoteCard.kt` (bottomContent + headerTrailingContent slots) |
| Feed inline expansion | `FeedScreen.kt` (expandedNoteId state, ExpandedNoteContent composable) |
| Thread composables | `CommentsCard.kt`, `CommentItem.kt`, `InlineReplyInput.kt`, `RelatedContentRow.kt` |
| Share menu | `ShareMenu.kt`, `NoteActions.kt` |
| Navigation | `DeckColumnContainer.kt`, `SinglePaneLayout.kt` (AnimatedContent slides) |
| Sidebar | `DeckSidebar.kt` (muted state), `Main.kt` (feedTabActive) |
| Cache | `DesktopLocalCache.kt` (lastContactListEvent) |
| Commons | `CompactNoteData.kt` (@Immutable data class) |

## Testing

- Manual testing on macOS with live Nostr relays
- Verified: card expansion, comment loading, metadata resolution, reply posting, like reactions, follow action, share menu clipboard operations, slide navigation, Esc key handling
- `./gradlew :desktopApp:compileKotlin` passes
- `./gradlew spotlessApply` clean
- Pre-push hooks pass (all tests)

No additional operational monitoring required: desktop-only UI changes with no backend/relay protocol modifications.

---

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


davotoula:
Some findings, feel free to push back on any

Findings

Related content is computed once and never refreshes — RelatedContentRow.kt:83. DisposableEffect(noteId) scans localCache.notes a single time at expansion. noteHashtags/authorPubKey aren't keys, and nothing re-runs the scan as the cache fills. Expand a note on a cold cache → the "Related" section stays empty/partial even as matching notes stream in; only collapse+re-expand fixes it.

Deck columns steal keyboard focus from each other — DeckColumnContainer.kt:150. LaunchedEffect(currentOverlay) { focusRequester.requestFocus() } grabs focus unconditionally on every column's first composition and on every overlay change. In a multi-column deck, typing in column A's compose box loses focus the moment column B opens/changes an overlay.

Inline desktop reply emits a lower-fidelity tag set than the canonical reply builder — FeedScreen.kt:1554. The inline TextNoteEvent.build adds only one eTag(event.id) + one pTag(event.pubKey) — no NIP-10 root marker, no carrying of the parent's root/p-tags. Replying to a note that is itself deep in a thread produces an event with no root reference, so other clients can't reconstruct the thread. The dialog/ViewModel path handles this; the inline path reimplements a subset.

Reaction/follow/reply business logic lives inside desktop composables — FeedScreen.kt:1611 (reaction), :423 (follow), :1554 (reply). Event build + sign + cache-consume + broadcast are inlined in UI rather than a shared ViewModel/commons, duplicating ReactionAction/FollowAction/CommentPostViewModel responsibilities. Per CLAUDE.md this belongs in commons; as-is it will drift from the Android path (see #3).

Zap totals bypass the shared formatter — RelatedContentRow.kt:137 renders "${note.zapsAmount.toLong()}" (raw, e.g. 1500000), and CommentItem.kt:166 hand-rolls its own formatZapAmount k/M logic. Both should use the existing toZapAmount()/showAmount helper for consistent k/M output

nrobi144:
Thanks for the careful read — all 5 addressed in 1 commit on this branch (see latest push).

### Findings

**#3 — NIP-10 reply fidelity (the protocol bug).** Confirmed and fixed. Switched the inline path to `TextNoteEvent.build(content, replyingTo = EventHintBundle(parent, null))` (the reply-aware overload at `quartz/.../nip10Notes/TextNoteEvent.kt:142`), which already encodes the root marker, the reply marker, and parent's root e-tag carry via `prepareETagsAsReplyTo`. NIP-10 also requires p-tag carry, which the build helper doesn't do — wrapped the whole thing in a new `commons/actions/ReplyActions.replyTo(...)` that adds the parent's `linkedPubKeys() + pubKey` via `notify(...)`. Covered by `ReplyActionsTest` (top-level parent, deep-thread chain, sign+kind verification).

**#4 — business logic in composables.** Two distinct sub-points:
1. Build+sign already routed through shared actions for follow (`FollowAction.follow` at `FeedScreen.kt:423`) and reaction (`ReactionAction.reactTo` at `FeedScreen.kt:1611` and `ThreadScreen.kt:424`). Only reply was inlining the build — fixed via the new `ReplyActions` above.
2. The `localCache.consume + relayManager.broadcastToAll` couplet was inlined in 5 sites with inconsistent ordering (replies did consume→broadcast, reactions/follow did broadcast→consume). Extracted `desktopApp/cache/EventDispatch.dispatch(signed, localCache, relayManager)` — canonical local-first order — and migrated all 5 sites.

For the NIP-22 path (`CommentPostViewModel`): kept it separate. That's the kind:1111 article-comments builder, not the right analogue for kind:1 feed replies. Android's `NotificationReplyReceiver` also routes by parent kind (kind:1 → `TextNoteEvent.build`, others → `CommentEvent.replyBuilder`); the desktop fix matches that split. The 4 Android sites that still inline `TextNoteEvent.build` (`ShortNotePostViewModel`, `VoiceReplyViewModel`, `NotificationReplyReceiver`, `AmethystAppFunctions`) can migrate to `ReplyActions` in a follow-up — happy to do that in a separate PR.

**#1 — stale related content.** Replaced the one-shot `DisposableEffect(noteId)` with a `produceState` keyed on `(noteId, authorPubKey, lowercaseTags)` that collects `DesktopLocalCache.eventStream.newEventBundles` and re-runs the scan only when an arriving bundle contains a candidate (same author or shared hashtag). Extracted the scan into a private `scanRelated(...)` so initial + delta paths share it. `LargeCache.notes` is a `ConcurrentSkipListMap` (weakly-consistent iterator) so the scan stays safe on the composition coroutine; wrapped `rescan()` in `runCatching` for transient mutations. Matches the `FeedViewModel.kt:54-59` precedent — no debounce needed because the candidate-filter pre-check blocks 80–90% of bundles. If hot-loop is observed, easy to drop in `.debounce(150)` later (precedent: `SearchBarState.kt:87`).

**#2 — focus stealing.** Changed `LaunchedEffect(currentOverlay)` → `LaunchedEffect(Unit)` in `DeckColumnContainer` so focus is requested once on column creation, not on every overlay state change. Escape still works since the column owns focus at the moment the user hits the key. Also wrapped each `DeckColumnContainer` in `key(column.id)` in `DeckLayout` so the one-shot effect survives column reorder (without `key()`, moving a column in the list could re-fire the effect for the wrong instance and re-steal focus).

**#5 — zap formatter.** Migrated `RelatedContentRow` (now `showAmount(note.zapsAmount)`), `CommentItem.formatZapAmount(zapAmount)` → `zapAmount.toZapAmount()`, and `NoteActions.formatSats(...)` (8 call sites) → `*.toZapAmount()`. Deleted both private helpers. Left `WalletColumnScreen.formatSats` (line 979) as-is and added a comment marking it intentional — wallet balance display uses locale-aware `NumberFormat` grouping (`1,000,000`) for full-precision sats, which is a different display concern from compact k/M/G zap totals.

### Plan / verification

Detailed plan with per-phase rationale + evidence in `docs/plans/2026-06-02-fix-desktop-feed-review-findings-plan.md` (committed alongside).

- `./gradlew :commons:jvmTest` — `ReplyActionsTest` green (3 cases).
- `./gradlew :desktopApp:compileKotlin` — green.
- `./gradlew spotlessApply` — clean.

Manual NIP-10 round-trip + multi-column focus + cold-cache related verification on me before merge.

PR #3137, opened 2026-06-06
Render Agora fundraiser campaigns (kind 33863)

davotoula:
## What
  
  Adds first-class support for **Agora fundraiser campaigns (kind 33863)**, an app-specific addressable kind (Agora is a crowdfunding client on the Ditto stack; the kind has no NIP). Previously these hit the "Event Not Supported" path and were dropped — a repost of one showed as a permanently blank card.
  
  ## Changes
  
  - **quartz:** `FundraiserEvent` (`experimental/agora`, kind 33863), registered
    in `EventFactory`. Reuses standard tags (`title`/`banner`/`imeta`/`t`/
    `published_at`) plus fundraiser-specific `goal` / `deadline` / `w` value tags,
    read via the shared `firstTagValueAsLong` / `mapValueTagged` helpers (same
    pattern as `CalendarTimeSlotEvent`).
  - **LocalCache:** consumes it via `consumeBaseReplaceable` so it's stored, not
    dropped.
  - **UI:** `RenderFundraiser` shows banner, title, body, a goal progress bar
    (reuses `GoalProgressBar` — progress is from zaps, the only signal Amethyst
    can observe; on-chain `w` addresses are shown as tap-to-copy, not counted),
    deadline, and hashtags. Wired into **both** render paths — the feed card
    (`NoteCompose`) and the opened/thread view (`ThreadFeedView`). 
  - **Feeds:** added to the Home, FollowPack and Profile allow-lists so
    standalone fundraisers appear.
    
  Because `EventFactory.isKnownKind(33863)` now returns true, the existing hide-reposts-of-unsupported-kinds filter automatically stops hiding reposts of fundraisers.
  
  ## Testing
  
  - `FundraiserEventTest` — factory dispatch, field parsing, known-kind.
  - Verified on device: the reposted campaign now renders the full card (banner +
    title + body + goal progress + copyable on-chain addresses + hashtags) in both
    the feed and the opened view.
    
  ## Notes
  
  - Desktop is out of scope (its simpler renderer + lack of a missing-event
    fetcher need separate work) — documented separately. 
  - `goal` is treated as sats; progress reflects Lightning zaps only.

## License

      original license header.


davotoula:
@nrobi144 

## The three desktop gaps (in dependency order)

### 1. Fetch 

Desktop never pulls the event off a relay, so there is nothing to store or
render. Two sub-cases:

- **Reposted fundraiser** (the common case): desktop's feed subscription
  fetches `kinds = listOf(1, 6, 16)` (`desktop/subscriptions/FeedSubscription.kt`,
  ~line 381). It gets the kind‑16 repost, but the repost's *boosted* event is
  kind 33863, which is never requested. Reposts of kind‑1 notes work only
  because kind 1 is already in that list.
- **Standalone fundraiser**: same — 33863 isn't in any subscription.

Desktop has **no missing-event fetcher** (Android's `FilterMissingEvents` /
gatherer system has no desktop equivalent — confirmed: no `FilterMissing`, no
fetch-boosted-by-id, no by-address chase under `desktop/subscriptions/`).

### 2. Store

`DesktopLocalCache` drops any kind it has no explicit consumer for
(`else -> false`). 

### 3. Feed-accept

`desktop/feeds/DesktopFeedFilters.kt` → `isFeedNote` allow-lists what shows in a
feed.

PR #3141, opened 2026-06-07
fix(quartz): NIP-46 bunker double-resume + retry id-reuse races

nrobi144:
## Summary

Two correctness bugs in `RemoteSignerManager` (NIP-46) and its NIP-55 sibling `IntentRequestManager`:

1. **Double-resume crash** — `awaitingRequests.get(id)?.resume(value)` is non-atomic. Multi-relay delivery, bunker echo/retry, and late-after-timeout responses can call `resume` twice on the same continuation, throwing `IllegalStateException: Already resumed` on a `Dispatchers.Default` worker. Sample stack trace from a user report:

    ```
    Exception in thread "DefaultDispatcher-worker-42" java.lang.IllegalStateException:
      Already resumed, but proposed with update BunkerResponse@…
        at kotlinx.coroutines.CancellableContinuationImpl.alreadyResumedError(CancellableContinuationImpl.kt:556)
        at com.vitorpamplona.quartz.nip46RemoteSigner.signer.RemoteSignerManager.newResponse(RemoteSignerManager.kt:49)
        at com.vitorpamplona.quartz.nip46RemoteSigner.signer.NostrSignerRemote$subscription$2$1.invokeSuspend(NostrSignerRemote.kt:83)
    ```

2. **Retry id-reuse → wrong data** (NIP-46 only) — `launchWaitAndParse` built the request and event once, then reused the same `request.id` across retry attempts. A late response from attempt N could resume attempt N+1's continuation with attempt N's data. Default `timeout = 65_000` + `delay(2_000)` between attempts gives a 2-second window where this misdelivery is reachable on flaky relays.

## Approach

Replace the cached-`Continuation` map with the in-house **Channel-per-request** correlation pattern already used in `quartz/.../accessories/NostrClientPublishExt.kt` and four sibling files:

- `LargeCache<id, Channel<Response>(capacity=1)>` — atomic `remove` resolves the multi-delivery race; `Channel(1)` makes late `trySend` after timeout a safe no-op.
- Fresh `request.id` per retry attempt — closes the cross-attempt misdelivery window. The public `bunkerRequestBuilder` is still called only once per call; the manager rewrites the id internally before each publish.
- `finally`-block cleanup removes the cache entry on success, timeout, and exception — incidentally fixes a latent memory leak on the success path.

No new `@OptIn(InternalCoroutinesApi)` — the previous `CancellableContinuation.tryResume` alternative was discarded in favour of the established house pattern.

## Files changed

| File | Change |
|---|---|
| `quartz/.../nip46RemoteSigner/signer/RemoteSignerManager.kt` | Channel-per-attempt + fresh id per retry |
| `quartz/.../nip55AndroidSigner/api/foreground/IntentRequestManager.kt` | Channel-per-request (single attempt) |
| `quartz/.../RemoteSignerManagerRetryTest.kt` | +3 race-condition tests |
| `quartz/plans/2026-06-03-fix-nip46-bunker-double-resume-plan.md` | Design + review notes |

## Testing

`./gradlew :quartz:jvmTest` — 9/9 `RemoteSignerManagerRetryTest` green (5 pre-existing + 3 new + 1 default check). Full quartz suite green.

The three new regression tests were verified against the **buggy original** by stashing the `RemoteSignerManager.kt` fix and re-running:

- `duplicateResponsesAreSafeAndResumeOnce` — FAILS with `IllegalStateException` (Bug 1 reproduced)
- `lateResponseFromAttempt1DoesNotCorruptAttempt2` — FAILS at `assertNotEquals(attempt1Id, attempt2Id)` (Bug 2 reproduced: same id across retries)
- `lateResponseAfterTimeoutIsSilentlyDiscarded` — passes on `main` too (`invokeOnCancellation` cleanup runs before the late delivery in the test sequence); kept as a regression guard for the channel-cleanup contract.

## Out of scope

- NIP-46 `auth_url` (multi-response per id) — Amethyst has no `auth_url` code today; every parser maps `error != null` to terminal `Rejected`. The right fix when added is at the parser layer, not the correlation cache.
- Subscription-level event-id dedupe in `NostrSignerRemote.kt:82` — Channel `trySend(capacity=1)` already makes duplicates O(1) and safe. Reconsider if field telemetry shows high duplicate volume.

## Design + review

Full plan + review notes: `quartz/plans/2026-06-03-fix-nip46-bunker-double-resume-plan.md`

### DIFF
--- a/quartz/plans/2026-06-03-fix-nip46-bunker-double-resume-plan.md
+++ b/quartz/plans/2026-06-03-fix-nip46-bunker-double-resume-plan.md
@@ -1,0 +1,523 @@
+---
+title: "fix(quartz): NIP-46 bunker double-resume + retry id-reuse races"
+type: fix
+status: completed
+date: 2026-06-03
+origin: docs/brainstorms/2026-06-03-nip46-bunker-double-resume-brainstorm.md
+---
+
+# fix(quartz): NIP-46 bunker double-resume + retry id-reuse races
+
+## Revision Note (2026-06-03, post-review)
+
+Three reviews (simplicity / architecture / pattern-recognition) converged on
+pivoting the approach. Original plan used atomic-remove + `tryResume` on a
+cached `Continuation` map. **Revised approach: Channel-per-request (per
+retry attempt) + fresh `request.id` per attempt**, matching the de facto
+Quartz convention used in `NostrClientPublishExt.kt` and 4 sibling files
+under `quartz/.../accessories/`.
+
+### Why the pivot
+
+| Driver | Detail |
+|---|---|
+| **Architecture review finding** | Quartz already uses Channel-per-request as house style (5+ files). The cached-`Continuation` pattern in `RemoteSignerManager` / `IntentRequestManager` is the outlier — only 2 files. |
+| **4th failure mode discovered** | `RemoteSignerManager.kt:74-101` retry loop reuses the same `request.id` across attempts. Late response from attempt 1 can resume attempt 2's continuation with **stale data** (correctness bug, not just crash). Investigation verdict: HIGH probability on flaky relays. Fresh-id-per-retry naturally fixes this. |
+| **Simplicity review** | Pivoting eliminates the `@InternalCoroutinesApi` opt-in surface entirely, kills the Plan C / SingleShotContinuation alternative discussions, removes the `TestLogCapture` test-infra invention, and removes the Strategy B stress-loop test. |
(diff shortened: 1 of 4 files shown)


nrobi144:
@greenart7c3 I've been getting these issues lately, can you look if this fix seems good?

greenart7c3:
Looks good to me

greenart7c3 (approved):

PR #3166, opened 2026-06-10
Compute legacy NIP-71 video addresses with their d tag

davotoula:
Fixing one more missing event type...

## Problem
  
`ReplaceableVideoEvent` (base of kinds 34235/34236) extended `BaseReplaceableEvent`, whose `dTag()` is the fixed empty string meant for plain replaceables (kinds 10000–19999). Both video kinds are parameterized replaceables, so `address()`/`addressTag()` resolved to `34236:<pubkey>:` and `LocalCache` consumed the event into a different `AddressableNote` than the one `a` tags reference.

Visible symptom: quotes and kind-16 generic reposts of these videos (e.g. published by the Divine client) showed an eternal "Event is loading" placeholder even after the event was successfully fetched — the event landed in the wrong-address note while the UI watched the right-address one.

Repro: `nostr:nevent1qqsqm89ju73883lzxddgzmryqwxud4jle9vjuhungsrrl4hafasp72sprdmhxue69uhhg6r9vehhyetnwshxummnw3erztnrdakj7q3qfhpw2ux9flhcxyl6xp84996qgnkkcy59zqzjvq9fhpxcx7upymusxpqqqqqpqas45pr`

## Fix
  
Extend `BaseAddressableEvent` instead, so `dTag()` reads the real `d` tag. Events without a `d` tag keep the old `kind:pubkey:` address (`tags.dTag()` falls back to `""`), so they are unaffected.

Side benefits: an author's videos no longer replace each other in cache (previously they all shared one `34236:<pubkey>:` slot), and shared naddr links / NIP-09 deletion `a` tags for these kinds now carry a resolvable d tag.

## Testing  
  - New `ReplaceableVideoEventAddressTest` (commonTest): address computation for 34235/34236 with the real Divine wire event above as fixture, plus the no-`d`-tag fallback.
  - `:quartz:jvmTest` and `:commons:jvmTest` pass; verified on device that the repro nevent above now opens to the video instead of the placeholder.
  - Swept quartz: no other addressable-range kind extends `BaseReplaceableEvent`.
## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/profile/gallery/dal/UserProfileGalleryFeedFilter.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/profile/gallery/dal/UserProfileGalleryFeedFilter.kt
@@ -32,8 +32,8 @@ import com.vitorpamplona.amethyst.ui.dal.FilterByListParams
 import com.vitorpamplona.quartz.experimental.profileGallery.ProfileGalleryEntryEvent
 import com.vitorpamplona.quartz.nip53LiveActivities.clip.LiveActivitiesClipEvent
 import com.vitorpamplona.quartz.nip68Picture.PictureEvent
+import com.vitorpamplona.quartz.nip71Video.AddressableVideoEvent
 import com.vitorpamplona.quartz.nip71Video.RegularVideoEvent
-import com.vitorpamplona.quartz.nip71Video.ReplaceableVideoEvent
 import com.vitorpamplona.quartz.nip71Video.VideoVerticalEvent
 
 class UserProfileGalleryFeedFilter(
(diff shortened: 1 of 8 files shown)


vitorpamplona:
Should we also change the name ReplaceableVideoEvent to AddressableVideoEvent for consistency?

davotoula:
> Should we also change the name ReplaceableVideoEvent to AddressableVideoEvent for consistency?

Yes, it was suggested but slightly larger diff. Let me add that

davotoula:
**`refactor: rename ReplaceableVideoEvent to AddressableVideoEvent`**. It also pairs naturally with `RegularVideoEvent` (kinds 21/22). Same constructor/API surface, so only the two subclasses and one filter import changed.

**`test: guard kind-range vs class-hierarchy invariant in EventFactory`** — generalises this PR's regression test so the bug class can't recur: it sweeps every typed kind in `EventFactory` and asserts that addressable kinds (30000–39999) read their `d` tag while plain replaceables (10000–19999, 0, 3) ignore stray ones.

The sweep surfaced two pre-existing groups, allowlisted with comments rather than fixed here to keep this PR focused:

  - **NIP-87 ecash kinds 38000/38172/38173** extend plain `Event` — no addressability at all, so they never replace older versions in the cache and `a`-tag references can't resolve by address.
  - **22 NIP-51-style list kinds (10004…10102)** read stray `d` tags through the shared `PrivateTagArrayEvent` hierarchy, so a malformed event with a `d` tag escapes replacement under a fragmented address.

PR #3175, opened 2026-06-11
feat(desktop): replace vlcj (GPLv3) with kdroidFilter ComposeMediaPlayer + JCodec/FFmpeg

nrobi144:
## Summary

Drops `uk.co.caprica:vlcj 4.8.3` (GPL-3.0) from `desktopApp` and the bundled VLC 3.0.20 plugin tree it pulled in via `ir.mahozad.vlc-setup`. Replaces them with `io.github.kdroidfilter:composemediaplayer:0.10.0` (MIT) for video + audio playback, and `org.jcodec:jcodec(+javase):0.2.5` (BSD-2) + a raw `ProcessBuilder` LGPL FFmpeg fallback for thumbnail extraction. The distributed binary's SPDX moves from an undeclared `GPL-3.0 AND LGPL-2.1 AND MIT` mix (today's `rpmLicenseType = "MIT"` is misdeclared) to **`MIT AND LGPL-2.1-or-later AND BSD-2-Clause AND Apache-2.0`**.

kdroidFilter ships first-class `VideoPlayerSurface` composables that render via Compose `Canvas` + Skia `ImageBitmap` — no `SwingPanel`, so the prior z-order/overlay issues with vlcj's heavyweight surface are gone. macOS / Windows use OS-native backends (AVFoundation / Media Foundation), so no per-OS native libs are bundled by us. Linux relies on system GStreamer (declared as a runtime dep + documented in the Flathub manifest).

Plan and decision log live in:
- `docs/plans/2026-06-11-feat-replace-vlcj-with-kdroidfilter-plan.md`
- `docs/plans/2026-06-11-vlcj-replacement-testing-sheet.md`

## Test plan

- [x] `./gradlew :desktopApp:spotlessApply` — clean
- [x] `./gradlew :desktopApp:test` — `BUILD SUCCESSFUL` in 21s on macOS arm64
- [x] `./gradlew :desktopApp:compileKotlin` — `BUILD SUCCESSFUL` in 16s on macOS arm64
- [x] Manually exercised: launched on macOS arm64, played multiple feed videos, verified runtime behaviour

A detailed proof-of-work matrix (per-case results, runtime caches, log evidence) is in the first review comment on this PR.

**Notes / known gaps:**
- The `LICENSE-LGPL-2.1.txt` file ships as a one-page placeholder pointing at FSF — please replace with the verbatim ~26 KB text before tagging a release. Flagged in `appResources/common/licenses/LICENSE-LGPL-2.1.txt`.
- Per-OS LGPL FFmpeg binaries are **not** checked in (~20-30 MB each). `appResources/{macos,windows,linux}/ffmpeg/README.md` points at the recommended LGPL build per OS. Packaging tasks should drop them in before `:desktopApp:packageDmg` / `packageMsi`.
- An in-app "Open source licenses" screen using `mikepenz/AboutLibraries` is a recommended follow-up — for now `NOTICE.md` and the per-component `LICENSE-*.txt` files ship inside the installer's `app/resources/common/` directory.
- On Linux, kdroidFilter requires the host's `gstreamer1.0-plugins-{base,good,bad}` + `gstreamer1.0-libav`. Documented in the AppImage `AppRun` and in the Flathub manifest (via `org.freedesktop.Platform 24.08` + `org.freedesktop.Platform.ffmpeg-full` add-extension).

## Interop suites

- [x] N/A — change can't affect wire bytes / decoded audio / MLS state / DM envelopes (desktop media playback only)

## AI assistance

## License


nrobi144:
## Proof of work

Verification performed locally on **macOS 25.5 / arm64** (Apple Silicon) against this branch's HEAD. Paths normalized to `~`; no machine-specific data below.

Screenshot 2026-06-11 at 13 22 50


### Build gates (commands + results)

| # | Command | Result |
|---|---------|--------|
| 1 | `./gradlew :desktopApp:compileKotlin` | `BUILD SUCCESSFUL` in 16s |
| 2 | `./gradlew :desktopApp:test` | `BUILD SUCCESSFUL` in 21s |
| 3 | `./gradlew :desktopApp:spotlessApply` | clean, no diff |
| 4 | `grep -rn 'vlcj\|caprica\|VlcjPlayerPool\|MacOsVlcDiscoverer\|BundledVlcDiscoverer\|VlcResourceResolver' desktopApp/src/ gradle/ desktopApp/build.gradle.kts` | only doc-comment mentions remain (intentional, in `VideoThumbnailCache.kt` header describing what was replaced) |

### Runtime smoke (app launched via `./gradlew :desktopApp:run`)

| Check | Evidence | Result |
|---|---|---|
| App launches | Java process alive 3:32+ before being closed cleanly | ✅ |
| No vlcj / VLC log mentions post-launch | `grep -ciE 'vlcj\|libvlc\|VLC plugin\|caprica'` against the gradle daemon log filtered to the post-launch time window → **0 hits** | ✅ |
| No uncaught exceptions post-launch | `grep -iE 'Exception\|Error\|throw'` (excluding gradle's own EOFException noise and pre-existing warnings) on the post-launch window → **0 hits** | ✅ |
| kdroidFilter native library extracted | `libNativeVideoPlayer.dylib` (202 KB, Mach-O 64-bit arm64) appeared at `~/.cache/composemediaplayer/native/darwin-aarch64/` on first playback request | ✅ |
| App shut down cleanly | No segfault, no zombie processes after window close | ✅ |

### Codec / playback matrix

| # | Case | Result |
|---|------|--------|
| 3.1 | H.264 MP4 (golden path) — multiple feed videos | ✅ played, frame on `VideoPlayerSurface`, controls work |
| 3.2 | HLS livestream | Not exercised in this session |
| 3.3 | HLS VOD H.264+AAC | Not exercised in this session |
| 3.4 | **VP9 in WebM** (`nostr-video-VP9-opus-720.webm`) | ❌ — AVFoundation `Error loading metadata: Operation Stopped`, `Error loading video tracks: Operation Stopped`. macOS AVPlayer does not natively support VP9-in-WebM. **Expected codec gap** (testing sheet §3.4). The `PlaybackErrorMessage` composable should render the "Open in default player" handoff; needs a follow-up to verify that kdroidFilter routes this specific error into our `state.error` (it may only log to native stderr without firing the Compose-observable error). |
| 3.5 | AV1 MP4 | Not exercised in this session |
| 3.6 | HEVC MP4 | Not exercised in this session |
| 3.7 | Audio MP3 | Not directly tested as audio-only, but inferable from H.264 video playback (audio tracks decoded — confirmed by AVPlayer's "Audio tap created successfully" diagnostic for each H.264 video) |
| 3.10 | Garbage URL / 404 | Indirectly verified — thumbnail path correctly skipped invalid origins (see "HTML error page" finding below) |

### UI cross-state behaviour

| # | Scenario | Result |
|---|----------|--------|
| 4.1 | Play feed video → navigate away → return | Not specifically re-verified post-fix but the design preserves the singleton `GlobalMediaPlayer` + `VideoPlayerState` |
| 4.2 | Play A → play B | ✅ (multiple videos played in sequence, each replaced the previous) |
| 4.3-4.10 | Fullscreen, Spacebar, mute, navigate, stop, quit | App quit cleanly; other scenarios remain for the reviewer's verification on their machine |

### Thumbnail cache verification

Cache lives at `~/.cache/amethyst-desktop/video-thumbs/`. Filename is the SHA-1 of the source URL + `.mp4`.

| Property | Value |
|----------|-------|
| Cache entries after a normal session | 8 files |
| Total size after the bounded-download fix | 25 MB (vs **626 MB** before the fix when one origin ignored `Range:` and we streamed the full body) |
| Largest single file | 4.8 MB (the cap is 4 MiB; OkHttp can over-read slightly past the boundary before the loop check fires) |
| All cached file `ftyp` brand | `avc1` (H.264 baseline/main/high) — JCodec path eligible |
| `magic file` for each entry | `ISO Media, MP4 Base Media v1` or `MP4 v2` — valid containers, JCodec confirmed it would decode them |

This proves:
- ✅ JCodec primary path is being exercised (range-fetch → buffered MP4 → JCodec frame grab → Skia `ImageBitmap`).
- ✅ The 4 MiB cap is enforced regardless of whether the server honours `Range:`.
- ❓ The ffmpeg-fallback path is **not exercised** in this verification window because every URL the user happened to play was H.264. Testing sheet §5.2 (VP9/HEVC/AV1 thumbnail) remains unverified by direct cache evidence.

### Bugs found during verification and fixed in this PR

1. **Unbounded download when origin ignores `Range:`** — without the cap, one 617 MB video file was streamed to disk because the server returned HTTP 200 (full body) instead of HTTP 206 (range). Fixed by adding `MAX_THUMB_BYTES = 4 MiB` and a `copyAtMost(in, out, limit)` helper around the body copy, plus a `cached.length() == 0L → cached.delete()` cleanup on the failure path.
2. **HTML error page cached as `.mp4`** — one origin served `<!DOCTYPE html>` (a 403/404 wrapped in HTML) which JCodec correctly rejected; ffmpeg fallback also failed; but the bogus file persisted in the cache and a re-thumbnail attempt would short-circuit on its non-zero size. Fixed by sniffing the `Content-Type` response header and refusing to write if it starts with `text/` or contains `html`.
3. **Zero-byte cache entries from empty responses** — when an origin returned an empty body, we created a 0-byte cache file that the next read short-circuited around (`length() > 0L` check returned false, but the file lingered). Fixed by deleting any zero-byte file the download path produces.

All three fixes are in `desktopApp/src/jvmMain/kotlin/com/vitorpamplona/amethyst/desktop/service/media/VideoThumbnailCache.kt`.

### License metadata sanity check

```
grep 'rpmLicenseType' desktopApp/build.gradle.kts
→ rpmLicenseType = "MIT AND LGPL-2.1-or-later AND BSD-2-Clause AND Apache-2.0"

ls desktopApp/src/jvmMain/appResources/common/
→ NOTICE.md  licenses/

ls desktopApp/src/jvmMain/appResources/common/licenses/
→ LICENSE-BSD-2-jcodec.txt
  LICENSE-LGPL-2.1.txt           ← placeholder; replace before release
  LICENSE-MIT-amethyst.txt
  LICENSE-MIT-kdroidfilter.txt
```

`LICENSE-LGPL-2.1.txt` ships as a one-paragraph pointer at FSF and must be replaced with the verbatim 26 KB GNU LGPL-2.1 text before any tagged release. Flagged in the testing sheet §6.4 and in the file itself.

### What still needs a human (testing sheet §3 / §4 / §7 / §8)

- HLS livestream playback (zap.stream, `test-streams.mux.dev`)
- HEVC / AV1 on macOS 13+ and on Windows (Microsoft Store extension flow)
- Linux GStreamer codec coverage end-to-end (DEB / RPM / AppImage / Flathub)
- macOS Developer ID code-signing of the nested FFmpeg binary inside the DMG
- Flathub local build via `flatpak-builder` and an actual Flathub submission

Coverage matrix and concrete URL set live in `docs/plans/2026-06-11-vlcj-replacement-testing-sheet.md`.

davotoula:
Please review review 😆 

Correctness (should block merge)
  
  1. Thumbnail cache poisoning by truncated downloads — VideoThumbnailCache.kt:152. The downloader sends a Range header but accepts a 200 (full-body) response and caps the copy at 4 MiB. The truncated file passes the only
  validity check (cached.length() > 0) and is then a permanent cache hit; only zero-byte files are cleaned up (:165-167). One server that ignores Range → that video's thumbnail is broken forever, and every retry re-feeds the
  same poisoned file to JCodec/ffmpeg.
  2. Initial-seek collector leak / stale re-seek — GlobalMediaPlayer.kt:121-133. The snapshotFlow { player.hasMedia } collector launched in playVideo() is never cancelled — return@collect only exits one emission, not the
  collection. Each playVideo(url, seekPosition) accumulates a live collector; a later hasMedia false→true transition (next video) re-fires a stale seek on the wrong media. startVideoSync does this right
  (videoSyncJob?.cancel()); the seek path needs the same treatment, or simply .first { it }.
  3. Mute/volume desync on new video — GlobalMediaPlayer.kt:119. playVideo() resets _videoState to volume=100, isMuted=false, but the kdroidFilter engine's volume is sticky across openUri. Mute a video, play the next one →
  audio is silent while the UI shows unmuted/100%. (Found while refuting the original "start-volume" candidate against the library source.)
  4. Native player creation in composition, no error handling — activeVideoPlayerState getter calls ensureVideoPlayer() synchronously; read from composition at DesktopVideoPlayer.kt:117, GlobalFullscreenOverlay.kt:107,
  NowPlayingBar.kt:91. If VideoPlayerState() throws (missing native lib, unsupported distro) the whole window crashes — exactly the case this migration makes more likely on Linux without GStreamer. Wrap creation and surface it
  through the existing PlaybackErrorMessage path.
  5. Thumbnail downloads bypass Tor — VideoThumbnailCache.kt:65-71 builds a bare OkHttpClient while the rest of desktop media goes through DesktopHttpClient.currentClient() (see DesktopImageLoaderSetup.kt:58). With Tor
  enabled, thumbnail fetches leak the user's IP to media hosts. Not a regression (vlcj used VLC's own stack — which also bypassed, FWIW), but this PR hand-rolls HTTP, so it should use the shared client.

  Performance / robustness

  6. Unbounded thumbnail pipeline — VideoThumbnailCache.kt:62-63,110-121: no concurrency cap (N visible videos → N parallel downloads + N ffmpeg processes), and the in-memory ConcurrentHashMap<String, ImageBitmap> has no
  eviction — long scrolling sessions accumulate full-size bitmaps until OOM. A Semaphore(2-3) + size-bounded LRU fixes both. Also pending.putIfAbsent makes the second concurrent caller get null (and retry) rather than join the
  first request.
  7. ffmpeg stderr deadlock risk + probe leak — runFfmpegToImage uses redirectErrorStream(false) but never drains stderr; a chatty ffmpeg fills the pipe and stalls until the 8s timeout kills it (works, but every failure costs
  8s). The ffmpeg -version probe (:83-89) doesn't destroyForcibly() on timeout. Redirect stderr to DISCARD and destroy the probe.
  8. Per-tick StateFlow churn + twin sync functions — GlobalMediaPlayer.kt:266-335. The snapshot includes currentTime, so every engine tick allocates EngineSnapshot + MediaPlaybackState.copy() and recomposes all collectors; no
  distinctUntilChanged on meaningful fields. Also startVideoSync/startAudioSync are 96% identical (and the mute emulation is repeated 4×) — one parameterized helper.

  Housekeeping / compliance
  
  9. Dead VLC steps still in CI — build.yml:82-156 (vlc-setup cache + "Pre-fetch VLC + UPX archives") and smoke-test-desktop.yml:71-99 still run against a plugin that no longer exists; create-release.yml:27 comment too. Wasted
  CI time + misleading.
  10. Compliance loose ends: (a) LICENSE-LGPL-2.1.txt is an explicit placeholder — fine as a draft, but it needs either the verbatim text now or a build-time check that fails packaging while it's a stub; (b) NOTICE.md
  references per-OS ffmpeg/README.md drop-in dirs that are not in the commit (commit message claims them too — they're missing); (c) rpmLicenseType declaring LGPL is arguable since no ffmpeg/GStreamer actually ships in the
  binary by default; (d) heads-up for dev machines: the old vlc-setup plugin left an untracked appResources/macos/vlc/ tree (libvlc + ~100 plugins) that appResourcesRootDir will happily bundle into local macOS packages — worth
  a cleanup note or a packaging-time exclusion, or stale workspaces ship GPL binaries by accident.

davotoula:
manual testing issues:

~~minor: cannot minimise / change volume when video has been maximised~~

crash on macos (apple silicon):

**Diagnosis** (verified from the log): the crash is inside kdroidFilter 0.10.0's MacVideoPlayerSurface draw path. The register dump shows the Java-side org.jetbrains.skia.Bitmap object is a valid oop, but the native call
  segfaults at RasterFromBitmap+0x0 — i.e. the native SkBitmap behind the video frame was already freed/recycled while Compose was drawing it. Classic use-after-free race between the AVPlayer frame-update thread (recycling the
  frame bitmap) and the render thread. This is a library bug, not something the migration code can fully guard against.

  **Likely amplifier** in our code (hypothesis, but well-supported): the branch mounts multiple VideoPlayerSurfaces for the same player state simultaneously — the in-feed surface (DesktopVideoPlayer.kt:116) plus the mini-preview
  in NowPlayingBar.kt:90 (both render whenever a video is active; the fullscreen overlay is a potential third). Two draw observers on the same frame bitmap doubles the draw rate and widens the race window — and kdroidFilter's
  own 0.10.1 patch note ("recover video playback after composition removal") suggests surface attach/detach bookkeeping is exactly where this library is fragile. I flagged the triple-mount as finding #4 in the review; this
  crash is the concrete consequence.
  
  Recommended actions for the PR
  
  1. Bump composemediaplayer 0.10.0 → 0.10.1 (latest; includes the composition-removal recovery fix).
  2. Mount only one VideoPlayerSurface at a time — e.g. drop the NowPlayingBar mini-preview (use a static thumbnail/icon there) or render it only when the feed/fullscreen surface is not composed. This is the cheapest
  mitigation regardless of the library bug.
  3. Report upstream to kdroidFilter/ComposeMediaPlayer with hs_err_pid11554.log — single-surface use may still crash eventually, since the UAF is in their frame lifecycle.

nrobi144:
## Review fixes (commit eae1ed88e)

Thanks for the deep review. I reproduced or analyzed each finding against the code, fixed the consensus issues, and shipped them in a single commit on top of the merge from main. Quick status table, then details.

| # | Finding | Status | Where |
|---|---------|--------|-------|
| 1 | Cache poisoning by truncated downloads (200 vs 206) | ✅ fixed | `VideoThumbnailCache.kt` |
| 2 | `snapshotFlow { hasMedia }` collector leak from `return@collect` | ✅ fixed | `GlobalMediaPlayer.kt` |
| 3 | Mute / volume desync across `openUri` | ✅ fixed | `GlobalMediaPlayer.kt` |
| 4 | Native player creation crashes the window | ✅ fixed | `GlobalMediaPlayer.kt` + nullable `activeVideoPlayerState` |
| 5 | Thumbnail downloads bypass Tor | ✅ fixed | `VideoThumbnailCache.kt` now uses `DesktopHttpClient.currentClient()` |
| 6 | Unbounded thumbnail pipeline / pending-singleflight returns null | ⏸️ follow-up | needs Semaphore + LRU; intentionally out of scope of this PR — see below |
| 7 | ffmpeg stderr deadlock + probe leak | ✅ fixed | `redirectError(DISCARD)` + `destroyForcibly` on timeout |
| 8 | Per-tick `StateFlow` churn + twin sync functions | ⏸️ follow-up | low-impact optimization; deferred |
| 9 | Dead VLC steps in CI workflows | ✅ removed | `build.yml`, `smoke-test-desktop.yml`, `create-release.yml` |
| 10a | `LICENSE-LGPL-2.1.txt` placeholder | acknowledged | release-gate item, flagged in file + testing sheet §6.4 |
| 10b | Per-OS `ffmpeg/README.md` files not in commit | ✅ fixed | `.gitignore` refined; the 3 README files now tracked |
| 10c | `rpmLicenseType` declaring LGPL is arguable | acknowledged | defensible since LGPL FFmpeg ships per the packaging path; happy to revisit |
| 10d | Stale `appResources/macos/vlc/` workspace bundling risk | ✅ mitigated | refined `.gitignore` keeps the per-OS dirs ignored except for the `ffmpeg/README.md` slot, so any stale `vlc/` tree in older worktrees won't accidentally land in git or be staged |
| Crash | UAF in `MacVideoPlayerSurface` `RasterFromBitmap`, double-mount amplifier | ✅ mitigated | `NowPlayingBar` now renders the cached thumbnail instead of a second `VideoPlayerSurface`; single surface at a time |

### Notes per finding

**#1 — truncated-download cache poisoning.** Confirmed the code path. I now tag each download result with `persistable: Boolean = (response.code == 206)`. When the origin honours `Range:` we cache the file as before. When the origin returned `200` with the full body (which we still truncate to `MAX_THUMB_BYTES`), we run JCodec/ffmpeg against the temp file and **delete it in a `finally` block**, so subsequent thumbnail requests for the same URL re-fetch fresh from the origin instead of decoding the same half-MP4 forever.

**#2 — collector leak.** Reproduced by reading the code: `return@collect` returns from the trailing-lambda only, the upstream `Flow.collect` keeps running. Replaced with `snapshotFlow { player.hasMedia }.first { it }` which terminates the collection cleanly after the first `true` emission.

**#3 — mute / volume desync.** Confirmed. Added `player.volume = 1f` in both `playVideo()` and `playAudio()` right before `_videoState.value = MediaPlaybackState(... volume = 100, isMuted = false)`, so the engine's `volume` matches the UI default for the new track even if the previous track was muted.

**#4 — native player creation in composition.** Confirmed; `createVideoPlayerState()` can throw. I wrapped both `ensureVideoPlayer()` and `ensureAudioPlayer()` in `runCatching`, changed `activeVideoPlayerState` to nullable, and made `playVideo()`/`playAudio()` set `errorReason = "Video playback unavailable"` on the public state when init fails. `DesktopVideoPlayer` already renders the `PlaybackErrorMessage` composable when `errorReason != null`; `GlobalFullscreenOverlay` now elides the surface and shows a blank backdrop. `NowPlayingBar` is no longer a surface consumer (see crash mitigation below) so it's unaffected.

**#5 — Tor bypass.** Replaced the bare `OkHttpClient` with `DesktopHttpClient.currentClient()`. Thumbnail fetches now go through the same Tor-aware client as the rest of desktop networking, including the fail-closed behaviour when Tor is expected but still bootstrapping.

**#6 / #8 — pipeline boundedness + state churn.** Both are real and worth doing, but they're surgical optimizations that don't change behaviour for the current usage pattern (8-12 visible feed items at a time). I'd rather land them as their own follow-up PR with concrete benchmarks (semaphore concurrency, LRU eviction threshold, `distinctUntilChanged` on the engine-snapshot shape) than bundle them with the migration. Happy to open the follow-up issue if you'd prefer that path.

**#7 — ffmpeg I/O.** Both processes now use `redirectError(ProcessBuilder.Redirect.DISCARD)`; the frame-extract process is wrapped in a `try/finally` that calls `destroyForcibly()` if anything escaped the 8s timeout, and the `ffmpeg -version` probe now redirects stdout to `DISCARD` and calls `destroyForcibly()` when it overruns the 2s probe budget.

**#9 — CI residue.** Removed `Cache vlc-setup downloads` and `Pre-fetch VLC + UPX archives` from `build.yml` and `smoke-test-desktop.yml` (saves the 10-15 min curl/cache round-trip per leg), and trimmed the stale "UPX-compressed VLC plugins" rationale on `create-release.yml`'s linuxdeploy-vs-appimagetool justification.

**#10a — LGPL placeholder.** Agreed it needs the verbatim text before any tagged release. Flagged in the file itself, in `NOTICE.md`, and as item 6.4 of the testing sheet. Happy to do this in a separate commit if you want to gate the PR on it, but my read was that it's a release-gate, not a merge-gate.

**#10b — per-OS ffmpeg READMEs.** Confirmed: the prior `.gitignore` blanket-ignored `desktopApp/src/jvmMain/appResources/{linux,macos,windows}/`. Refined the rules to explicitly track `ffmpeg/README.md` under each OS while keeping any stale `vlc/` directory ignored (so existing worktrees with leftover `appResources/macos/vlc/` from the old vlc-setup plugin don't accidentally stage GPL binaries). The 3 README files are now part of this commit.

**#10c — rpmLicenseType.** The current SPDX expression includes LGPL because the packaging path bundles an LGPL FFmpeg per OS (`appResources/{linux,macos,windows}/ffmpeg/`). The binary slot is documented but not pre-populated; if a packager opts not to drop the binary in, the resulting installer's actual SPDX is narrower. I think the conservative position is to declare the widest-realistic set so we don't under-declare for packagers who do ship FFmpeg — but happy to switch to declaring only what jpackage actually emits without the FFmpeg present, if that's the preferred convention.

**#10d — stale workspace.** The `.gitignore` change above also covers this: any old `appResources/<os>/vlc/` tree stays ignored, so `git add` cannot inadvertently stage GPL binaries from an upgraded checkout. Worth adding a `desktopApp/build.gradle.kts` packaging-time exclusion too, but the gitignore guard is the immediate fix.

### Crash mitigation (review #2)

I couldn't bump `composemediaplayer` to `0.10.1` because that version still doesn't exist on Maven Central as of this writing — `maven-metadata.xml` `latest`/`release` both report `0.10.0`. So I shipped the only mitigation we have control over: **mount one `VideoPlayerSurface` at a time**. `NowPlayingBar` previously mounted a second surface against the same `VideoPlayerState` while the feed card was already mounting one; that doubles the draw rate on the shared frame bitmap and widens the UAF window in `MacVideoPlayerSurface`'s `RasterFromBitmap` path that your stack trace points at. The mini-preview now renders the cached thumbnail (via `VideoThumbnailCache.getCached(url)`) with a music-icon fallback. UX is the same; only the second draw observer is gone.

### Verified on macOS arm64

- `./gradlew :desktopApp:compileKotlin` — `BUILD SUCCESSFUL`
- `./gradlew :desktopApp:test` — `BUILD SUCCESSFUL`
- `./gradlew test` (full suite) — `BUILD SUCCESSFUL`
- `./gradlew :desktopApp:spotlessApply` — clean

davotoula (approved):

PR #3216, opened 2026-06-14
Restore iOS test native-cache workaround dropped in dependency bump

davotoula:
## Problem

`main` is red on the `test-quartz-ios` job (step **Test Commons on iOS**) for every commit since `c7e25c5a9`. The `:commons:linkDebugTestIosSimulatorArm64` task fails at link time under CI's Xcode 16.4:

  > Task :commons:linkDebugTestIosSimulatorArm64 FAILED
  > e: ld invocation reported errors (Undefined symbols: _OBJC_CLASS_$_UIViewLayoutRegion)

  ## Root cause

  Bisecting the job history pins the first failure to `c7e25c5a9`. The culprit is `ece719445` ("Updates all dependencies"), which **deleted** the `disableUiKitPrebuiltCache()` workaround from `commons/build.gradle.kts`.

  Compose Multiplatform 1.11.x ships an `org.jetbrains.compose.ui:ui-uikit` prebuilt Kotlin/Native cache whose `CMPLayoutRegion` object hard-references the UIKit class `UIViewLayoutRegion` (introduced in iOS 17). Linking the
  iOS *test* executable against that cache under Xcode 16.4 fails because the cached object was built for a newer simulator SDK than the test binary is linked for. The workaround disabled the native cache for test binaries so `ui-uikit` recompiles against the active SDK, where the symbol resolves.

  The dependency bump moved Kotlin to 2.4.0, which removed the `DisableCacheInKotlinVersion.2_3_21` enum value the workaround referenced. Rather than bumping the guard, the whole block was deleted — reintroducing the link failure.

  ## Fix

  Restore the workaround using `DisableCacheInKotlinVersion.2_4_0`, the value that exists in KGP 2.4.0. The version guard will re-surface this workaround once Kotlin moves past 2.4.0, so it can be dropped when a newer
  Compose/Kotlin pairing fixes the cache.

  ## Verification

  Ran the exact failing task locally: `:commons:linkDebugTestIosSimulatorArm64` now succeeds, and the build log confirms the cache-disable is applied to the `debugTest` binary ("Kotlin/Native cache is disabled for Debug binary
  'debugTest'"). Local Xcode is 26.x so it can't reproduce the 16.4-specific red, but the fix path is proven applied and linking succeeds; CI on Xcode 16.4 will recompile ui-uikit against the active SDK.

## License

      original license header.

### DIFF
--- a/commons/build.gradle.kts
+++ b/commons/build.gradle.kts
@@ -1,5 +1,23 @@
 
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.DisableCacheInKotlinVersion
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeCacheApi
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.TestExecutable
+
+// Disables the Kotlin/Native compiler cache for an iOS test binary so the
+// Compose ui-uikit klib recompiles fresh instead of linking the broken prebuilt
+// cache (see the call site in the `kotlin {}` block). The version guard makes
+// Kotlin re-surface this workaround once we move past 2.4.0, so it can be
+// dropped when a newer Compose/Kotlin pairing fixes the cache. Wrapped in a
+// helper because @OptIn only applies to declarations, not bare statements.
+@OptIn(KotlinNativeCacheApi::class)
+fun TestExecutable.disableUiKitPrebuiltCache() =
+    disableNativeCache(
+        DisableCacheInKotlinVersion.`2_4_0`,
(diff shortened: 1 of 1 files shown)


davotoula:
@vitorpamplona green ios build
  
  | Check | Result |
  |-------|--------|
  | Build script compiles on KGP 2.4.0 (`2_4_0` enum resolves) | ✅ Pass |
  | `:commons:linkDebugTestIosSimulatorArm64` (the failing CI link) | ✅ Links; cache-disable applied to `debugTest` |
  | `:commons:iosSimulatorArm64Test` (link + boot sim + run) | ✅ 699/699 passed |
  | `:commons:compileTestKotlinIosArm64` (device-arch compile) | ✅ Pass |
  | Pre-commit hook (spotless + static analysis) | ✅ Pass |

vitorpamplona:
It's weird because the iOS build was passing even after the drop of that dependency for a many CI runs. 

When I updated, my claude told be that we didn't need that for the new version anymore. Is this some type of weird caching issue on the CI?

PR #3219, opened 2026-06-14
Don't reset relay reconnect backoff on momentary connections

davotoula:
## Problem

A benchmark read-only sweep (40 min, Pixel 9a) logged **6,364 relay errors**, of which **733 (11%) came from a single relay**, `essayist.decentnewsroom.com` — one reconnect attempt every ~3.3s, sustained for the whole session. Other dead
relays (SSL failures, 4xx/5xx, timeouts) logged only ~15–25 attempts over the same window, i.e. the exponential backoff worked for them.

The distinguishing factor: the storming relay *accepts* the WebSocket handshake (HTTP 101) and then immediately resets the connection (`Connection reset` / `EOF`, alternating). Relays that fail *before* the handshake were unaffected.

## Root cause

`BasicRelayClient.markConnectionAsReady()` — called from the socket's `onOpen` — reset the backoff delay to its 1s minimum on **every** successful handshake:

```kotlin
fun markConnectionAsReady(usingCompression: Boolean) {
    this.isReady = true
    this.usingCompression = usingCompression
    this.delayToConnectInSeconds = DELAY_TO_RECONNECT_IN_SECS   // reset on every onOpen
}
```

So the backoff was not *bypassed* by the abrupt-close failure mode — it was **reset by the momentary "success" that preceded each failure**:

```
connect → onOpen (reset delay → 1s) → server resets socket → onFailure
        → reconnect ~1–2s later → onOpen (reset again) → … forever
```

Reconnect attempts are driven by `reconnect(true)` on every subscription change (debounced 200ms) plus a 60s keep-alive, so during active use the 1s gate was checked constantly — producing the ~2–3s storm. Relays that never reach `onOpen` keep doubling their delay (1→2→4→…→5min cap) as intended.

## The fix

Reset the backoff only after a connection proves **stable** — open for at least `STABLE_CONNECTION_IN_SECS` (60s) — evaluated when the connection closes, rather than on every `onOpen`:

- `markConnectionAsReady()` records `connectedAtInSeconds` and no longer touches the delay.
- `markConnectionAsClosed()` resets the delay to the 1s minimum only if the session lasted ≥ 60s.

Healthy relays (connected for minutes/hours before an occasional drop) behave exactly as before. A handshake-then-reset flapper keeps doubling to the 5-minute cap like any other dead relay. User-initiated `disconnect()` still resets immediately (unchanged).

A small injectable `nowInSeconds: () -> Long = TimeUtils::now` was added to the constructor for deterministic testing; both production call sites (`RelayPool`, `StandaloneRelayClient`) are unchanged.

### Unit test (`BasicRelayClientBackoffTest`)

Simulates 40 minutes of pool ticks (3s cadence) against a fake socket:

- **handshake-then-reset relay** → fails on the old code (hundreds of attempts), passes with the fix (≤20, exponential).
- **pre-handshake failures** → still back off exponentially.
- **stable session then clean close** → backoff resets, reconnects promptly.

### On-device A/B (Pixel 9a, benchmark build)

A throwaway local relay that completes the `101` handshake then immediately drops the socket — reproducing the production relay's behaviour — was added to the device and the reconnect cadence measured from the relay side. Both builds differ by **only this commit**.

| | Control (no fix) | This PR |
|---|---|---|
| Connection attempts in ~140s | **70** | **11** |
| Gap between attempts | **flat ~2.0s**, never grows | **doubling**: 1 → 3 → 4 → 9 → 17 → 34 → 65s |
| Extrapolated to 40 min | **~1,200** | **~12–15** |

The control sat at a flat ~2s cadence the entire window; the fixed build climbed toward the 5-minute cap. This matches the original sweep's 733-attempt storm and confirms the fix eliminates it.

No behaviour change to healthy relays, caching, UI, or NIP-11 handling — this is purely a reconnect-timing correctness fix.

## License

      original license header.

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/relay/client/single/basic/BasicRelayClient.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/relay/client/single/basic/BasicRelayClient.kt
@@ -31,6 +31,7 @@ import com.vitorpamplona.quartz.nip01Core.relay.sockets.WebSocketListener
 import com.vitorpamplona.quartz.nip01Core.relay.sockets.WebsocketBuilder
 import com.vitorpamplona.quartz.utils.Log
 import com.vitorpamplona.quartz.utils.TimeUtils
+import kotlin.concurrent.Volatile
 import kotlin.concurrent.atomics.AtomicBoolean
 import kotlin.concurrent.atomics.ExperimentalAtomicApi
 import kotlin.coroutines.cancellation.CancellationException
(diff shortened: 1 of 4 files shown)


davotoula:
@vitorpamplona any concerns? Want to wait for after the release?

vitorpamplona:
Trying to understand and map the impacts of this right now...

davotoula:
To me it looks like a minor bug.

I noticed large amounts of failures on a specific relay. Back off wasn't working on that relay. 

There was a quirk if a relay accepts the connection the back off is reset but then the relay failed after that. Result was frequent errors from the relay.

davotoula:
Good relays unaffected.. Relays that fail to connect unaffected.

vitorpamplona:
One of the things we did was to reconnect as soon as possible to the relay of zap payments (one off) because some relays fail in the first connection but succeed in the second. So, for the zap to go there are two tries as fast as possible

vitorpamplona:
Cool! it all works!

davotoula:
I did a lot of manual testing with and without the fix to confirm it's effective

PR #3186, opened 2026-06-12
feat(desktop): unhealthy-relay review banner + popup

nrobi144:
## Summary

Surfaces relays unresponsive for 7+ days across the user's NIP-65 (10002), DM (10050), and Search (10007) relay lists on **Desktop**. A non-modal warm-pink banner appears above feed columns whenever the classifier finds anything; tapping it opens an anchored popup with one row per unhealthy relay and per-row **Remove** / **Open Dashboard** / **Snooze 7d** actions plus a banner-level **Snooze all 7d**. Blocked (10006) is excluded from detection but still included in the multi-list Remove action.

Scope is Desktop-only for v1. The `commons` module is platform-neutral and ready for Android to follow whenever someone picks it up.

## PoW
Screenshot 2026-06-11 at 18 16 05
Screenshot 2026-06-11 at 18 16 17

## What's in the diff

- **quartz**: `RelayStat` gains best-effort `lastConnectAt` + `lastIncomingAt` timestamps (epoch seconds, 0 = never observed). `RelayStats` listener pushes them on `onConnected` / `onIncomingMessage`. Durable history lives outside quartz in the commons store.
- **commons/relays/health/** (new package):
  - `classifyRelayHealth()` — pure function with v1 gates: first-run grace (7d after `firstScanAt`), offline grace (skip if no relay has responded anywhere in the window), Tor-mode skip (relay timing is intentionally lossy through Tor), per-relay snooze, blocked-only exclusion.
  - `RelayHealthStore` — account-scoped, supervised scope, 5s debounced persist, 60s ticker so snoozes expire without a write trigger.
  - `RelayHealthListener` wires the quartz lifecycle into the store.
  - `RelayHealthPersistence` interface (no expect/actual — single impl per platform via injection, easier to test and to swap backends).
  - `RelayListMutator` interface + `RelayRemovalResult` sealed type.
  - Shared `UnhealthyRelayBanner` (`errorContainer @ 50% alpha` — same convention as `ChessSyncBanner` / `ProfileBroadcastBanner`) and `UnhealthyRelayRow` (static outlined tag chips, no ripple).
- **Desktop wiring**:
  - `PreferencesRelayHealthPersistence` — `java.util.prefs.Preferences`, per-account via 8-char pubkey prefix (matches `DesktopAccountRelays` precedent).
[...]

PR #3221, opened 2026-06-15
fix: RelayHealthStore threading + sleep-resume socket recovery (follow-up to #3186)

nrobi144:
## Summary

Follow-up to #3186 — addresses the three unresolved threading concerns davotoula
raised on the merged PR, plus a related socket-recovery bug that surfaces after
the host machine sleeps for hours/days.

### `RelayHealthStore` threading fixes (commons)

- `schedulePersist()` wraps the blocking `persistence.save()` in `withContext(ioDispatcher)`.
  On Desktop the store is constructed with the composition `rememberCoroutineScope()`
  (Main / AWT EDT), so `prefs.flush()` was previously running on the UI thread.
- `close()` now snapshots state, cancels internal jobs, and fires the final save
  on a detached, self-cancelling `Dispatchers.IO` scope. Previously it did
  `runCatching { persistence.save(state.value) }` synchronously from
  `DisposableEffect.onDispose`, blocking the composition thread for the duration
  of the disk flush during every account switch.
- `persistJob` / `tickJob` are now `@Volatile`, and a `@Volatile closed` flag
  guards `schedulePersist` so it short-circuits after teardown. Resolves the
  plain-`var`-raced-between-relay-network-thread-and-composition-thread comment.
- `desktopApp/Main.kt` passes `Dispatchers.IO` to the store so persistence flushes
  land on an IO-bounded dispatcher instead of `Dispatchers.Default` (the default
  on commonMain, kept that way so iOS still compiles — `Dispatchers.IO` is
  internal on iOS Native in kotlinx-coroutines 1.11).

### `NostrClient.keepAliveJob` sleep/resume detection (quartz)

Cross-platform fix for: leave the app open, close the laptop lid for a few
days, open it again — offline banner is stuck, feeds don't refresh until the
process restarts.

Root cause: after a long OS suspend, OkHttp's `pingInterval` timer is paused,
the relay server has long since dropped the connection, but `BasicRelayClient`
still reports `isConnected() == true` and `needsToReconnect() == false` because
the socket reference is non-null. `keepAliveJob.reconnectIfNeedsTo()` is a
no-op for those relays until the next ping fails — which can be a while.

Fix: track wall-clock overshoot of the scheduled tick. If `delay(60s)` actually
returned more than `5 * KEEP_ALIVE_INTERVAL_MS` (5 min) later than expected,
treat the host as having resumed from sleep and force `relayPool.disconnect()
+ connect()` instead of the conservative `reconnectIfNeedsTo` path. Threshold
is constant + tunable; 5x is conservative enough to avoid firing on GC stalls
or brief scheduler pauses.

## Test plan

- [x] `./gradlew :commons:compileKotlinIosArm64 :quartz:compileKotlinIosArm64` — green
- [x] `./gradlew :commons:compileKotlinJvm :quartz:compileKotlinJvm :desktopApp:compileKotlin` — green
- [x] `./gradlew :commons:verifyKmpPurity :quartz:verifyKmpPurity` — green
- [x] Added `RelayHealthStoreCloseTest` covering close-idempotency, post-close
      no-op schedulePersist, and internal-scope ownership

Sleep/resume detection isn't directly unit-testable (it's about wall-clock
delta inside an `infinite while delay` loop); reviewed by inspection and the
constant is over-conservative on purpose.

## Interop suites

## AI assistance

## License

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/relays/health/RelayHealthStore.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/relays/health/RelayHealthStore.kt
@@ -37,6 +37,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import kotlin.concurrent.Volatile
 
 /**
  * Per-account, durable record of relay liveness used to drive the "unhealthy relay" review UI.
@@ -55,6 +56,10 @@ class RelayHealthStore(
     private val persistence: RelayHealthPersistence,
     private val torEnabledProvider: () -> Boolean = { false },
     parentScope: CoroutineScope? = null,
+    // Caller-supplied dispatcher used for both classification and persistence I/O.
+    // Default is `Dispatchers.Default` so commonMain stays iOS-compatible — JVM hosts
+    // (Android/Desktop) should pass `Dispatchers.IO` so `prefs.flush()` doesn't sit on
+    // a CPU-bound worker.
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.Default,
 ) {
     companion object {
@@ -84,8 +89,11 @@ class RelayHealthStore(
     private val _unhealthy = MutableStateFlow<PersistentList<UnhealthyRelay>>(persistentListOf())
     val unhealthy: StateFlow<PersistentList<UnhealthyRelay>> = _unhealthy.asStateFlow()
 
-    private var persistJob: Job? = null
-    private var tickJob: Job? = null
+    @Volatile private var persistJob: Job? = null
+
+    @Volatile private var tickJob: Job? = null
+
+    @Volatile private var closed = false
 
     init {
         // Persist the firstScanAt seed if we just stamped it.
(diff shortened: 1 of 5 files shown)


vitorpamplona:
We should probably move the reconnect loop out of Quartz's NostrClient and into the app space since we can capture sleep/resume states differently between desktop and mobile. Then you can call 
```
NostrClient.reconnect(
    onlyIfChanged = false,
    ignoreRetryDelays = true,
)
```

Or am I missing something? 

If you can get the sleep/wake up events from the computer, all you need to do is call: disconnect(), connect()

nrobi144:
> We should probably move the reconnect loop out of Quartz's NostrClient and into the app space since we can capture sleep/resume states differently between desktop and mobile. Then you can call
> 
> ```
> NostrClient.reconnect(
>     onlyIfChanged = false,
>     ignoreRetryDelays = true,
> )
> ```
> 
> Or am I missing something?
> 
> If you can get the sleep/wake up events from the computer, all you need to do is call: disconnect(), connect()

makes sense, will investigate 👍

nrobi144:
Done in 552540e77: reverted the keep-alive change in NostrClient and moved the wake heuristic to `desktopApp/network/SleepResumeMonitor.kt`. `Main.kt` runs it next to the metrics LaunchedEffect and calls `relayManager.client.reconnect(onlyIfChanged = false, ignoreRetryDelays = true)` on detected wake.

Real OS sleep events (NSWorkspace / D-Bus PrepareForSleep / WM_POWERBROADCAST) can layer in later without touching Quartz.

PR #3240, opened 2026-06-17
Hide Android system bars in immersive scroll and full-screen media

davotoula:
Not sure why we never hid android system bar when watching full screen video, it always annoyed me 😆 Also hide system bar when immersive scrolling is on in UI preferences (I've left the navigation bar when scrolling, happy to remove if it's better UX).

## What
  
  Two places where the Android OS system bars used to stay painted now go immersive:
  
  1. **Scroll immersive mode** — when the in-app top/bottom chrome slides away on scroll, the OS **status bar** now hides too, and reappears the moment the chrome comes back. Only the top status bar is touched; the bottom OS
  navigation bar is left alone so gesture nav is unaffected.
  2. **Full-screen media** — opening the full-screen image/video viewer now hides **both** OS bars for the duration, with swipe-to-reveal, restoring them on close.
  
  ## Why
  
  The app is edge-to-edge, so the OS bars are drawn permanently over content. During immersive scroll the status bar lingered over the feed, and full-screen video/images played underneath a visible status + navigation bar.
  Neither was ever programmatically hidden — there has never been a `WindowInsetsControllerCompat.hide(...)`, `SYSTEM_UI_FLAG_*`, Accompanist `SystemUiController`, or `FLAG_FULLSCREEN` anywhere in the app's history — so this adds that behavior for the first time rather than restoring it. 
  
  ## How
  
  `DisappearingScaffold` gains a small `ImmersiveStatusBarEffect` that drives the activity window's status-bar visibility from the bar collapse fraction via `WindowInsetsControllerCompat`. It is gated on the existing
  `allowBarHide` path, so it is naturally governed by the current "automatically hide navigation bars" setting (the bars only collapse when immersive scrolling is active).
  
  The status bar is binary (it cannot slide), so it toggles on settle: it hides only once the tracked in-app bar reaches the fully-collapsed edge. It tracks the **top** bar's collapse fraction — the top and bottom bars clamp
  to their own heights and so collapse at different rates, and keying off the top keeps the status bar in step with the chrome directly beneath it. On screens with no top bar it falls back to the bottom bar so the status bar
  still hides in sync with the bottom navigation. The bar is always restored on dispose, so leaving an immersive screen can never strand a hidden bar. 
  
  `ZoomableContentDialog` (the shared full-screen media dialog) gains a `DisposableEffect` on its own dialog window that hides both system bars with `BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE` and restores them when the dialog
  closes.
  
  No new dependencies, no new settings, and no changes to the platform-neutral scroll-math classes (`DisappearingBarState` / `DisappearingBarNestedScroll`).
  
  ## Testing
  
  Verified on a Pixel 9a (Android 16, `playBenchmark` build):
  
  - Scroll a feed down → status bar hides once the chrome settles; scroll up → it returns together. With "automatically hide navigation bars" off, the status bar stays visible while scrolling.
  - Full-screen a video and an image → both OS bars hide, an edge swipe peeks them transiently, exit restores them.
  - Background/foreground while immersed → status bar restored; navigating away never leaves a hidden bar.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ZoomableContentDialog.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/ZoomableContentDialog.kt
@@ -53,6 +53,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedButton
 import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
@@ -71,10 +72,13 @@ import androidx.compose.ui.layout.boundsInWindow
 import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.util.lerp
 import androidx.compose.ui.window.Dialog
 import androidx.compose.ui.window.DialogProperties
 import androidx.core.net.toUri
+import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsControllerCompat
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.google.accompanist.permissions.ExperimentalPermissionsApi
 import com.google.accompanist.permissions.isGranted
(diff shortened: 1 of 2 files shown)


vitorpamplona:
Will check later. But boes it work with the 3 button navigation?

davotoula:
> Will check later. But boes it work with the 3 button navigation?

I intentionally left the navigation bar during scrolling but removed it in media full screen (video and picture).

Works well with gesture navigation. Let me actually test it with 3 button navigation.

davotoula:
3 button navigation works well:

- It stays visible in immersive feed scroll so users can click back
- it is hidden in full screen image and video. When users tap video/image our back button becomes visible at tope of image/video.

PR #3241, opened 2026-06-17
Promote Kenyan Swahili to the base resource qualifier

davotoula:
two Swahili languages files were empty. Users with base Swahili selected in app got only English strings, never the translated versions.

This change promotes Kenyan Swahili to be base so users get that as minimum.

If this works there are other languages that can be simplified in a similar way...

## License

      original license header.


davotoula:
The better fix is to use Crowdin UI to copy (export and then import) content from Kenyan Swahili  to base Swahili but UI can be buggy.

davotoula:
The default setup should be base language with majority of translations and regional files with a few overrides.

Not sure why our Crowdin got created with empty bases and several identical translations in each regional file.

I'll keep investigating

PR #3260, opened 2026-06-18
fix(desktop): macOS forced re-login on cold boot — ProGuard strips java-keyring backend

nrobi144:
## Summary

The desktop release DMG's ProGuard config (`compose-rules.pro`) was keeping `pt.davidafsilva.apple.**` — a library no longer in the dependency graph. The actual macOS keychain dep is `com.github.javakeyring:java-keyring`, which reflection-loads its OS-specific backend at `Keyring.create()`. The shrink pass stripped the backend classes, so every cold boot of the release DMG threw `BackendNotSupportedException`, the `SecureKeyStorage` fallback silently returned null (no password prompt in a GUI session), and every account whose key lived in the OS keychain (nsec, NIP-46 bunker ephemeral, NWC secret) was kicked back to the login screen on each launch. Dev / `./gradlew :desktopApp:run` skips ProGuard, which is why it never surfaced in development. Bunker users were the loudest reporters because re-pasting a bunker URI is high-friction; nsec users likely re-pasted quickly and never reported.

This PR fixes the keep rules and adds a small diagnostic so a future regression in this area is visible to the user, not silent.

## What changed

**Primary fix (`desktopApp/compose-rules.pro`):**
- Replace dead `pt.davidafsilva.apple.**` keeps with `com.github.javakeyring.**`.
- Keep native methods + constructors on the `internal.**` backends so reflection-loading survives the shrink pass on macOS / Linux / Windows release artifacts.
- Updated the comment block to point at the real library and document the symptom.

**Defense in depth (so a future strip is visible):**
- `AccountManager._keychainUnavailable: StateFlow<Boolean>`, mirroring the existing `_storageCorruption` / `_forceLogoutReason` diagnostic channels.
- `loadInternalAccount` + `loadBunkerAccount` raise the signal when `accounts.json.enc` points at a key the keychain cannot return.
- `LoginScreen` shows a single-line error banner above the LoginCard when the signal is set. Cleared on any successful login (key / bunker / nostr-connect).

**Tests (`AccountManagerLoadAccountTest`):** four new cases — Internal-no-privkey signals, Bunker-no-ephemeral signals, `clearKeychainUnavailable()` resets, happy path does NOT signal.

## Test plan

- [x] Ran `./gradlew test` — full suite green locally
- [x] Ran `./gradlew :desktopApp:test --tests "*AccountManagerLoadAccountTest"` — new + existing cold-boot tests green
- [ ] **Manual repro on macOS release DMG — needs a reviewer with a Mac.** I do not have a Mac to package + install the signed DMG. Suggested verification:
  1. Build current `main` DMG, install, log in with a fresh bunker URI, quit, relaunch → expect the bug (login screen).
  2. Switch to this branch, rebuild DMG, repeat → expect bunker session restored.
  3. Sanity check `security find-generic-password -s amethyst-desktop` shows the entry after the first login (proves write-side works).

Same fix should incidentally cover Linux DEB / Windows MSI (the keep rule covers all `internal.**` backends), but I have not verified those either.

## Interop suites

## AI assistance

The brainstorm + plan that drove this fix is checked in at `docs/plans/2026-06-18-fix-desktop-macos-bunker-relogin-plan.md` (originally also a local-only `docs/brainstorms/2026-06-18-fix-macos-bunker-relogin-brainstorm.md` per the repo's gitignore). Deferred follow-ups are listed in the plan's Implementation Status section.

## License

- **What to monitor:** any new "forced re-login on macOS" reports after the next release.
- **Validation checks:** see Test plan above — primary check is the manual DMG cold-boot recipe.
- **Expected healthy behavior:** macOS users on the new release stay logged in across app restarts (bunker, nsec, NWC).
- **Failure signal / rollback trigger:** if a reviewer's Mac verification shows the cold-boot bug *still* reproduces on this branch, the prime hypothesis (ProGuard strip) was wrong and H2 (hardened-runtime entitlements) needs investigation per the plan doc.
- **Validation window:** one release cycle; reverting the ProGuard rule change alone is the rollback.

### DIFF
--- a/desktopApp/build.gradle.kts
+++ b/desktopApp/build.gradle.kts
@@ -1,6 +1,7 @@
 import de.undercouch.gradle.tasks.download.Download
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 import java.nio.file.Files
+import java.util.zip.ZipFile
 
 plugins {
     alias(libs.plugins.jetbrainsKotlinJvm)
@@ -272,4 +273,77 @@ val createReleaseAppImage by tasks.registering(Exec::class) {
     // Bypass FUSE requirement on CI runners (ubuntu-latest lacks libfuse.so.2).
     // AppImage standard env var: extracts + runs without mounting.
     environment("APPIMAGE_EXTRACT_AND_RUN", "1")
+}
+
+// ============================================================================
+// Native lib regression guard
+// ============================================================================
+// `pt.davidafsilva.apple:jkeychain` ships its `osxkeychain.so` native library
+// as a resource at the JAR root, and `OSXKeychain.loadSharedObject()` reaches
+// it with `getResourceAsStream("/osxkeychain.so")`. If a future ProGuard
+// upgrade, dep swap, or build-config change strips this resource from the
+// release distributable, every macOS user is forced back to the login screen
+// on every cold boot (the Keychain backend can't load, `SecureKeyStorage`
+// falls back to its password-prompted encrypted file, which never prompts
+// in a GUI cold boot → keys silently null → bunker / nsec / NWC unrecoverable).
+//
+// As of the current build the resource DOES survive ProGuard (it ships in the
+// proguarded jkeychain-1.1.0-*.jar with all 117 KB intact), so this task is
+// a regression guard, not a workaround. It's wired onto every release task so
+// it fails the build immediately if the .so disappears.
+val verifyJkeychainNativeSurvivesProguard by tasks.registering {
+    description = "Fail the release build if osxkeychain.so is stripped from proguarded output (would break macOS Keychain at runtime)"
+    group = "verification"
+    dependsOn("proguardReleaseJars")
+    val proguardDir = layout.buildDirectory.dir("compose/tmp/main-release/proguard")
+    inputs.dir(proguardDir).withPropertyName("proguardOutput").withPathSensitivity(PathSensitivity.RELATIVE)
+    val marker = layout.buildDirectory.file("verify-jkeychain-native.ok")
+    outputs.file(marker)
+    doLast {
+        val dir = proguardDir.get().asFile
+        require(dir.isDirectory) { "ProGuard output dir missing: $dir" }
+        val jars = dir.listFiles { f -> f.extension == "jar" } ?: emptyArray()
+        require(jars.isNotEmpty()) { "No proguarded jars under $dir" }
+        val foundIn = jars.firstOrNull { jar ->
+            val zip = ZipFile(jar)
+            try {
+                val entries = zip.entries()
+                var found = false
+                while (entries.hasMoreElements()) {
+                    if (entries.nextElement().name == "osxkeychain.so") {
+                        found = true
+                        break
+                    }
+                }
+                found
+            } finally {
+                zip.close()
+            }
+        }
+        require(foundIn != null) {
+            "REGRESSION: osxkeychain.so missing from every proguarded jar in $dir.\n" +
+                "OSXKeychain.loadSharedObject() calls getResourceAsStream(\"/osxkeychain.so\") — if this\n" +
+                "resource is not at classpath root in the release distributable, every macOS user is\n" +
+                "forced to re-log-in on every cold boot of the release DMG."
+        }
+        marker.get().asFile.writeText("verified osxkeychain.so survived ProGuard in ${foundIn.name}\n")
+        println("verifyJkeychainNativeSurvivesProguard: osxkeychain.so present in ${foundIn.name}")
+    }
+}
+
+// Gate every release-packaging task on the verification so a stripped .so
+// never reaches a release artifact.
+listOf(
+    "packageReleaseDmg",
+    "packageReleaseMsi",
+    "packageReleaseDeb",
+    "packageReleaseRpm",
+    "packageReleaseDistributionForCurrentOS",
+    "packageReleaseUberJarForCurrentOS",
+    "createReleaseDistributable",
+    "runReleaseDistributable",
+).forEach { name ->
+    tasks.matching { it.name == name }.configureEach {
+        dependsOn(verifyJkeychainNativeSurvivesProguard)
+    }
 }
--- a/desktopApp/compose-rules.pro
+++ b/desktopApp/compose-rules.pro
@@ -82,14 +82,12 @@
 #   - androidx.sqlite (sqlite-bundled): nativeThreadSafeMode() stripped,
 #     so `BundledSQLiteDriver.threadingMode` threw NoSuchMethodError on the
 #     first relay-store query (LocalRelayStore.refreshStats).
-#   - com.github.javakeyring (java-keyring): Keyring.create() reflection-loads
-#     the OS-specific backend (OSXKeychainBackend / SecretServiceBackend /
-#     WinCredentialStoreBackend). Without these keeps the macOS backend was
-#     stripped, BackendNotSupportedException fired on every cold boot, the
-#     SecureKeyStorage fallback silently returned null, and any account whose
-#     key lived in the keychain (nsec, NIP-46 bunker ephemeral, NWC secret)
-#     was forced to re-log-in every launch of the release DMG. The dev/Gradle
-#     run path skips ProGuard, which is why it never surfaced in development.
+#   - pt.davidafsilva.apple (jkeychain): macOS Keychain JNI for nsec storage.
+#     Note: this lives in the dependency graph as a transitive runtime dep of
+#     `com.github.javakeyring:java-keyring` — `ModernOsxKeychainBackend` holds
+#     a `private pt.davidafsilva.apple.OSXKeychain` field. Keep both the
+#     class and its private native methods (loadSharedObject / the
+#     _addGenericPassword / _findGenericPassword bridge).
 #
 # secp256k1-kmp is already covered by the `-keep class fr.acinq.secp256k1.**`
 # rule mirrored from mobile above.
@@ -98,10 +96,9 @@
     native <methods>;
     static <methods>;
 }
--keep class com.github.javakeyring.** { *; }
--keepclassmembers class com.github.javakeyring.internal.** {
+-keep class pt.davidafsilva.apple.** { *; }
+-keepclassmembers class pt.davidafsilva.apple.** {
     native <methods>;
-    <init>(...);
 }
 
 # kmp-tor — loads native Tor daemon via JNI reflection
--- a/desktopApp/compose-rules.pro
+++ b/desktopApp/compose-rules.pro
@@ -82,7 +82,14 @@
 #   - androidx.sqlite (sqlite-bundled): nativeThreadSafeMode() stripped,
 #     so `BundledSQLiteDriver.threadingMode` threw NoSuchMethodError on the
 #     first relay-store query (LocalRelayStore.refreshStats).
-#   - pt.davidafsilva.apple (jkeychain): macOS Keychain JNI for nsec storage.
+#   - com.github.javakeyring (java-keyring): Keyring.create() reflection-loads
+#     the OS-specific backend (OSXKeychainBackend / SecretServiceBackend /
+#     WinCredentialStoreBackend). Without these keeps the macOS backend was
+#     stripped, BackendNotSupportedException fired on every cold boot, the
+#     SecureKeyStorage fallback silently returned null, and any account whose
+#     key lived in the keychain (nsec, NIP-46 bunker ephemeral, NWC secret)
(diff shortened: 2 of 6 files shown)


nrobi144:
## Update — prime hypothesis refuted by binary PoW

Built `./gradlew :desktopApp:proguardReleaseJars` on both `main` (unfixed) and this branch (fixed) and inspected `java-keyring-1.0.4-*.jar` in `desktopApp/build/compose/tmp/main-release/proguard/`.

**Result: both branches contain identical keychain backend bytecode.** The strip hypothesis (H1 in the linked plan) was wrong.

### `main` proguarded `java-keyring.jar` contents (relevant subset)
```
com/github/javakeyring/internal/osx/OsxKeychainBackend.class           (4631)
com/github/javakeyring/internal/osx/ModernOsxKeychainBackend.class     (4009)
com/github/javakeyring/internal/osx/SecurityLibrary.class               (815)
com/github/javakeyring/internal/osx/CoreFoundationLibrary.class         (354)
com/github/javakeyring/internal/osx/NativeLibraryManager.class         (1417)
com/github/javakeyring/internal/KeyringBackendFactory.class            (2043)
com/github/javakeyring/internal/freedesktop/FreedesktopKeyringBackend  (4542)
com/github/javakeyring/internal/kde/KWalletBackend.class               (3840)
com/github/javakeyring/internal/windows/WinCredentialStoreBackend.class(5107)
pt/davidafsilva/apple/OSXKeychain.class                               (5158)
```

### This branch's proguarded `java-keyring.jar` (same subset)
**Identical bytes for the keychain operational path.** `OSXKeychainException.class` is 541b vs 764b on main (small lambda stripped) but `OSXKeychain.class` and `ModernOsxKeychainBackend.class` are byte-identical, and `javap -p` shows all public methods + private native methods (`_addGenericPassword` / `_findGenericPassword` / `_deleteGenericPassword` / `loadSharedObject`) are intact on both.

### What this means

1. **The ProGuard rule change in this PR is effectively a no-op** for the reported bug. ProGuard is preserving the macOS Keychain backend regardless of which keep rule label we use, presumably via transitive reachability + the global `-keepnames class ** { *; }`.
2. **The compose-rules.pro comment was misleading me, not actually mis-targeted.** `pt.davidafsilva.apple` IS a transitive runtime dep — `ModernOsxKeychainBackend` has a `private pt.davidafsilva.apple.OSXKeychain keychain` field — so the original keep rule wasn't dead code, it was correctly keeping the underlying JNI bridge.
3. **The user-reported "re-login every cold boot" must have a different root cause.** Strongest remaining candidates:
   - **H2 (hardened runtime / dylib signing).** `OSXKeychain.loadSharedObject()` unpacks `libosxkeychain.dylib` from the jar to `/tmp` and `System.load()`s it. macOS Hardened Runtime + a notarized DMG **rejects loading an unsigned dylib at runtime** unless `com.apple.security.cs.disable-library-validation` is in the app's entitlements. Dev `:desktopApp:run` doesn't apply hardened runtime, which would explain the dev-works / release-fails split.
   - **H4 (new): jpackage strips the dylib resource.** Worth checking whether the DMG actually contains `libosxkeychain.dylib` at all in its packaged jar (compose-multiplatform's jpackage path sometimes flattens jar resources).

### What's worth keeping in this PR vs reverting

- 🟡 **ProGuard rule change** — neutral / no-op for the bug. Could be reverted to keep the original (correct) `pt.davidafsilva.apple.**` keep, or kept since both produce equivalent output. Mild preference for reverting so the comment block stays accurate.
- 🟢 **Defense-in-depth keychain-unavailable signal + LoginScreen banner** — still useful regardless of root cause. Whenever the bug recurs (this version of it or a future one), the user gets a hint instead of a silent login screen.
- 🟢 **Plan doc** — still useful as the investigation record.

I'll wait for guidance before pushing more — likely move is to revert the ProGuard change, keep the diagnostic banner, and open a follow-up issue to investigate H2 (hardened runtime / dylib signing) with a Mac reviewer.

nrobi144:
## Update 2 — second hypothesis (H1b: ProGuard strips osxkeychain.so) also refuted; H2 too

Continued the PoW dig today and was wrong twice more before getting clear data. Posting the full trace so the next person doesn't repeat the steps.

### What I tested

| # | Hypothesis | Test | Result |
|---|------------|------|--------|
| H1b | ProGuard strips the `osxkeychain.so` native resource | `unzip -l` the proguarded jars *all* of them, not just the keyring-named one | **REFUTED** — `osxkeychain.so` ships intact (117 016 bytes) in proguarded `jkeychain-1.1.0-*.jar`. My earlier inspection missed this jar because my grep pattern only matched `*keyring*` |
| — | Real-world `Keyring.create()` + `setPassword` + `getPassword` round-trip against the proguarded classpath | `javac` + `java -cp <proguarded jkeychain + java-keyring jars>` of a `KeychainTest.main` | **PASS** — round-trip succeeds, value returns intact, deletePassword OK |
| H2 | Hardened runtime + Library Validation blocks the unsigned `osxkeychain.so` at `System.load()` time | `codesign -d --entitlements -` on a locally built `Amethyst.app` | **REFUTED** — `com.apple.security.cs.disable-library-validation = true` is in the default entitlements compose-multiplatform applies |
| — | Locally built release distributable launches without keychain errors | Foreground `./Amethyst.app/Contents/MacOS/Amethyst` for 10s, capture stdout/stderr | **OK** — only unrelated VLC plugin-cache warnings, no keychain noise |

### What this means

I cannot reproduce the user-reported bug on a locally built (`./gradlew :desktopApp:createReleaseDistributable`, adhoc-signed, hardened-runtime + disable-library-validation enabled) release distributable on this Apple-Silicon Mac. The keychain stack works at every layer I can poke from here.

Remaining possibilities — none of which I can verify alone:

1. **CI-built signed/notarized DMG differs from a local build** in a way that breaks the keychain only when distributed. Notarization doesn't typically change entitlements but does run additional Apple-side scans. Would need a Mac reviewer with a fresh macOS account to install the released DMG and observe.
2. **Affected user's environment is unique** — a specific macOS version, a stale Keychain ACL from a prior code-signing identity, a quarantine xattr issue, or an upgrade-path corruption of `accounts.json.enc`. Need the affected user's `Console.app` log filtered on "Amethyst" while they hit the bug + an `ls -la ~/.amethyst/` + a `security find-generic-password -s amethyst-desktop`.
3. **Bug is elsewhere on the cold-boot path**, not keychain-related at all (e.g. `accounts.json.enc` decryption silently failing for bunker entries while other types work). Would need actual repro to chase.

### What's in this branch now

- ✅ `AccountManager._keychainUnavailable` + LoginScreen diagnostic banner + 4 unit tests — **kept**. Useful regardless of which failure mode eventually turns out to be the cause. Right now any cold boot where `accounts.json.enc` points at a bunker / nsec account but the keychain returns null will get a visible banner instead of silently dropping to login screen.
- ✅ `verifyJkeychainNativeSurvivesProguard` Gradle task — **added**. Wired onto every release-packaging task. Scans proguarded jars for `osxkeychain.so`; fails the release build if it disappears. Cheap, and would have caught H1b if it had been real. This is the test from the "we want to fix the tests to catch these too" ask, even though it guards against a hypothetical class of bug, not the one we're actually hunting.
- ❌ ProGuard keep-rule change — reverted in the previous commit.

### Recommended next step

I'd appreciate one of:
- (Best) Affected user's `Console.app` log lines during a failing cold boot + `ls -la ~/.amethyst/` + `security find-generic-password -s amethyst-desktop` output. With those three artifacts I can probably point at the actual cause.
- (Failing that) A Mac reviewer who can install the most recent *released* DMG on a clean macOS user, reproduce, and capture the same three artifacts.

Happy to keep digging once we have any of that.

nrobi144:
## Update 3 — actual repro attempt on the released DMG (v1.12.1, arm64)

Downloaded `amethyst-desktop-1.12.1-macos-arm64.dmg` from the GitHub release page, mounted, and observed what happens when the bundled Amethyst launcher runs `OSXKeychain.loadSharedObject()`.

### Setup differences I noticed

```bash
$ codesign -d --verbose=4 /Volumes/Amethyst/Amethyst.app | grep flags
CodeDirectory v=20400 size=1659 flags=0x2(adhoc) hashes=45+3 location=embedded

$ codesign -d --entitlements - /Volumes/Amethyst/Amethyst.app
(empty — no entitlements)
```

vs. a locally built `./gradlew :desktopApp:createReleaseDistributable` on my macOS 26.5:

```
flags=0x10002(adhoc,runtime)        # hardened runtime ON
# entitlements include com.apple.security.cs.disable-library-validation = true
```

Different `codesign` tool versions between CI (`macos-14` runner) and my local Mac apparently emit different defaults. So the released DMG ships **without** hardened runtime AND **without** any entitlements file. The local distributable ships **with** both. That's a real environmental difference between what users get and what a maintainer building locally sees.

### Actual repro — AMFI does log a rejection

Launched the released `Amethyst.app/Contents/MacOS/Amethyst` directly so it would write to the unified log. On every launch of the released bundle on macOS 26.5:

```
kernel: AMFI: '/private/var/folders/_r/.../T/osxkeychain<rand>.so' has no CMS blob?
kernel: AMFI: '/private/var/folders/_r/.../T/osxkeychain<rand>.so': Unrecoverable CT signature issue, bailing out.
kernel: AMFI: code signature validation failed.
```

Reproduced four times across four separate launches, each with a different `<rand>` temp filename. This is the temp file produced by `OSXKeychain.loadSharedObject()` (it `createTempFile("osxkeychain", ".so")` then copies the unsigned jar resource).

### …but on macOS 26.5 the rejection doesn't actually block

To test whether the AMFI log corresponds to a real load failure, I redirected the bundled jpackage launcher to a tiny `KeychainTestStdout.main` instead of `MainKt` (modified `Amethyst.app/Contents/app/Amethyst.cfg`, re-signed with `codesign --force --sign - --deep`). Two variants:

1. **Original released `jkeychain-1.1.0-*.jar`** (unsigned `osxkeychain.so` at root)
2. **Modified jar** where I extracted `osxkeychain.so`, `codesign --sign -`'d it ad-hoc, and re-packed at the same path

Both launched via the bundled `Amethyst` binary, ran end-to-end inside the jpackage runtime, and wrote a result file:

```
=== unsigned .so (released as-is) ===
[pow] Keyring.create() OK: com.github.javakeyring.Keyring
[pow] setPassword OK
[pow] getPassword=OK
[pow] deletePassword OK
[pow] RESULT=PASS
+ AMFI: "Unrecoverable CT signature issue, bailing out. code signature validation failed."

=== pre-signed .so ===
[pow] Keyring.create() OK: com.github.javakeyring.Keyring
[pow] setPassword OK
[pow] getPassword=OK
[pow] deletePassword OK
[pow] RESULT=PASS
+ no AMFI rejection of osxkeychain.so
```

On macOS 26.5, AMFI logs the rejection but `dlopen()` succeeds anyway and the keychain operation works. The pre-signed `.so` variant only suppresses the log noise; it does not change the user-visible outcome on this OS version.

### Where this leaves us

I still **cannot reproduce the user-visible "every cold boot, bunker forgotten" bug** on this Apple-Silicon macOS 26.5 machine, even running the exact released DMG with the bundled jpackage runtime end-to-end. Strongest remaining possibilities:

1. **Stricter AMFI enforcement on the affected user's macOS version.** Older macOS (Sonoma 14, Sequoia 15) may actually *block* `dlopen` when AMFI reports the CT signature issue, whereas Tahoe 26 just logs. Pre-signing the `.so` (or shipping the JNI bridge as a properly signed bundled dylib instead of jar-extract-to-tmp) would be the fix.
2. **Bug isn't keychain-related at all.** Could be `DesktopAccountStorage.currentAccount()` returning null on cold boot for a reason unrelated to the OS keychain — `accounts.json.enc` decryption failing, master key not in keychain, etc.
3. **Bug is environmental on the user's box.** Old Keychain ACL pinned to a previous signing identity, quarantine xattr on `~/.amethyst/`, etc.

### What I need from the affected user to close this

Three artifacts (any subset would help — all three would be ideal):
- `sw_vers` output (macOS version + build)
- `log show --predicate 'process == "Amethyst"' --last 5m --info` while they hit the bug
- `ls -la ~/.amethyst/` + `security find-generic-password -s amethyst-desktop -g 2>&1 | head -20` immediately after first bunker login

### What's in this branch right now

- ✅ `AccountManager._keychainUnavailable` signal + `LoginScreen` banner + 4 unit tests
- ✅ Gradle `verifyJkeychainNativeSurvivesProguard` regression guard (also confirmed `osxkeychain.so` survives ProGuard byte-for-byte)
- ❌ ProGuard rule swap reverted

Reopening the PR. Even if the keychain hypothesis is wrong, the diagnostic banner + regression guard are worth keeping; both fall out cleanly if a maintainer decides otherwise.

PR #3297, opened 2026-06-19
Fix disappearing bar overshooting on fast reveal fling

vitorpamplona:
## Summary

Fixed a bug where the disappearing bar (top and bottom height offsets) would overshoot its visible resting edge (0) when settling from a fast reveal fling. The critically-damped spring animation would cross its target once when given an initial velocity in the target's direction, causing the bar to briefly render past its resting position before springing back.

The fix adds bounds clamping to the `Animatable` in `animateOne()`, constraining offsets to the valid travel range `[-limit, 0]`. When the animation hits a bound, it ends crisply at the edge with no rebound.

## Test plan

- [x] Added `DisappearingBarStateTest` with a test case that reproduces the bug: a fast reveal fling (12000f velocity) from mid-collapse settles at the visible edge without overshooting
- [x] Ran `./gradlew test` — new test passes with the fix, would fail without it

The test uses a manually-stepped frame clock to sample both offsets after every frame and verify the peak values never exceed the visible edge (0), confirming no overshoot occurs.

## Interop suites

- [x] N/A — change is UI-only, affects layout animation bounds, no wire bytes or protocol state

## License

https://claude.ai/code/session_01EEGCrB5uRBAZES1PSp4Ctz

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/layouts/DisappearingBarState.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/layouts/DisappearingBarState.kt
@@ -86,8 +86,8 @@ class DisappearingBarState(
      */
     suspend fun resetToVisible() {
         coroutineScope {
-            launch { animateOne({ topHeightOffset }, 0f, 0f) { topHeightOffset = it } }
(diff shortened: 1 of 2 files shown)


davotoula:
Ooo.. This was highlighted as a risk when I implemented it but I didn't see the issue during manual testing. 

Nice one!

PR #3306, opened 2026-06-20
NIP-22: Reply to Amethyst threads as kind 1111 Comments

vitorpamplona:
## Summary

Implements NIP-22 Comment support for replies to Amethyst-authored thread roots. When replying to a brand-new kind-1 thread posted by Amethyst, replies are now sent as kind 1111 Comments instead of kind 1 replies. This improves thread organization by keeping comment threads separate from the main feed.

The change adds:
1. `isClient()` extension functions to check if an event was published by a specific client (NIP-89, case-insensitive)
2. Logic in Android and Desktop compose screens to detect Amethyst-authored thread roots and use `CommentEvent.replyBuilder()` instead of `TextNoteEvent.build()`
3. Notification reply handling that respects the same NIP-22 logic
4. Comprehensive unit tests for the new `isClient()` functionality

## Test plan

- [x] Ran `./gradlew test` — all tests pass, including new `IsClientTest` covering:
  - Matching client tags by name (case-insensitive)
  - Matching with additional tag fields (address, relay hint)
  - Non-matching scenarios (different client, missing tag, empty name)
- [x] Manually exercised the change:
  - Verified replies to new Amethyst threads produce kind 1111 events
  - Verified replies to non-Amethyst threads still produce kind 1 events
  - Verified replies to non-root threads still produce kind 1 events
  - Tested on both Android and Desktop platforms

## Interop suites

- [x] N/A — change affects event kind selection (kind 1 vs 1111) which is covered by existing NIP-22 protocol tests; wire format is handled by `CommentEvent` which is already tested

## License

https://claude.ai/code/session_01V7RyevA6jL1NuY7uev2agS

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/notifications/NotificationReplyReceiver.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/notifications/NotificationReplyReceiver.kt
@@ -29,11 +29,13 @@ import androidx.core.content.ContextCompat
 import com.vitorpamplona.amethyst.Amethyst
 import com.vitorpamplona.amethyst.LocalPreferences
 import com.vitorpamplona.amethyst.model.LocalCache
+import com.vitorpamplona.amethyst.model.accountsCache.AccountCacheState
 import com.vitorpamplona.quartz.nip01Core.hints.EventHintBundle
 import com.vitorpamplona.quartz.nip01Core.tags.people.PTag
 import com.vitorpamplona.quartz.nip10Notes.TextNoteEvent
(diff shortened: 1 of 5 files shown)


davotoula:
Sounds like a big change? Any backwards compatibility issues with other clients.. Or older versions of amethyst?

vitorpamplona:
Not that much. We have been showing kind 1111s as replies for more than a year now. So, all Amethyst users will see it. The old ones might still reply with kind 1, which we still render as usual. 

The only issue is that clients that don't display kind 1111 comments won't see the reply. But both the thread owner and the author of the reply are already in the new standard, so it should be good.

PR #3328, opened 2026-06-21
Add bidirectional local relay event sync

### LINKED ISSUE #2706: [FEATURE] Sync to local relay / Export database
Some people may spin up a local relay (citrine) in order to have their own archive. It would be nice to be able to include this relay during a relay sync. So that every event, outbox + inbox + dm, would be synced to it.

An alternative solution would be to allow exporting the entire amethyst database.

Bounty: 13,000 Sats

silentgeckoaudit3801:
## Summary

- include configured local relays as both event-sync sources and archive destinations
- copy authored, mentioned, and DM events to local relays while excluding the current source to prevent echo writes
- restore missing local-relay events through the existing outbox, inbox, and DM routing paths
- expose local archive send/accept counters in the sync UI

## Performance and safety

This reuses the existing paginated, 50-relay concurrency-limited sync stream. It does not load or serialize the whole cache, and it uses a dedicated event-id dedup set for local destinations. A local source is excluded from its own destination set, avoiding a write-back loop during large (~20,000 event) syncs.

## Validation

- updated every in-tree `EventSync` constructor call
- updated previews for the extended live-activity model
- source-level consistency checks completed; a full Android build was not available in the contribution environment
- fork CI spotlessCheck, :quartz:verifyKmpPurity, and :commons:verifyKmpPurity passed: https://github.com/silentgeckoaudit3801/amethyst/actions/runs/27917465017/jobs/82605215268

Fixes #2706

### DIFF
--- a/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/relays/eventsync/EventSyncTest.kt
+++ b/amethyst/src/androidTest/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/relays/eventsync/EventSyncTest.kt
@@ -67,6 +67,7 @@ class EventSyncTest {
                     outboxTargets = { setOf(vitor) },
                     inboxTargets = { setOf(vitor) },
                     dmTargets = { setOf(vitor) },
+                    localTargets = { emptySet() },
                     clientBuilder = {
                         NostrClient(socketBuilder, appScope)
                     },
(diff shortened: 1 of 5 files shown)


silentgeckoaudit3801:
Thanks for taking a look. Could you clarify whether the closure is because this approach is not desired, because a large-relay performance test is required first, or because the bounty is no longer available? I will not reopen it without maintainer direction, but I can revise the implementation or add a focused performance harness. The fork CI lint and KMP purity checks passed: https://github.com/silentgeckoaudit3801/amethyst/actions/runs/27917465017/jobs/82605215268

vitorpamplona:
You didn't test on a device. We make that clear on our documentation for contributors.

PR #3332, opened 2026-06-22
Missing notifications on fresh install.

solarizeddev:
I opened a claude session to debug why I can not find notifications on fresh install, so claude code came up with these two fixes:

> **1. Don't cap notifications to the last 7 days.** Every cold start, the fetchers fall back to `since = oneWeekAgo()`. With the EOSE watermark not persisted, no event DB, and no backward pagination, that 7-day window is the *only* history ever requested — so Amethyst shows far fewer notifications than clients querying the same relays with just a `limit`. Reproduces in Global mode. Fix: drop the `oneWeekAgo()` fallback in `AccountNotificationsEoseFrom{Inbox,Random}RelaysManager` so a null `since` lets `limit` govern; the watermark still drives the live tail after EOSE.

>**2. Show reactions/reposts before their parent loads.** In Curated mode, `tagsAnEventByUser` only accepts a reaction/repost when the target note is already cached (`replyTo.last().author == me`). On cold start the parent often isn't loaded, so the interaction is dropped — and which ones drop differs every launch. Fix: accept on the reaction's NIP-25 `p` tag / repost's NIP-18 `p` tag (which name the target author), keeping the cached-parent check as fallback. Adds unit tests.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/DebugUtils.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/DebugUtils.kt
@@ -24,6 +24,7 @@ import android.app.ActivityManager
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.os.Debug
+import android.os.Trace
 import androidx.core.content.getSystemService
 import com.vitorpamplona.amethyst.model.LocalCache
 import com.vitorpamplona.quartz.nip01Core.core.Event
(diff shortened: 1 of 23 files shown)


solarizeddev:
Also wondering if notifications should be cached on-disk instead of only in memory. Today the feed is rebuilt from relays on every cold start, so anything a relay has pruned is gone for good.

vitorpamplona:
hum.. we might need to strike a balance here because if we load ALL notifications like the solution provides, that will take many minutes for some accounts. 

On 2, the Curated feed filter should not change. There is a new option for a true Global feed where you can see reactions on other people's posts that may cite you. 

On the disk comment, yes, but I still need to figure out how to delete things the user doesn't care anymore, since for most users, the disk space for events alone could get over 4GB easily.

solarizeddev:
Right. Claude proposes:

- Backward pagination: load a small recent page on cold start, fetch older pages on scroll.
- A notification only store that persists only the events behind notification cards, capped by count/age with LRU eviction, instead of the whole event graph.

I can close and you take it from here or I can dig deeper with claude, up to you.

solarizeddev:
Anyway If you want to have a look at claude's proposal:

https://github.com/solarizeddev/amethyst/commits/claude/notif-local-store/

PR #3401, opened 2026-06-27
fix(video): show browser-fallback overlay when a decoder silently stalls

vitorpamplona:
## Problem

A reported video (`nostr:nevent1qqspntv53znsnh0jvpkx39yeqrckjhkdevjqmpkpw99fe679lkzwh2spzpmhxue69uhkummnw3ezumt0d5hsygy0k9qtf6xaa7tuujuzr5j8y79p5s6nxcnz8ajqy9yykde0jjqqpspsgqqqqqqsspwkgx`) never plays — it sits at `00:00 / 10:48` forever on a blank/buffering box, with no error and no fallback.

**Root cause: it's HEVC/H.265 (`hvc1`, an iPhone/Damus upload).** Verified:
- File is fine — full 34 MB downloads in ~5s, `ffmpeg` decodes it end-to-end with zero errors, `moov` is at the front (streamable).
- The device decoder (the emulator's *software* HEVC decoder; also Brave on the desktop) can't keep up. It **silently parks the player in `STATE_BUFFERING`** — the buffer fills to the LoadControl cap (~11s), the playhead never leaves 0, and **no `PlaybackException` is ever raised**.
- `WatchPlaybackErrors` only listened for `onPlayerErrorChanged`, so the existing `RenderPlaybackError` "Open in browser" overlay never triggered.

## Fix

Add a decode-stall watchdog to `WatchPlaybackErrors` that polls the controller (1s) and synthesizes a `PlaybackException(ERROR_CODE_DECODING_FORMAT_UNSUPPORTED)` — which drives the **existing** overlay — once the player is, continuously for 8s:
- in `STATE_BUFFERING` and `playWhenReady`,
- with **≥2s of media buffered ahead** of a **frozen** playhead.

The buffer-ahead guard distinguishes a hung decoder (buffer full, won't drain) from genuine **network starvation** (buffer depleted — never flagged), so slow connections keep showing the normal spinner and play when bytes arrive.

**Self-healing / "if it works, just play":**
- Overlay clears on the `STATE_READY` transition (any recovery passes through it; the player keeps decoding *under* the overlay — we never stop it).
- Belt-and-suspenders: the watchdog also drops the overlay the instant the playhead advances again, so a slow *device* that flashes the message then succeeds just plays.

Also: narrowed the recovery-clear to `STATE_READY` only (clearing on `STATE_BUFFERING` would wipe the synthetic error instantly), and clear on `onMediaItemTransition` so a pooled player starting a new video resets cleanly.

## Verification

Reproduced the stall on the emulator via the exact event, then confirmed the fix live — after ~8s the overlay shows:

> 🎥̶ **Can't play this video** — The codec or format isn't supported on this device (`ERROR_CODE_DECODING_FORMAT_UNSUPPORTED`). Try opening it in your browser. **[ Open in browser ]**

`:amethyst:compilePlayDebugKotlin` clean, `spotlessApply` run, pre-commit/pre-push hooks pass.

## Notes for review
- **Timing:** 8s before the fallback appears (`STALL_TIMEOUT_MS`) — deliberately conservative; trivially tunable.
- **Scope:** Android-only (`amethyst/`). Desktop already has its own `PlaybackErrorMessage` path; not touched here. A hung *live* HEVC decoder is also covered (buffer-ahead guard still protects normal live buffering) — flag if you'd rather exempt live.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/playback/composable/WatchPlaybackErrors.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/playback/composable/WatchPlaybackErrors.kt
@@ -20,10 +20,30 @@
  */
 package com.vitorpamplona.amethyst.service.playback.composable
 
+import android.os.SystemClock
+import androidx.annotation.OptIn
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.MutableState
+import androidx.media3.common.MediaItem
 import androidx.media3.common.PlaybackException
 import androidx.media3.common.Player
+import androidx.media3.common.util.UnstableApi
+import kotlinx.coroutines.delay
+
+// How often the decode-stall watchdog samples the controller's position/buffer.
+private const val STALL_POLL_INTERVAL_MS = 1_000L
+
+// The decoder is only considered "fed" once it has comfortably more than bufferForPlaybackMs
+// (750 ms in feedTunedLoadControl) of media ready ahead of the playhead. Below this we treat a
+// frozen playhead as ordinary network starvation, not a decode failure, and leave it alone.
+private const val STALL_MIN_BUFFER_AHEAD_MS = 2_000L
+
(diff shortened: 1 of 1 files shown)


davotoula:
I just noticed this on my pixel 1. Works as intended

PR #3371, opened 2026-06-25
Add and update translations across 50+ locales

vitorpamplona:
## Summary

This PR adds new string resources and updates existing translations across 50+ language locales in the Android app. The changes include additions to Portuguese (Portugal), Chinese (Singapore, Hong Kong, Mainland), Korean, Vietnamese, Turkish, Spanish, Japanese, Italian, Russian, Ukrainian, and many other languages. A small number of existing strings were also removed or modified in some locales (e.g., Greek, Arabic, Tamil, Esperanto, Hungarian, French, Dutch, German, Spanish variants, Czech, Swedish, Indonesian, Thai, and Hindi).

The bulk of the additions appear to be new feature strings (likely related to audio rooms, relay management, or other recent features) that have been translated by the community or translation service.

## Test plan

- [ ] N/A — change is translation-only; no code logic affected

The changes are isolated to `strings.xml` resource files across multiple locales. No functional code was modified, so no manual testing of app behavior is required. The Android build system will validate XML syntax during compilation.

## Interop suites

- [ ] N/A — change can't affect wire bytes / decoded audio / MLS state / DM envelopes

## License

- [ ] By submitting this PR, I agree to license my contribution under the MIT license. Any code I did not author personally carries its original license header.

https://claude.ai/code/session_01Hfm5Lf3FfwDgPn5oiMVqUo

PR #3461, opened 2026-07-03
Consolidate Czech onto the base values-cs qualifier

davotoula:
This is path/try towards consolidating Spanish later as well...

What

Czech now lives in a single values-cs file: a byte-identical promote of the latest Crowdin export from values-cs-rCZ (2905 strings). The regional dir is deleted, crowdin.yml maps Czech's export code to cs, and cs-CZ is dropped from locales_config.xml so the picker shows one "Čeština" entry.

Why

Crowdin treats Czech as base cs, but its Android export code is hardcoded to cs-rCZ, so the real translation kept landing in the regional dir while the base file froze in March 2025 — stale strings that still shipped via per-key fallback and duplicated the picker entry. The fix is a one-line languages_mapping override (cs: cs), the same mechanism already proven by sw-KE: sw. No Crowdin-side changes needed.

Strategy

This is the template for the rest of the catalog, with two cases:

**Single-region languages** (Czech, German, Japanese, …): one Crowdin-owned base file via a one-line mapping, no regional dirs, picker lists the base. Untranslated strings fall back to English instead of a stale local copy.

**Multi-region languages** (Spanish, French, Portuguese, Chinese, …): the widest-reach variant becomes the full base (e.g. Latin American Spanish → values-es); other variants stay as thin overrides holding only the strings that genuinely differ. Regional translators leave everything else untranslated, skip_untranslated_strings: true omits it on export, and Android's per-key fallback (regional → base → English) fills the rest. New source strings get translated once per language, not once per region.

Czech goes first as the cheapest end-to-end proof: repo-only, no Crowdin UI work, no base-choice decision.

Also fixes

Drops 38 translations of translatable="false" strings (search-keyword indexes, channel names, proper nouns) that #3371 had bulk-filled into the frozen values-cs.

Verification

New values-cs is byte-identical to the current Crowdin export — 2905 strings, no duplicate keys, zero translatable="false" leakage. After merge, the next Crowdin sync should update values-cs directly and must not recreate values-cs-rCZ; that round-trip confirms the pattern before repeating it elsewhere.

## License

      original license header.


davotoula:
This is the strategy I think we should move towards:

Single-region languages (Czech, German, Japanese, …): one Crowdin-owned base file via a one-line mapping, no regional dirs, picker lists the base. Untranslated strings fall back to English instead of a stale local copy.

Multi-region languages (Spanish, French, Portuguese, Chinese, …): the widest-reach variant becomes the full base (e.g. Latin American Spanish → values-es); other variants stay as thin overrides holding only the strings that genuinely differ. Regional translators leave everything else untranslated, skip_untranslated_strings: true omits it on export, and Android's per-key fallback (regional → base → English) fills the rest. New source strings get translated once per language, not once per region.

PR #3483, opened 2026-07-06
feat(desktop): Web-of-Trust score badges + shared index relays + amy wot verbs

nrobi144:
## Summary

Adds Web-of-Trust (WoT) score badges to Desktop's feed avatars and person-cards, exposes matching `amy wot get/list/sync` verbs so the CLI can compute and share the same graph, and unifies Desktop + `amy` on a shared, user-editable set of *index relays* for kind-0 / kind-3 lookups.

## Pow
Screenshot 2026-07-06 at 09 38 46
Screenshot 2026-07-06 at 09 40 13
Screenshot 2026-07-06 at 09 42 46
Screenshot 2026-07-06 at 09 51 21

**Behaviour**

- Small counter-chip on the bottom-right of stranger avatars showing how many people you follow follow *them*. Suppressed on self, followed authors, and NIP-51 spam-exempt keys. Tooltip on hover: *"N of the people you follow follow this person"*. `99+` overflow.
- Batch REQ fires once per account with a 2 s ready timeout — no per-avatar network chatter. Guardrail: accounts following ≥ 2000 pubkeys skip the batch entirely (`LocalWoTReady = true`, no badges).
- Reactive: following someone new triggers a `loadKind3Batched` for their kind-3 and increments credited pubkeys once it arrives; unfollow decrements symmetrically. Kind-3 churn from another client diffs correctly with no double-count.
- Same badge slot forwarded to `UserSearchCard` so search-result person cards get the trust cue too. `NewDmDialog` deliberately left un-badged (recipient picker isn't a "stranger" surface).
- `amy wot get <npub|hex>` / `amy wot list --threshold N --limit M [--json]` / `amy wot sync` — the CLI computes the same WoT map from the local event store and can top up kind-3 events over the network.
- New **Index Relays** editor on the *Relays → Configure* tab (buffered Save + Reset-to-defaults, matching the sibling Search / DM / Blocked editors). Persists to `Preferences.userRoot().node("com/vitorpamplona/amethyst/relays/index")` — same node `amy wot sync` reads, so both surfaces query the same relays. Falls back to a built-in 4-relay default when unset. Changes to the running Desktop coordinator take effect on next launch (documented in the section).

**Not in this PR (deferred)**

- Notification-tab avatars (custom 56 dp card, doesn't route through `NoteCard`).
- DM-recipient-picker badges in `NewDmDialog`.
- Feed / notification threshold gating (badges are display-only in v1).
- Live re-targeting of the coordinator's index-relay set without a relaunch.
- Adopting `commons/AmethystDefaults.DefaultIndexerRelayList` (Purple Pages / Coracle) as the default — deserves its own review because it's a user-visible behaviour change.

## Test plan

Platform: Desktop JVM, macOS 15 (arm64), window 1440×900. Also ran `./gradlew test` (full suite green) and `./gradlew :desktopApp:run` for manual verification.

- [x] Ran `./gradlew test` — full suite green
- [x] Manually exercised the change (see notes)

Notes:

- Cold start with a well-connected account (~350 follows): batch REQ observed once; badges landed on strangers within ~2 s, none on self or followed authors.
- Threading: kind-3 churn from another client (published via `amy send`) diffs the credited-pubkeys map without double-count; badge counts update within a frame.
- Guardrail: logged in with a mega-follow test account (> 2 000 follows) — `LocalWoTReady` flipped true immediately, no batch REQ, no badges rendered.
- Empty-graph account: `LocalWoTReady` flipped true after the 2 s fallback, no badges.
- Account switch: old scores dropped, new account's follow set drives a fresh WoT map, no leaked badges.
- Search screen: person-card badges render for strangers; still absent on follows/self.
- Index Relays editor: verified add / remove / Save / Reset-to-defaults / Enter-key add / invalid-URL error / "restart to apply" toast.
- Index Relays persistence: added a relay → closed app → relaunched → still there. Removed one → closed → relaunched → still gone.
- `amy wot get` / `wot list` / `wot sync` — all three verbs tested against a local event store; `wot sync` observed hitting the overridden relay set after configuring Index Relays in the Desktop UI.

Full manual testing sheet in `desktopApp/plans/2026-07-01-wot-score-manual-testing-sheet.md`.

## Interop suites

## AI assistance

## License


davotoula:
Findings (most severe first)

1. desktopApp/.../cache/DesktopLocalCache.kt:515 — accountPubkey race can leave the Following feed empty and let a follow action wipe the user's entire follow list. (CONFIRMED) consumeContactList now gates _followedUsers/lastContactListEvent on event.pubKey == accountPubkey, but records lastContactListByAuthor[pubkey] before that gate — and accountPubkey is only assigned in a MainContent LaunchedEffect (Main.kt:1542) that races the local-store hydration launched earlier on Dispatchers.IO (Main.kt:893). If hydration consumes the user's cached kind-3 first, the follow set is never populated, and the same event later arriving from a relay is rejected by the createdAt <= prev check. In that state, FeedScreen.kt:544 reads lastContactListEvent == null and FollowAction.follow calls ContactListEvent.createFromScratch(...) — publishing a fresh kind-3 that destroys the real follow list. Every updated test had to manually set accountPubkey before hydrate(), which is the tell: make it a constructor parameter or otherwise guarantee ordering.

2. commons/.../wot/WoTService.kt:171 — the ≥2000-follows guardrail is defeated; the batch REQ still fires and badges still render. (CONFIRMED) handleFollowSet assigns myFollows = newFollows before the MAX_FOLLOWS check and returns without resetting it, and Main.kt:1561 calls loadKind3Batched(follows) gated only on isNotEmpty(). Incoming kind-3s then pass handleKind3's follower !in myFollows gate and fully repopulate reverseIndex/_scores — exactly the CPU/memory blow-up the guardrail (and the PR description's "skip the batch entirely, no badges") promises to prevent. Fix: clear myFollows (or set a disabled flag) in the guard branch, and skip the batch in Main.kt when over the limit.

davotoula:
Affecting android app:

Tier 1 — in commons/commonMain (compiled into the Android app; bites on adoption)


commons/.../wot/WoTService.kt:171 — guardrail bypass. The root cause (myFollows assigned before the MAX_FOLLOWS check and never reset) is entirely in common code. Any Android wiring would inherit it, and Android is where mega-follow accounts and constrained devices make the CPU/memory blow-up most painful. The desktop half (no size gate before loadKind3Batched) would likely be copy-pasted into Android wiring too.


commons/.../assemblers/FeedMetadataCoordinator.kt:325 — no-retry queuedKind3Pubkeys. Common code. On Android this gets worse than desktop: flaky mobile connectivity makes the "relays unreachable at login → WoT silently empty for the session" trigger far more likely, and it also inherits the doze/network-switch churn.

commons/.../wot/WoTService.kt:149 (+ comments at 43–45) — recomposition storm and false "per-key isolation" comments. SnapshotStateMap semantics are identical on Android Compose; per-op withMutableSnapshot commits and whole-map invalidation would churn every visible badged avatar on much weaker hardware. The misleading comments are the real Android trap — a future Android integrator will trust them.


commons/.../wot/WoTService.kt (no close()/dispose() API) — lifecycle gap behind the desktop leak. The leak itself (finding 4) is desktop wiring, but the missing teardown API is a commons design gap: Android's account-switch flow would leak writer coroutines the same way unless it remembers to hand-roll scope cancellation.


commons/.../assemblers/FeedMetadataCoordinator.kt:293 — unsynchronized eoseReceived set. Common code, concurrent per-relay Dispatchers.IO callbacks exist on Android too. Low ceiling (worst case is waiting the full 5 s timeout), and the pattern is pre-existing — but the PR doubles its footprint in shared code.

vitorpamplona:
Need to stop using lists of relays and use outbox properly, otherwise these numbers will always be wrong. Kind 0 and 3 must be downloaded from the outbox relay (10002, write) of each user. Basically, find all 10002 events via index relays (purple pages, etc, then parse them all to find a list of relays per author, invert the map to get a list of authors per relay, then use that list to download posts, kind 0 and contact lists from each author. 

Also. Damus relay will be gone by the end of the month.

vitorpamplona:
Maybe this is a good time for me to implement Graperank on Amy so that we can have a larger WoT scoring service in the CLI... 🤔

nrobi144:
## Review round 1 — all comments addressed

### Bugs (davotoula, comments [1](https://github.com/vitorpamplona/amethyst/pull/3483#issuecomment-4892248528) + [2](https://github.com/vitorpamplona/amethyst/pull/3483#issuecomment-4892272000))

| # | Fix | Commit |
|---|---|---|
| 1 | **P0 `accountPubkey` race** in `DesktopLocalCache.consumeContactList` — reordered Main.kt so pubkey binds before hydration; belt-and-braces gate inside `consumeContactList`. Regression test hydrates a self kind-3 with pubkey unbound, then replays the same event after bind, and asserts the follow list populates. | `d8961c0d75` |
| 2 | **`MAX_FOLLOWS` guardrail bypass** — `handleFollowSet` now clears `myFollows` and flips a `disabled` StateFlow *before* returning; `handleKind3` early-returns while disabled; Main.kt short-circuits the batch. Guardrail auto-releases if the follow set later shrinks. | `5166216e2e` |
| 3 | **`WoTService.close()`** — implements `AutoCloseable`; ops channel closes, writer coroutine exits, post-close ops drop silently. Main.kt wires it via `DisposableEffect(iAccount)`. | `5166216e2e` |
| 4 | **Misleading `SnapshotStateMap` doc** — rewritten to accurately describe per-key observation as a `SnapshotStateMap` property, with `withMutableSnapshot` correctly attributed to *coalescing* an op's writes. | `5166216e2e` |
| 5 | **`queuedKind3Pubkeys` no-retry** — mark-on-EOSE instead of mark-on-send. Zero-EOSE timeout rolls out of the new `inFlightBatched*` set so the next call retries. | `5217035f94` |
| 6 | **`eoseReceived` race** — replaced the shared `MutableSet` with `BatchEoseGate`: EOSE notifications funnel through a `Channel`, a single consumer coroutine owns the `seen` set. KMP-safe. | `5217035f94` |

### Outbox model (@vitorpamplona [comment](https://github.com/vitorpamplona/amethyst/pull/3483#issuecomment-4892302009))

Reworked kind-3 fetching in `commons/commonMain` around `RelayListRecommendationProcessor`:

- **Phase 1 → Phase 2 → Phase 3** pipeline in a new `OutboxDispatcher` (`dddeae74b6`).
- Desktop `DesktopRelaySubscriptionsCoordinator` and `amy wot sync` wired onto it (`bb2a83c1fe`).
- Phase 2 batched-parallel — one `client.subscribe` fans out to every recommended relay concurrently, so 23 relays complete in ~4 s (per-relay timeout) instead of 23×4 s sequentially (`8bcebd886e`).
- Damus relay — confirmed no production default on this branch references `relay.damus.io`. Only common-test fixtures do.

### Proof of work — cold-start login trace

Debug log from `./gradlew :desktopApp:run` on a real account with 587 follows and 4 index relays:

\`\`\`
16:56:50.183  DEBUG [OutboxDispatcher] start authors=587 newKind3=587 cachedOutbox=0 toDiscover=587 indexRelays=4
16:56:51.002  DEBUG [OutboxDispatcher] phase1 done eosed=4/4 10002-events=621 discovered=509
16:56:51.006  DEBUG [OutboxDispatcher] phase2 recommendations=23 relays-with-work=23
16:56:55.016  DEBUG [OutboxDispatcher] phase2 done kind3=199 kind0=0
16:56:55.018  DEBUG [OutboxDispatcher] phase3 fallback authors=78 kinds=[3] relays=4
16:56:55.280  DEBUG [OutboxDispatcher] phase3 done kind3=47 kind0=0
16:56:55.281  DEBUG [WotOutbox] fetchKind3Only authors=587 covered=509 fallback=78 kind10002=621 kind3=246
\`\`\`

**Reading the trace:**
- Phase 1 (index-relay 10002 discovery): 4/4 relays EOSE'd in **819 ms**; 621 kind-10002 events, **509 of 587 followed users** (86.7 %) publish an outbox.
- Phase 2 (per-outbox-relay kind-3): 23 recommended relays covered in **4.01 s** — parallel, not sequential. 199 kind-3s.
- Phase 3 (index-relay fallback): remaining 78 authors get index-relay REQ, 47 more kind-3s in **262 ms**.
- Total pipeline: **~5.1 s** end-to-end. 246 kind-3 events ingested → WoTService fed. Badges rendered on stranger avatars after that.

### Test suite

\`\`\`
./gradlew :commons:jvmTest :desktopApp:test :cli:test :quartz:jvmTest
BUILD SUCCESSFUL
\`\`\`

New regression tests:
- 2 in `DesktopCachePipelineTest` — accountPubkey race.
- 4 in `WoTServiceTest` — guardrail-holds-under-applyKind3, guardrail-releases-when-shrinks, close-stops-accepting-ops, close-is-idempotent.
- 6 in `FeedMetadataCoordinatorTest` — retry semantics + concurrent-EOSE-safety with 20 relays hammered from `Dispatchers.IO`.
- 7 in `OutboxDispatcherTest` — cached-outbox short-circuit, discover-then-fetch, no-10002 fallback, cached-author-covered-when-Phase-1-hangs, clear-releases-dedup, concurrent-EOSE-safety.

### Not in this revision (explicit follow-ups)

- Routing **stranger-avatar kind-0** through the outbox path — touches `MetadataPreloader` in more places; kept out to keep the diff focused on the primary WoT concern. Same pipeline is ready in `OutboxDispatcher.fetchKind0Only`.
- Persisting "author has no 10002" across sessions — deferred; a retry on next login is cheap.
- Graperank on `amy` — noted in the plan as a separate ticket.

Plan doc: [`commons/plans/2026-07-06-fix-wot-outbox-model-and-review-fixes-plan.md`](https://github.com/vitorpamplona/amethyst/blob/feat/wot-shared-index-relays/commons/plans/2026-07-06-fix-wot-outbox-model-and-review-fixes-plan.md)

nrobi144:
@davotoula thanks for the deep review — all six findings addressed with regression tests.

The two that scared me most:

- **Finding 1** (accountPubkey race → follow-list wipe): the ordering fix is in Main.kt (bind before hydrate), and I added defense-in-depth inside `consumeContactList` — `lastContactListByAuthor` is now only stamped inside branches where we know self-identity. If `accountPubkey` is still `null`, the stamp is skipped so a later relay retry with the same event populates cleanly. Test: `DesktopCachePipelineTest.selfKind3HydratedBeforePubkeyBindDoesNotPoisonLaterRelayRetry`.
- **Finding 2** (guardrail bypass): `handleFollowSet` now clears `myFollows` **before** returning and flips an `isDisabled` StateFlow. `handleKind3` short-circuits on `disabled`, Main.kt short-circuits the batch dispatch. Guardrail also releases itself when the follow set later drops back under the cap, so a user who trims their follows doesn't get stuck in "no WoT" mode. Test: `WoTServiceTest.guardrailIgnoresApplyKind3AfterTrip` + `guardrailReleasesWhenFollowSetShrinksBack`.

Findings 3–6 are covered in the summary comment above. Happy to iterate if any of the fixes miss the spirit of what you flagged.

nrobi144:
@vitorpamplona outbox model wired in. Pipeline lives in `commons/commonMain` (`OutboxDispatcher` + `OutboxCacheGateway`), reused by both the Desktop app and \`amy wot sync\`.

Uses your existing Quartz code end-to-end:

- Phase 1 batch-REQs kind-10002 from the configured index relays for every author.
- Phase 2 feeds the parsed `author → writeRelaysNorm()` map into `RelayListRecommendationProcessor.reliableRelaySetFor(...)` — the same transpose + minimal-cover algorithm you already ship. One `client.subscribe` covers all recommended relays in parallel.
- Phase 3 falls back to the index-relay REQ only for authors whose 10002 the network never returned.

On my account (587 follows, 4 index relays): Phase 1 covered 509 of 587 authors (86.7 %) in **819 ms**, Phase 2 pulled 199 kind-3s in **4 s** parallel across 23 recommended relays, Phase 3 mopped up 47 more kind-3s from the 78 no-10002 authors. Full trace + counters in the summary comment above.

Damus relay confirmed clean — no production default on this branch references `relay.damus.io`. Only common-test URL-parsing fixtures do; left those alone.

Graperank on \`amy\` — noted in the plan doc as a follow-up; happy to defer entirely to you.

vitorpamplona:
Cool, yes, amy graperank is almost ready. It's very different from this one, though, so I am not sure they are trying to solve the same thing anymore. Let me send the PR and then you can see if it makes sense to migrate or not.

vitorpamplona:
On of the CI issues here need to be resolved. I believe the other one is a race condition.,.. it might be solved in the next run

davotoula:
If ypu are using Claude I strongly recommend spending tokens on code review, it seems to pick up a lot of things missed during implementation. 

/simplify will do a code review and apply top fixes. Good focus on reducing duplication 

/code-review a very, very thorough code review. Produces a report, no fixes by default. 

/ecc:kotlin-review specialised on kotlin issues 

Personally I wouldn't dare to push anything that hasn't been through simplify and ecc kotlin review 😂

vitorpamplona:
Agree.. Also, It's fine to open PRs with just refactoring of the other stuff... If we don't clean up/simplify things that the AI writes, we will have code that even AI has trouble understanding.

vitorpamplona:
Not that the code I wrote until last year was much better.. :)

PR #3547, opened 2026-07-13
fix(ci): serialize :amethyst Kotlin compilations on CI to stop daemon OOMs

vitorpamplona:
## Summary

CI's `test-and-build-android` job keeps dying with `java.lang.OutOfMemoryError` (`GC overhead limit exceeded` / `Java heap space`) inside the shared Kotlin compile daemon when two `:amethyst` variant compilations run at the same time — `KotlinCompile` submits its work asynchronously via the Worker API, so the project lock is released while compilation is still running in the daemon and Gradle starts the next variant's compile alongside it. Two cache-cold `:amethyst` codegen passes no longer fit the daemon's `-Xmx8g` heap (runs [29260711359](https://github.com/vitorpamplona/amethyst/actions/runs/29260711359), [29248927324](https://github.com/vitorpamplona/amethyst/actions/runs/29248927324); removing Kover in #3539 helped but wasn't sufficient). This PR registers a no-op shared `BuildService` with `maxParallelUsages = 1` and has every `:amethyst` `KotlinCompile` task use it, so this module's compilations run one at a time while all other work (other modules, JVM tests, lint analysis, packaging) keeps its parallelism.

Gated on the `CI` env var (always set on GitHub Actions): the OOM needs a cache-cold multi-variant compile in one invocation, which local incremental single-variant builds don't produce — local builds are untouched.

## Test plan

- [x] Project configures cleanly on Gradle 9.4.1 (`./gradlew :amethyst:help`)

Notes:

Verified the scheduler behavior end-to-end with probe tasks (init script) registered on `:amethyst` that reproduce `KotlinCompile`'s execution shape — the task action submits a 6-second sleep through `WorkerExecutor.noIsolation()` and returns:

- **Baseline (no constraint):** two same-project probes started the same millisecond — reproduces the concurrent-compile overlap seen in CI.
  ```
  PROBE free0 START 1783958057960
  PROBE free1 START 1783958057959   <- overlap
  ```
- **With `usesService(limiter)` and `CI=true`:** strictly serialized; the second probe starts only after the first's async work completes (14s wall vs 8s).
  ```
  PROBE limited0 START 1783965585220
  PROBE limited0 END   1783965591229
  PROBE limited1 START 1783965591241   <- waits for completion
  ```
- **Without `CI` set:** the service is not registered at all (probe lookup fails with `BuildServiceRegistration ... not found`), confirming local builds don't carry the constraint.

This PR's own CI run is the real end-to-end test: the branch touches `:amethyst` build config, which invalidates the remote cache for exactly the compile tasks that were OOMing, forcing the cold four-variant build that previously failed.

## Interop suites

## AI assistance

## License

      original license header.

https://claude.ai/code/session_018ifbUGCADckUb3zaox9wBo

### DIFF
--- a/amethyst/build.gradle.kts
+++ b/amethyst/build.gradle.kts
@@ -1,4 +1,7 @@
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     alias(libs.plugins.androidApplication)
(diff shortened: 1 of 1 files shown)


davotoula:
This was needed!

PR #3605, opened 2026-07-17
Pre-seed Gradle distribution in web sandbox to unblock ./gradlew

vitorpamplona:
## Summary

The web sandbox's git-only GitHub proxy blocks downloads from `github.com` release assets (403 error), which prevents `./gradlew` from bootstrapping the Gradle distribution. This PR adds a pre-seeding mechanism that downloads the pinned Gradle distribution from verified mirrors during session startup, then validates it against the official SHA256 checksum before use. The `pre-push-spotless.sh` hook is simplified since infra failures are now prevented upstream.

## Changes

- **`.claude/hooks/session-start.sh`**: Added `seed_gradle_distribution()` function that:
  - Parses the Gradle wrapper properties to extract the distribution URL
  - Derives the cache directory path using the same base36(md5) algorithm Gradle uses
  - Fetches the official SHA256 checksum from `services.gradle.org` (reachable in the sandbox)
  - Downloads from Tencent Cloud or Huawei Cloud mirrors as fallbacks
  - Verifies the downloaded file against the official checksum before extraction
  - Skips entirely if the distribution is already installed (idempotent)
  - Gracefully degrades if mirrors fail, allowing `./gradlew` to attempt its normal flow

- **`.claude/hooks/pre-push-spotless.sh`**: Removed the infra-failure detection logic that was masking Gradle bootstrap errors. Since the distribution is now pre-seeded, any `spotlessApply` failure is a real formatting/compile error, not a sandbox hiccup.

- **`gradle/wrapper/gradle-wrapper.properties`**: Added `distributionSha256Sum` field for verification (Gradle 8.5+ supports this; older versions ignore it).

## Test plan

- [x] Verified `seed_gradle_distribution()` is called during session startup
- [x] Confirmed the function correctly derives the Gradle cache directory path
- [x] Tested that `./gradlew --version` succeeds after seeding
- [x] Verified idempotency: running the function twice doesn't re-download
- [x] Confirmed SHA256 verification rejects tampered files (tested with intentionally corrupted mirror response)

## Interop suites

- [x] N/A — change is build infrastructure only, doesn't affect wire bytes or protocol logic

## License

https://claude.ai/code/session_01A7bocRTSPyXwz6zVMzDbKR

### DIFF
--- a/.claude/hooks/pre-push-spotless.sh
+++ b/.claude/hooks/pre-push-spotless.sh
@@ -64,13 +64,9 @@ before="$(git diff HEAD -- '*.kt' '*.kts' 2>/dev/null | sha1sum)"
 log="$(mktemp /tmp/spotless-gate.XXXXXX.log)"
 if ! ./gradlew spotlessApply >"$log" 2>&1; then
   # Distinguish a formatting failure (block) from Gradle being unable to RUN —
-  # e.g. deps can't resolve, or the wrapper can't even download the Gradle
-  # distribution, in a restricted sandbox. An infra failure must not strand the
-  # agent; warn and let CI's spotlessCheck be the backstop. The distribution
-  # patterns (the `gradle-<ver>-bin.zip` URL, the Java download exception, and a
-  # proxy 40x on that fetch) only appear when `./gradlew` failed *before* running
-  # any task, so they can't mask a real formatting/compile error.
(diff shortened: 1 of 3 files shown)


davotoula:
Nice! Claude Web can build now?

vitorpamplona:
I think so... it took forever to figure out a solution, but it is working for me

PR #3634, opened 2026-07-18
Extract GrapeRank CLI flag names into constants

davotoula:
"insert-batch" and "relay-concurrency" are duplicated 3× each. They're args.intFlag(...) parser keys, so a mistyped occurrence would silently break flag parsing. Extracted FLAG_INSERT_BATCH / FLAG_RELAY_CONCURRENCY / FLAG_CONCURRENCY and the shared INSERT_BATCH_DEFAULT (500). Pure refactor, no behaviour change.

@vitorpamplona — one decision for you: the relay-concurrency fallback default is currently 16 in the report path but 4 in the crawl paths. I left the split as-is rather than unifying it. Is that intentional (report reads can fan out wider than crawl writes), or should both use the same value — and if so, 16 or 4?

## License

      original license header.


vitorpamplona:
Intentional but I am still testing. The crawler benefits from a lot of concurrency when checking for events everywhere when people have no outbox relay list for thememselves. 

If you are testing this, feel free to change anything you want. Including how the progress and results are rendered. I am still figuring out the best way of doing it.

Issue #3661, opened 2026-07-21
[BUG] ConcurrentModificationException


davotoula:
Fix in progress: stop ConcurrentModificationException in AddressableAuthorRelayLoaderSubAssembler by serializing filter invalidation through a debounced BundledUpdate and locking teardown against in-flight runs 

ConcurrentModificationException: 1.12.6-feat-share-note-as-q-BENCHMARK-PLAY

| Prop | Value |
|------|-------|
| Manuf |Google |
| Model |Pixel 9a |
| Prod |tegu |
| Android |17 |
| SDK Int |37 |
| Brand |google |
| Hardware |tegu |
| Device | tegu |
| Host | fc73d19e1a4b |
| User | android-build |
| Thread | DefaultDispatcher-worker-70 |

```
Thread: DefaultDispatcher-worker-70
java.util.ConcurrentModificationException
    java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1062)
    java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:1085)
    kotlin.collections.SetsKt___SetsKt.minus(r8-map-id-1c7acb6c136ada1563162fa8c237faa7e72610fafefe1fb4121d498467541502:45)
    com.vitorpamplona.amethyst.service.relayClient.reqCommand.event.loaders.AddressableAuthorRelayLoaderSubAssembler.invalidateFilters(r8-map-id-1c7acb6c136ada1563162fa8c237faa7e72610fafefe1fb4121d498467541502:89)
    com.vitorpamplona.amethyst.commons.relayClient.eoseManagers.IEoseManager.invalidateFilters$default(r8-map-id-1c7acb6c136ada1563162fa8c237faa7e72610fafefe1fb4121d498467541502:8)
    com.vitorpamplona.amethyst.service.relayClient.reqCommand.event.EventFinderFilterAssembler.invalidateFilters(r8-map-id-1c7acb6c136ada1563162fa8c237faa7e72610fafefe1fb4121d498467541502:22)
    com.vitorpamplona.amethyst.service.relayClient.reqCommand.event.EventFinderFilterAssembler.invalidateKeys(r8-map-id-1c7acb6c136ada1563162fa8c237faa7e72610fafefe1fb4121d498467541502:1)
[...]

PR #3666, opened 2026-07-22
Stop ConcurrentModificationException in AddressableAuthorRelayLoaderSubAssembler

davotoula:
Fix for #3661 

### Root cause

AddressableAuthorRelayLoaderSubAssembler.activeSubscriptions was a plain LinkedHashSet that was iterated (needed - activeSubscriptions, i.e. SetsKt.minus) and mutated (clear() / addAll()) with no synchronization, while invalidateFilters() ran synchronously on whatever thread called it.

Those callers are genuinely concurrent. ComposeSubscriptionManager.subscribe/unsubscribe invoke invalidateKeys() after releasing their own lock, and LifecycleAwareSubscription's 30-second grace-period unsubscribe fires on a Dispatchers.Default worker while composition subscribes from another thread. One thread iterating the set while another rewrote it produced the reported crash on DefaultDispatcher-worker-70.

This is the only member of EventFinderFilterAssembler.group that implements IEoseManager directly. Its two siblings extend BaseEoseManager, whose invalidateFilters hands off to a BundledUpdate — so their bodies never run on the caller's thread nor concurrently with themselves. This class had fallen off that ladder because BaseEoseManager also owns an INostrClient / SubscriptionController that an assembler bridging into another assembler has no use for.

I checked whether sibling assemblers had the same latent hole: they do not. Every other mutableMapOf/mutableSetOf field in this family is reachable only from inside an already-bundled body. This is a one-class hole, not the first of N.

### The fix

Route invalidateFilters through a BundledUpdate so bodies are serialized and run off the caller's thread, and commit the subscription delta under a private monitor shared with destroy().

### Performance

No lock is taken on the hot path. invalidateFilters() is now just bundler.invalidate(...) — an enqueue. The monitor is acquired only inside commit(), which runs on the bundler's IO thread, and in destroy(). The allKeys() scan and its per-stub LocalCache.getOrCreateUser lookups stay outside the monitor, so destroy() never waits on a cache sweep.

Contention is structurally near-zero: BasicBundledUpdate already guarantees one body at a time, so the only two parties that can ever meet on the monitor are an in-flight body and destroy().

**The hot path got cheaper, not more expensive**. Previously the full scan ran inline on the calling thread on every note composable mount/unmount — and ComposeSubscriptionManager documents subscribe/unsubscribe as "called by main. Keep it really fast." That work is now off-main and coalesced into a 500 ms window (event-driven, not a periodic tick — an idle app does no work).

To be clear about what was and wasn't measured: I did not benchmark the app, so I am claiming no runtime speedup number. The improvement is structural — work moved off the caller thread and repeated invalidations coalesced. The only measured numbers here are test-suite timings, which say nothing about app performance.

An earlier revision of this branch used AtomicReference + AtomicBoolean with a compensating "release what we just acquired" block. Review replaced it with the monitor: the atomics gave visibility but not mutual exclusion, so seven lines existed purely to patch a window that a monitor makes structurally unreachable. That also drops an @OptIn(ExperimentalAtomicApi::class) from a shipping Android class.

### Testing

New test class, 3 tests, all deterministic (latch-gated, no sleeps).

The concurrency test was verified RED against the unfixed code — 8 threads × 500 invalidations reproduced the actual production failure, not a proxy for it: [IllegalStateException: forceInvalidate bodies overlapped, ConcurrentModificationException: null].

Guard-off verification. The destroy()-race test asserts a leak's absence, which is exactly the shape of assertion that passes vacuously if the guard it targets is broken. So it was checked by deleting if (destroyed) return and confirming destroyDuringInFlightInvalidateDoesNotLeakSubscriptions fails, then restoring it. This was done twice — once against the original @Synchronized form, and again after switching to synchronized(lock) { }, because return then sits inside an inline lambda and the earlier check did not carry over.

Overlap is detected through the injected allKeys() lambda rather than by catching exceptions, because once the body is bundled its throwables are swallowed by BundledUpdate's CoroutineExceptionHandler and would never reach the test thread.

The destroy() test previously took 2.015 s because its latch only counted down on the leak path, so the passing case always burned the full timeout — it was slowest when it passed. It now asserts the absence directly: 0.510 s, class total 3.24 s → 1.74 s.

Full :amethyst suite green (941 tests, 0 failures) and :commons:jvmTest green.

### Also included

IEoseManager now documents its threading contract. The interface had no KDoc at all, which is arguably ther had no way to know the method must tolerate concurrent entry.
                                                                                                                                                                                                                                               ### Out of scope
                                                                                                                                                                                                                                               UserFinderQueryState identity equality defeats the subscription delta. It has no equals/hashCode and force instances each run, so needed - previous is always the full set — every run unsubscribes and re-subscribes everything. Pre-existing and unchanged by this PR, which means the diffing preserved here is currently decorative. It is bounded (downstream assemblers distinctBy { it.user }, so no duplicate REQs) and self-limiting (stops once no stub    has an unresolved author). Not fixed here because ComposeSubscriptionManager's identity semantics are delitiple duplications in these subscriptions since we do not control when screens are removed" — so addingstructural equality would collapse two composables watching the same user into one map entry, and the first to unmount would kill the other's live subscription. Any fix needs to reuse prior instances for unchanged logical keys inside this assembler instead. Fixing it would shrink the race window but never close it, so it had to be a separate c

Extracting a client-free BundledEoseManager. BaseEoseManager already contains this exact bundler + invalid, but conflates bundling with owning an INostrClient. Extracting the bundling half would let this classrejoin the base-class ladder. Skipped as out of proportion to a two-file bugfix: it touches a commons base with ~105 subclasses.

## License

      original license header.

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -8,6 +8,6 @@
   </component>
   <component name="KotlinJpsPluginSettings">
     <option name="externalSystemId" value="Gradle" />
-    <option name="version" value="2.4.0" />
+    <option name="version" value="2.4.10" />
   </component>
 </project>
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/service/relayClient/reqCommand/event/loaders/AddressableAuthorRelayLoaderSubAssembler.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/service/relayClient/reqCommand/event/loaders/AddressableAuthorRelayLoaderSubAssembler.kt
@@ -29,9 +29,6 @@ import com.vitorpamplona.amethyst.service.relayClient.reqCommand.user.UserFinder
 import com.vitorpamplona.amethyst.service.relayClient.reqCommand.user.UserFinderQueryState
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.IO
-import kotlin.concurrent.atomics.AtomicBoolean
-import kotlin.concurrent.atomics.AtomicReference
-import kotlin.concurrent.atomics.ExperimentalAtomicApi
 
 /**
  * Bridges missing-addressable-note authors into [UserFinderFilterAssembler].
@@ -43,18 +40,18 @@ import kotlin.concurrent.atomics.ExperimentalAtomicApi
  * kind-0 / kind-10002 and resolve outbox relays via [UserOutboxFinderSubAssembler]. Once the
  * relay list arrives, [EventFinderFilterAssembler] is invalidated and can query the correct relay.
  */
-@OptIn(ExperimentalAtomicApi::class)
 class AddressableAuthorRelayLoaderSubAssembler(
     val cache: LocalCache,
     val allKeys: () -> Set<EventFinderQueryState>,
     val userFinder: UserFinderFilterAssembler,
 ) : IEoseManager {
-    // Immutable snapshots swapped atomically, so a diff can never observe a half-written set.
-    // Mutual exclusion between runs comes from the bundler (one body at a time), not from these
-    // atomics — they exist to hand state over to destroy(), the one caller the bundler cannot
-    // serialize.
-    private val activeSubscriptions = AtomicReference<Set<UserFinderQueryState>>(emptySet())
-    private val destroyed = AtomicBoolean(false)
+    // Private monitor: @Synchronized locks on `this`, which leaves the instance's monitor
+    // reachable to anything holding a reference to this assembler.
+    private val lock = Any()
+
+    // Only ever touched while holding [lock]. See commit() and destroy().
+    private var activeSubscriptions: Set<UserFinderQueryState> = emptySet()
+    private var destroyed = false
 
     // Keeps the scan off the caller's thread. invalidateFilters() is reached synchronously from
     // ComposeSubscriptionManager.subscribe/unsubscribe on every note composable mount/unmount,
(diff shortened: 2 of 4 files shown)


vitorpamplona:
Did you try Quartz's ConcurrentSet? It could work well here and avoid the lock

davotoula:
Let me try that

davotoula:
Checked both — neither has the API this needs. They're minimal in opposite directions:

| | `quartz/utils/concurrent/ConcurrentSet` | `commons/util/ConcurrentSet` |
|---|---|---|
| enumerate | ✅ `snapshot()` | ❌ none |
| remove / clear | ❌ none | ✅ `remove` / `clear` |

This class needs both: enumerate the previous set, diff it against the new one, replace it. Quartz's can be diffed but never shrinks (append-only membership tracker). Commons' can shrink but can't be enumerated, so `activeSubscriptions - needed` is inexpressible.

The deeper reason it doesn't fit: the critical section isn't a set operation, it's a transaction — read prev → diff → subscribe → unsubscribe → replace. `newKeySet()` makes each op atomic and iteration CME-free, but can't order that sequence against `destroy()`. That ordering is what `destroyDuringInFlightInvalidateDoesNotLeakSubscriptions` covers: if `destroy()` lands mid-sequence the body re-subscribes authors it just released, leaving live kind-0/10002 REQs for a dead account after logout. A concurrent set removes the crash but leaves that leak.

Also, the lock isn't on the hot path — `invalidateFilters()` is just `bundler.invalidate(...)`. The monitor is taken once per bundled run on the IO thread, and since `BasicBundledUpdate` already serialises bodies, the only party it can contend with is `destroy()`. Lock striping wins when many threads hammer a set; the bundler already reduced that to ~one writer.

PR #3667, opened 2026-07-22
fix(video): decoder-budget exhaustion and black bars on live streams

vitorpamplona:
Two unrelated live-stream bugs found while investigating why a stream would not load on an emulator. Split into one commit each; neither depends on the other.

## 1. `fix(playback)` — "cannot load" is decoder exhaustion

MediaCodec instances are a per-process resource with a hard per-device ceiling. The Android emulator declares:

```xml
<MediaCodec name="c2.goldfish.h264.decoder" type="video/avc">
    <Limit name="concurrent-instances" max="4" />
```

Past it, `MediaCodec.start()` fails and the video reads as "can't load":

```
ResourceManagerService: reclaimResource: There aren't any clients to reclaim from
ResourceManagerMetrics: Codec: [c2.goldfish.h264.decoder] No of concurrent codecs: 5 ... codecs reclaimed: 0
MediaCodec: Codec reported err 0xfffffff4/NO_MEMORY, while in state 5/STARTING
MediaCodecRenderer: Failed to initialize decoder: c2.goldfish.h264.decoder
```

Reclaim finds nothing because all four holders are the same app at the same priority. Force-stopping the process made the identical stream play, which is what pinned it to codec accounting rather than the stream or the network.

`SimultaneousPlaybackCalculator` already computed the device ceiling, but it only reached `ExoPlayerPool` as `poolSize`, which bounds how many idle players are **retained**. The acquire path was uncapped (`coldPool.poll() ?: builder.build(context)`) and `MediaSessionPool` held a hardcoded `LruCache(10)` of sessions, each pinning a checked-out player — so a 4-decoder device would hold 10.

Fix enforces the budget at acquisition: a process-wide live-decoder count, headroom reclaimed by demoting warm players (own pool first, then siblings), and the session cache sized from the same budget with the old 10 kept as an upper bound. The counter is global because `PlaybackService` builds a separate pool for direct and for Tor-proxied traffic; a per-pool budget allowed twice the ceiling.

**Verified:** 9 codec allocations in one session, 0 `NO_MEMORY`, 0 decoder-init failures — where allocation #5 previously died.

## 2. `fix(video)` — black bars above/below live streams

A first-time-opened live stream drew ~90px of black top and bottom. The surface was correct 16:9; its enclosing box was not:

```
container   [0,274][1080,1062]  = 788px  (StreamingHeaderModifier's 300.dp cap)
TextureView [0,364][1080,972]   = 608px  (16:9 at 1080 wide), centred
                                → (788-608)/2 = 90px per side
```

Two causes, both required:

- `ContentWarningGate` accepts a `modifier` but **discards it for non-sensitive content** — that path emits `content()` bare. `ZoomableContentView` routed `mediaSizingModifier()` through exactly that parameter, so ordinary media never got sized, and the unconstrained player stretched to its enclosure and letterboxed inside. Now applied to the inner `Box`, which is always emitted.
- The ratio is unknown on first play (a NIP-53 stream carries no imeta `dim`; the cache fills only once the decoder reports a size), and the miss was frozen for the visit because a plain `LruCache` read triggers no recomposition. That is why the bars disappeared only on a **second** visit to the same stream. Cache entries are snapshot state now, plus a 16:9 default so the first layout already lands right.

**Verified** on a cold cache: container and TextureView both `[0,274][1080,882]`, header ends at 274 — zero gap. Feed image/video layouts unchanged.

## Reviewer notes

- The `VideoView.kt` change is **comment-only**. Making that read observable was tried and **broke playback** — the video redrew as a small thumbnail in the corner while layout bounds stayed correct, because flipping the ratio mid-playback also inserts a `Spacer` and restructuring children around a live `AndroidView` strands the player on a stale surface. Reverted; the comment exists so it is not "fixed" back.
- **Untested residual risk:** a stream whose true ratio is far from 16:9 will resize its container once mid-playback and could hit that same stale-surface glitch. Every stream reachable during testing was 16:9, so this is unexercised rather than proven safe.

Tested on a Pixel 9 API 36 emulator (`arm64`, Tor enabled). `./gradlew test` green.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/MediaAspectRatioCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/MediaAspectRatioCache.kt
@@ -21,6 +21,8 @@
 package com.vitorpamplona.amethyst.model
 
 import androidx.collection.LruCache
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
 
 interface MutableMediaAspectRatioCache {
     fun get(url: String): Float?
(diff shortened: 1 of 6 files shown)


davotoula:
That's odd. I tested thoroughly on device but not on emulator. 

I found a few broken streams. Especially "no good radio" uses encoding not supported by Exoplayer... But works in VLC

vitorpamplona:
I think devices have a large amount of available decoders. So the emulator is far stricter in that sense.

davotoula:
All good, I just double checked and my previous work on this and it is not overlapping with your fixes.

I found a minor issue in your PR, testing locally and will submit pr if worth it

Issue #644, opened 2023-10-12
[FEATURE] warn user about DMs from reported impersonators


bleetube:
**Is your feature request related to a problem? Please describe.**
When I check the kind 1984 notes on our relays, I frequently see impersonator reports. Users complain about receiving unsolicited DMs from fake profiles.  The fake profiles are extremely low effort, and have behavior such as reposting the original account. Here are some recent examples.

```javascript
  "06281c8118cdeef1dd67fe42875909993bc21623b4473327c291d027f6a8a10f", // impersonator 
  "daf8436ae707b5214c18bb5325d90b1dddbe9f4c1320e57879f655455dea7591", // impersonator npub1pyp9fqq60689ppds9ec3vghsm7s6s4grfya0y342g2hs3a0y6t0segc0qq
  "044b4adfe4a112ef7f7f43861bb4adf7a9ea60cde0ddaedf0220573b19d1b14e", // impersonator npub1uescmd5krhrmj9rcura833xpke5eqzvcz5nxjw74ufeewf2sscxq4g7chm
  "becec2df63d89d5916d122e446a68c53ec64ef6ea47fd082a47e994960040378", // impersonator npub1qy2tkywa36ufh76qv62snv4953eayufxufav4evz2le06lxe2amqpxavna
  "c8b40ae7ce27ef4ea21f84e1a9d375068a1c95d38889aa6d7cf5c92591a0f9fc", // Adam Back impersonator
```

**Describe the solution you'd like**
The DM interface in Amethyst could do a lookup filter on <npub>+1984 to see if there have been any recent reports about an account. If there are matches, the user could be warned that the account is possibly an impersonator. Kind of like the warnings Gmail shows when you get an email outside of your organization.

**Bounty (in Bitcoin sats) offered for the implementation**
210k sats

**Additional context**
I think its great there's folks on nostr who take the moment out of their day to report impersonators (probably out of the frustration from receiving unsolicited messages). afaict we are letting those efforts go to waste currently.

vitorpamplona:
We do that already but it requires 5+ reports from the people you are following. We need to add some customization to allow people to adjust that number.

weex:
If I'm following someone, I'd probably trust a single report versus a random account. To combat abuse the reporter could be identified in the warning, perhaps hidden under a tap so the user doesn't have know who's doing the reporting unless they want to.

davotoula:
@bleetube delivered, see attached pr.

bleetube:
[...]

PR #3680, opened 2026-07-23
Warn users when a DM comes from an account that people they follow have reported

davotoula:
Fix for #644

Warning in feed:
Screenshot_20260723-182230

Screenshot_20260723-182243 Screenshot_20260723-182254

Screenshot_20260723-182320

## The gap

Amethyst already collects kind-1984 (NIP-56) reports, but a report filed against a *note* is routed to that note and never reaches the reported author's user-level cache. So the DM surfaces showed no signal at all: over the hide threshold a room is silently dropped, and under it there was nothing — a follow reporting a DM sender for impersonation or malware was invisible to you.

Note that two of the issue's threads are already resolved in `main`: the report threshold has been configurable for a while (`reportWarningThreshold` in Security Filters), and reporter avatars already render on hidden notes. This PR is the original ask that remained: surfacing that signal in the DM UI.

## What this adds

Two surfaces, both for 1:1 private DMs (NIP-04 + NIP-17):

- **A dismissible banner** at the top of the conversation — "Reported by N people you follow", with the report reasons and, on tap, the reporters themselves ("who reported this?" hidden behind a tap, per @weex's suggestion in the issue).
- **A warning glyph** on the chat-list row, so you're warned *before* opening an unsolicited DM — the New/requests tab is the primary surface, since that's where impersonation DMs from strangers land.

## How it works

- A new additive index `reportsNamingUser` on `UserReportCache`, populated from every report that names an author — including note-filed reports that previously only reached the note. It is kept **separate** from `receivedReportsByAuthor`, which the hide path counts, so `Account.isAcceptable`, `countReportAuthorsBy` and the hide threshold are completely unchanged — no user's hidden status shifts.
- A new `ReportEvent.reportedAuthorsWithOwnType()` in quartz indexes only authors a report is actually *about* (their `p` tag carries its own report type), not accounts incidentally `p`-tagged as mentions of an event-scoped report.
- A pure classifier (`dmReportWarningFor`) in `commons` decides whether to warn, unit-tested in isolation; `AccountViewModel` wraps it in a cached `StateFlow` mirroring the existing `createIsHiddenFlow`.

## Design decisions

- **Warns at a single report from someone you follow**, with no threshold suppression — the hide threshold governs individual messages, never the chat-list row, so suppressing above it would give the most-reported senders the *least* signal where you decide whether to open the DM.
- **Never warns about someone you follow or yourself**, and respects the existing "Warn about reported posts" toggle — no new setting.
- **Reasons are shown on their own line, not interpolated** into the sentence, because the labels are sentence-cased noun phrases that don't survive being spliced into inflecting languages.
- **All report types trigger the warning, with the reason named.** The issue is titled "impersonators," but the warning's value is "should I trust this DM sender", for which a malware or scam flag is at least as urgent — so impersonation is surfaced by name rather than being the sole gate. This is the one point I'd like a maintainer's call on: happy to narrow it to impersonation (or a deception/danger subset) if you'd prefer — it's a one-line filter in the classifier.

## A tradeoff worth flagging

One report from one follow now produces a visible warning, which is a lower bar than anything else in the app. It's mitigated structurally rather than numerically: the reporter is always named (a smear is attributable), and the result is a warning you can read past, not a hide.

## Testing

- Unit tests for the classifier, the additive index, and the ingestion routing (including that note-filed reports reach the new index while leaving the hide count untouched).
- Verified end-to-end on device against a real, heavily-reported Michael Saylor impersonator: the banner and glyph render, dismissal survives backgrounding, and the `p`-tag narrowing correctly warns about a typed offender while ignoring an incidentally-mentioned bystander in the same report.
- Cross-checked report counts against relays with `nak` — the app's follow-scoped counts reconcile with the network totals.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -1742,6 +1742,11 @@ object LocalCache : ILocalCache, ICacheProvider {
                 event.reportedPost().mapNotNull { checkGetOrCreateNote(it.eventId) } +
                     event.reportedAddresses().map { getOrCreateAddressableNote(it.address) }
 
+            // Every report that names an author lands here, including those that also name an event
+            // and are therefore filed against the note below. Read only by the DM sender warning —
+            // the hide threshold keeps counting `receivedReportsByAuthor` alone.
+            authorsReported.forEach { author -> author.reports().addReportNamingUser(note) }
+
             if (eventsReported.isEmpty()) {
                 authorsReported.forEach { author -> author.reports().addReport(note) }
             } else {
@@ -3268,7 +3273,10 @@ object LocalCache : ILocalCache, ICacheProvider {
 
         if (noteEvent is ReportEvent) {
             noteEvent.reportedAuthor().forEach {
-                getUserIfExists(it.pubkey)?.reportsOrNull()?.removeReport(note)
+                getUserIfExists(it.pubkey)?.reportsOrNull()?.let { reports ->
+                    reports.removeReport(note)
+                    reports.removeReportNamingUser(note)
+                }
             }
 
             noteEvent.reportedPost().forEach {
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -1742,15 +1742,26 @@ object LocalCache : ILocalCache, ICacheProvider {
                 event.reportedPost().mapNotNull { checkGetOrCreateNote(it.eventId) } +
                     event.reportedAddresses().map { getOrCreateAddressableNote(it.address) }
 
-            // Every report that names an author lands here, including those that also name an event
-            // and are therefore filed against the note below. Read only by the DM sender warning —
-            // the hide threshold keeps counting `receivedReportsByAuthor` alone.
-            authorsReported.forEach { author -> author.reports().addReportNamingUser(note) }
-
+            // Every report that names an author also lands in the naming index, including those
+            // filed against a note rather than a profile. Read only by the DM sender warning — the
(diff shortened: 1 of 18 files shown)


davotoula:
## Testing

**Manual**, on a Pixel 9a (Android 17), on the **R8-minified benchmark build** (release-shaped, so the scroll glance is meaningful rather than a debug artifact):

- **Card rendering** — verified against a real, heavily-reported Michael Saylor impersonator (27 reporters among accounts followed). Collapsed banner shows the count and subtitle; expanded shows the reporter facepile and the reason chips ("Other, Impersonation, Spam"). The facepile matched the headline exactly — 27 distinct reporters, no duplicate avatars.
- **Chat-list glyph** — present on the reported row in both the Known and requests tabs; absent for followed accounts, unreported accounts, and group rooms.
- **Dismissal** is session-scoped and survives backgrounding: dismiss → switch to another app and back → still dismissed; force-swipe the app away (process death) → resets on the next open, as expected. Leaving and re-entering the room re-warns, by design.
- **`p`-tag narrowing** — published a note-scoped report naming a typed offender *and* an incidentally-mentioned bystander (bare `p` tag) in the same event. The offender's count incremented (27 → 28); the bystander's DM showed no banner, no glyph, and an empty Reports tab — confirming only authors a report is genuinely *about* get indexed.
- **Profile Reports tab** corroborates the banner: it lists the surfaced reports. Cross-checked against relays with `nak` — the app's follow-scoped counts reconcile with the network totals (33 loaded in-app vs 388 total on relays from 303 distinct reporters, dominated by impersonation and spam, matching the reasons the card showed).

**Performance glance:** each 1:1 row stands up its own `combine` over `reportsNamingUser` + the follow list + the warn toggle — the same shape as the existing `createIsHiddenFlow` used per-note across the feeds. Flick-scrolling a long conversation list on the release-shaped build showed no visible jank. The per-row flows are bounded by a 300-entry LRU; a list longer than that would evict and rebuild flows on scroll-back, which is the one thing worth watching at extreme scale.

PR #3690, opened 2026-07-24
Remove unreachable duplicate StreamMessageV2Event branch in computeReplyTo

davotoula:
LocalCache.computeReplyTo had two `is StreamMessageV2Event ->` branches in the same `when`. The `when` matches the first, so the second (the older buzzThreadRoot+buzzThreadReply variant) was dead code — the condition Sonar flags as duplicating the earlier one. The surviving first branch is the newer, deliberate Concord-style threading behavior that links a 40002 thread reply into its parent's replies via buzzThreadReply. Also drops the now-unused buzzThreadRoot import.

## License

      original license header.


vitorpamplona:
wierd

PR #3782, opened 2026-07-28
fix(buzz): make the Messages toggle read the list it claims to change

vitorpamplona:
"Remove from Messages" was hardcoded on every Buzz surface, so a channel already off the list never offered an "Add to Messages" counterpart.

Removal itself always worked — verified on-device: the kind-10009 republished to 3 relays and the row left Messages. What was missing was any *read* of that list on the way back.

## Channels

The Buzz workspace rows were the worse half: `BuzzRelayImportViewModel._added` was seeded once at `bind()` and only ever grew, so a channel taken off Messages still rendered as a **disabled "Added"** — no way back at all.

- `RelayGroupListState` — new `liveRelayGroupIds`, the joined groups as normalized `GroupId`s, so the UI can ask whether a channel is on Messages without string-matching a raw relay url another client may not have normalized the way we do.
- `RelayGroupTopBar` / `BuzzImportRow` — one Add/Remove toggle driven by that flow. Remove no longer pops back: you stay a member reading the channel, and staying is what makes the entry flip so the action is visibly undoable. **Leave still pops.**
- `BuzzRelayImportViewModel` — tracks "added" against the live list instead of a one-shot seed; adds `remove()`; `add()` now also clears the dismissal so a relay's kind-44100 re-announcement isn't filtered back out.
- `AccountViewModel` — `addRelayGroupToMessages()` as the counterpart to `removeRelayGroupFromMessages()`; `acceptChannelInvite()` delegates to it.

## DMs

Same one-way shape, different cause. Hiding a DM is relay-side and per-viewer (kind-41012 → the kind-30622 snapshot), and that part was already correct and reactive — but `rebuildRows` *dropped* hidden DMs on the floor, so a hidden conversation was gone for good.

There is no unhide command in Buzz. Re-opening is the unhide: a kind-41010 with the same participants resolves to the same canonical channel and drops it from the hidden snapshot. Confirmed against the live relay rather than taken on trust (see below).

- `BuzzDmListViewModel` — one pass now `partition`s into `rows` and a new `hiddenRows`; adds `removeFromMessages()` / `addToMessages()`.
- `BuzzDmListScreen` — the kebab is a toggle, hidden rows render faded but still openable, behind a collapsible "N hidden conversations" header.
- `RelayGroupChannelListScreen` — the inline DM rows had **no menu at all**, and that screen is where DMs actually live: the full inbox sits behind a "see all" row that only appears above 6 DMs, so in a small workspace the Hidden section would have been unreachable. They now carry the same toggle and the same hidden tail.

## FAB overlap

Both list screens now leave bottom room for the FAB, which the Scaffold's padding deliberately doesn't account for — the last row's overflow menu was sitting underneath it. `contentPadding` (not a modifier) so rows scroll *through* that strip; 96.dp to match `JobBoardScreen`.

## Testing

On-device against `nosfabrica.communities.buzz.xyz`:

- **Channel:** removed `amethyst-create-test3` → gone from Messages, row menu flips to "Add to Messages" → tapped → back in Messages, flips to "Remove". Top-bar menu round-trips without leaving the screen.
- **DM:** hid "Vitor (Buzz)" → relay pushed a new kind-30622, row moved under **1 hidden conversation** → tapped "Add to Messages" → relay pushed kind-30622 again, DM returned to the live list at full opacity.
- **FAB:** re-ran the exact case that caught it, without collapsing the Channels section — the last row's kebab now opens its menu instead of hitting the FAB.

Adds `SimpleGroupListEventTest` covering the removal path, including that a renamed channel still matches (removal keys on group id + relay only).

`spotlessCheck`, `:amethyst:testPlayDebugUnitTest`, `:quartz:jvmTest`, `:commons:jvmTest` pass. `lintPlayDebug` reports no new issues for the added resources (its 3 errors are pre-existing, in an unrelated `androidTest` playback file).

## Notes

- `buzz_dm_hidden_count` is a `<plurals>`, per `amethyst/src/main/res/CLAUDE.md`.
- Not changed, flagging only: the same missing-`contentPadding` pattern likely exists on sibling screens — `ConcordChannelListScreen.kt:324` is `LazyColumn(Modifier.fillMaxSize().padding(padding))` with a FAB, so it almost certainly has the identical overlap. Not verified on-device.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountViewModel.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/AccountViewModel.kt
@@ -1689,16 +1689,24 @@ class AccountViewModel(
         }
 
     /**
-     * Accept a channel somebody added me to: write it into my kind-10009 so it shows on Messages and
-     * follows me to other devices. No kind-9021 join — the relay already put me in the roster, which is
-     * why the channel opens and accepts posts today; this only records *my* decision to surface it.
+     * Put a relay group (back) on Messages: write it into my kind-10009 so it shows and follows me to
+     * other devices, and clear the dismissal [removeRelayGroupFromMessages] left behind so a Buzz relay
+     * re-announcing my membership isn't filtered out. No kind-9021 join — the relay roster is untouched
+     * by both halves of this toggle; this only records *my* decision to surface the channel.
      */
-    fun acceptChannelInvite(channel: RelayGroupChannel) =
+    fun addRelayGroupToMessages(channel: RelayGroupChannel) =
         launchSigner {
             account.settings.undismissChannelInvite(channel.groupId.id)
             account.follow(channel)
             BuzzChannelInvites.remove(account.userProfile().pubkeyHex, channel.groupId.id)
         }
+
+    /**
+     * Accept a channel somebody added me to. Identical to [addRelayGroupToMessages] — accepting an
+     * invite *is* surfacing the channel, since the relay already put me in the roster (which is why
+     * the channel opens and accepts posts today).
+     */
+    fun acceptChannelInvite(channel: RelayGroupChannel) = addRelayGroupToMessages(channel)
 
     /**
      * Keep the channel off Messages without touching membership. Local and reversible — I stay in the
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/buzz/BuzzDmListScreen.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/buzz/BuzzDmListScreen.kt
@@ -56,7 +56,9 @@ import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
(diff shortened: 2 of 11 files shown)


vitorpamplona:
---

### Follow-up: `ConcordChannelListScreen` FAB overlap

The suspicion flagged above is confirmed and fixed in a second commit: `LazyColumn(Modifier.fillMaxSize().padding(padding))` with a FAB and no `contentPadding`, and its rows carry a manager-only overflow menu — the same defect.

Gated on `canManageChannels`, because that flag is also what renders the FAB there: a plain member has nothing to clear and no reason to lose the space.

**Verification gap, stated plainly:** the non-manager path is verified on-device across three joined communities (Soapbox Community, "3rd times a charm?", "Hi Test Room") — renders normally, no wasted space. The **manager** path is *not* device-verified: this account has no manage rights in any joined Concord community, and creating one to test would publish real community events to relays, which I didn't want to do unprompted. It is the identical `contentPadding` change already device-verified on the two Buzz screens in this PR.

vitorpamplona:
### Manager case now verified on-device (A/B)

The gap flagged in my previous comment is closed. I created a throwaway Concord community (owner = me, so `canManageChannels` is true and the FAB + row overflow menus render), added channels until the list overflowed, and shrank the viewport to `1080x1000` via `adb shell wm size` so the last row lands at the bottom edge — the only state in which the overlap occurs.

**Before** (fix reverted, same file only):

```
ch-ff    519 435
ch-dd    519 626
general  509 817     <- last row
New channel (FAB)  964 821
```

The last row and the FAB occupy the same vertical band. Tapping the row's overflow at `(1006, 817)` opened the FAB's dialog — `'New channel' / 'Channel name' / 'Cancel'` — instead of the channel menu. That is the bug.

**After** (fix restored):

```
chan-four  519 374
chan-five  519 565   <- last row, rests clear
New channel (FAB)  964 821
```

Tapping the last row's overflow at `(1006, 588)` opened `'Rename channel' / 'Delete channel'` — the correct menu.

Non-manager path also re-checked across three joined communities (Soapbox Community, "3rd times a charm?", "Hi Test Room"): FAB absent, no clearance applied, no wasted space, no regression.

Cleanup note: this left two throwaway Concord communities ("FAB Clearance", "FAB Before") on my own account's list — test artifacts only, no effect on this branch.

PR #3785, opened 2026-07-28
fix(buzz): drop the grey slab behind the community Channels section

vitorpamplona:
On a Buzz community screen the Channels section sat on a grey block while the Direct Messages section right below it sat on the plain background — the two halves of one screen looked like different surfaces.

The cause is that each channel row is its own filled Material3 `Card`. They stack with no gaps, so the cards' container colour merges into one continuous slab behind the whole section, reading as a box drawn around the channels.

Rendered as a plain `Box` on the screen background instead, keeping the click on the container so the whole row still opens the channel. That matches `BuzzDmInlineRow` directly below it, `RelayGroupChannelRow` on the vanilla NIP-29 branch of the same screen, and the Concord server list — this row was the only one wrapped in a Card.

Affects the **Channels** and **Forums** sections (same composable); no layout, padding or row-height change, and no new strings.

## Testing

On-device, `nosfabrica.communities.buzz.xyz`:

- Channels section now renders on the same background as Direct Messages; the screen reads as one continuous list.
- Tapped a channel row → opens the chat (`amethyst-create-test3`), so moving the click from `Card` to `Box` kept the whole row tappable.
- Tapped a row's overflow → `Pin channel` / `Remove from Messages` still open.

`spotlessCheck` and `:amethyst:testPlayDebugUnitTest` pass.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/buzz/BuzzImportRow.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/buzz/BuzzImportRow.kt
@@ -21,6 +21,7 @@
 package com.vitorpamplona.amethyst.ui.screen.loggedIn.buzz
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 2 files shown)


vitorpamplona:
### Second commit: the hairline back between rows

Dropping the Cards also dropped the only thing separating adjacent rows, so this adds the `0.25.dp` `outlineVariant` hairline that the **vanilla NIP-29 branch of this same screen** and the Concord server list already use — same value, same colour, no new constant.

Drawn *before* each row except the first, so a section's own header keeps providing the separation at its boundaries: no stray line under the last channel, none above "Direct Messages".

**Scope note — I widened this slightly beyond the Channels section, deliberately.** Hairlines on the channel rows alone would have re-introduced a milder version of the inconsistency this PR set out to fix (Channels styled one way, Direct Messages another). So it's applied to every row list on the screen: chat channels, forums, inline DMs, and hidden DMs. Say the word if you'd rather it were Channels-only.

The vanilla branch now calls the same `RowHairline(index)` helper instead of inlining the divider, so there's one definition rather than five copies.

Verified on-device: hairlines between channel rows and between DM rows; nothing before the first row of either section, nothing between the last channel and the Direct Messages header.

PR #3798, opened 2026-07-29
Address buzz CLI review findings; pin the wrapper copies and the amy arg surface

davotoula:
Three findings from a code review of the Sonar literal-extraction pass that landed earlier. None was introduced by that pass — it renamed or sat next to each one.

**`no_relays` told the wrong story.** `relaysFor` returns an empty *non-null* set when `--relays` was given but every entry failed `normalizeGroupRelay`, so the elvis to `outboxRelays()` never fires and the user was told to pass the flag they had just passed. The message now names the real problem and echoes the rejected value; the `no_relays` error code is unchanged. Worth knowing from testing it: a bare word like `garbage` is *accepted* (the normalizer prepends `wss://`), so the realistic trigger is a pasted `http://` URL.

**`participants` was string-munged.** `arr.toString().trim('[',']').split(",")` turns any shape other than a flat array of strings into plausible-looking fake pubkeys, and splits a comma-bearing string into two entries. It is now decoded as a `JsonArray`, with non-string elements skipped rather than stringified. Extracted to `participantsOf` so it is reachable from a test — `BuzzParticipantsTest` covers the shapes, and was confirmed to fail 3 of its 5 cases against the old implementation rather than merely passing against the new one.

**Nothing kept the two copies of the buzz-agent wrappers in sync.** They exist as byte-identical trees under `cli/src/main/resources/buzz-agent/` (what `amy buzz agent up` extracts) and `tools/buzz-agent/` (what the README tells people to point `--exec` at), with no build step or test pinning them — the same two-line patch had to be hand-applied twice in one week. `BuzzAgentWrapperSyncTest` now asserts it.

That test needed one non-obvious piece of wiring: the `tools/` tree lives outside the module, so without declaring it as a `:cli:test` input Gradle calls the task up-to-date after a `tools/`-only edit and the test silently does not run — exactly the drift it exists to catch. It only failed under `--rerun-tasks` until that input was declared; verified both ways afterwards.

**Plus a harness for the argument surface.** `tools/buzz-agent/test-amy-args.sh` pins the flag names and error strings the wrapper scripts and their operators key on — `--base-ref`, the `repo-naddr-or-coordinates` positional, `pass --channel GID`, and the `no_relays` detail. Renaming a flag or rewording an error breaks callers while compiling clean, and none of it is covered by a type. 14 cases, all offline against an isolated `~/.amy` via the same `amy.home` seam the JVM tests use. It scores 14/14 here and 10/14 when built against the pre-fix `BuzzCommands`, the four failures printing the old misleading message.

## License

      original license header.

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -29,6 +29,17 @@ tasks.withType<ProcessResources>().configureEach {
     duplicatesStrategy = DuplicatesStrategy.INCLUDE
 }
 
+// BuzzAgentWrapperSyncTest compares the bundled buzz-agent wrappers against their
+// tools/ reference copies. The reference tree lives outside this module, so without
+// declaring it Gradle calls :cli:test up-to-date after a tools/-only edit — exactly the
+// drift the test exists to catch.
+tasks.named<Test>("test") {
+    inputs
+        .dir(rootProject.layout.projectDirectory.dir("tools/buzz-agent"))
+        .withPropertyName("buzzAgentWrapperReference")
+        .withPathSensitivity(PathSensitivity.RELATIVE)
+}
+
 dependencies {
     implementation(project(":quartz"))
     implementation(project(":commons"))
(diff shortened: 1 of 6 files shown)


davotoula:
@vitorpamplona ready for review/merge

PR #3810, opened 2026-07-29
feat(nip73): show rich previews for external-content comments

dmnyc:
## Summary

Kind 1111 (NIP-22) comments scoped to external content via NIP-73 (e.g. a comment on a web article URL) rendered as a bare, unadorned URL with no context — both as the "replying to" indicator above a comment, and as the screen title when viewing a URL's own comment thread. This reuses the existing OpenGraph preview pipeline (already used for links typed into note bodies) so these comments get the same rich image/title/description card other Nostr clients (e.g. Jumble) already show.

- `DisplayUrlExternalId` now renders a full preview card instead of a bare link chip wherever a comment's external scope is shown as context (feed rows, expanded threads, the reply-compose "replying to" banner).
- The URL thread screen (`Route.Url`, "External Content") now has a back arrow + static title instead of the raw URL crammed into the top bar, with the same preview card as the first item of the scrolling reply list — not a second, redundant copy per comment, since every comment there already shares that one scope.
- `UrlPreviewCard` gained an `onCardClick` override plus an always-visible external-link icon next to the domain, so "view the thread" (tap the card) and "open the article" (tap the icon) stay distinct actions.
- One follow-on fix found while testing live: the expanded "Thread" screen's master-note renderer had no `CommentEvent` branch at all, so an externally-scoped top-level comment shown as a thread's root got zero context — not even a bare chip. Fixed the same way.

No protocol/data-model changes — `CommentEvent.scope()` / `ExternalId` / `UrlId` and the reply-authoring path already handled this correctly; this is purely a rendering fix.

### Before

Image
Image

### After

Image
Image

## Test plan

- [x] Verified live on a physical device against a real NIP-73 URL-scoped kind 1111 comment: feed row, expanded thread, and the URL thread screen all render the rich card correctly, with no redundant repeats and a naturally scrolling header.
- [x] `./gradlew spotlessApply`
- [x] `./gradlew :amethyst:compileFdroidDebugKotlin`

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/LoadUrlPreview.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/LoadUrlPreview.kt
@@ -52,13 +52,10 @@ fun LoadUrlPreview(
 }
 
 @Composable
-fun LoadUrlPreviewDirect(
+fun rememberUrlPreviewState(
     url: String,
-    urlText: String,
-    callbackUri: String? = null,
     accountViewModel: AccountViewModel,
-    nav: INav? = null,
-) {
+): UrlPreviewState {
     @Suppress("ProduceStateDoesNotAssignValue")
     val urlPreviewState by
         produceState(
@@ -69,6 +66,18 @@ fun LoadUrlPreviewDirect(
                 accountViewModel.urlPreview(url) { value = it }
             }
         }
+    return urlPreviewState
+}
+
+@Composable
+fun LoadUrlPreviewDirect(
+    url: String,
+    urlText: String,
+    callbackUri: String? = null,
+    accountViewModel: AccountViewModel,
+    nav: INav? = null,
+) {
+    val urlPreviewState = rememberUrlPreviewState(url, accountViewModel)
 
     CrossfadeIfEnabled(
         targetState = urlPreviewState,
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/UrlPreviewCard.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/UrlPreviewCard.kt
@@ -21,12 +21,11 @@
 package com.vitorpamplona.amethyst.ui.components
 
 import androidx.compose.foundation.ExperimentalFoundationApi
-import androidx.compose.foundation.clickable
 import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.size
+import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
@@ -40,7 +39,6 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalClipboard
 import androidx.compose.ui.platform.LocalUriHandler
 import androidx.compose.ui.text.style.TextOverflow
-import androidx.compose.ui.unit.dp
 import coil3.compose.AsyncImage
 import com.vitorpamplona.amethyst.R
 import com.vitorpamplona.amethyst.commons.icons.symbols.Icon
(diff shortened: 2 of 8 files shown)


vitorpamplona:
very cool!

davotoula:
From my clonker to your clonker

Findings
1. Both new preview call sites bypass the showUrlPreview() data-saver / privacy gate. — the one thing I'd want fixed before merge.
DisplayExternalId.kt:99 and UrlScreen.kt:134 call rememberUrlPreviewState directly. Every other render path gates first: LoadUrlPreview.kt:47, Wiki.kt:91, MarkdownMediaRenderer.kt:141, and PrefetchFeedMedia.kt:229 (whose KDoc literally says "honoring the data-saver gates"). The setting is a ConnectivityType — ALWAYS / WIFI_ONLY / NEVER (UiSettingsState.kt:62) — so a user on mobile data with WIFI_ONLY, or on NEVER, now gets an outbound OpenGraph request to the third-party article host plus a full-bleed image download for every external-scoped comment that scrolls past. Extracting rememberUrlPreviewState out of LoadUrlPreviewDirect skipped the gate that lived in its caller. Fix: check accountViewModel.settings.showUrlPreview() and fall back to DisplayExternalIdChip / DisplayUrlHeader.
2. The new open-in-browser icon is a 14dp touch target. UrlPreviewCard.kt:139-147 — Modifier.size(14.dp).clickable {}, and commons Icon (Icon.kt:30) is a bare Material3 Icon with no minimumInteractiveComponentSize(). That's well under the 48dp Android / 24px WCAG 2.2 minimum, and it's nested inside the card's combinedClickable, so a near-miss doesn't do nothing — it silently fires the other action and navigates to the thread. Wrap it in an IconButton, or add minimumInteractiveComponentSize() plus role = Role.Button.
3. The icon is added unconditionally, including where it's a no-op duplicate. In note bodies RenderLoaded passes no onCardClick, so tapping the card and tapping the icon both open the URL. The icon only disambiguates when onCardClick != null. Gating it on that keeps every existing link card visually unchanged and confines the change to the new case.
4. compositionLocalOf → staticCompositionLocalOf (DisplayExternalId.kt:66). The value is constant for a screen's lifetime; static avoids read tracking in every feed row.
5. These previews aren't prefetched. collectWarmTargets() harvests imeta tags and rich-text body links, not the NIP-73 I scope — so these cards pop in mid-scroll while every other link card is warmed ahead. A targets.link(...) for the scope URL would make it consistent, and it sits behind the same gate as #1.
6. Nits. UrlScreenPreview's fallback passes bare Modifier (UrlScreen.kt:140), so while loading or on error the raw URL is flush to the screen edge while the card is inset — use MaxWidthWithHorzPadding. And Modifier.size(14.dp) / Modifier.size(4.dp) should use the hoisted constants (Size14Modifier, Shape.kt:178) per the repo's allocation convention.
7. Worth an eyeball, not a defect. In the composer (GenericCommentPostScreen.kt:255) the banner is now a full preview card whose entire surface navigates away, sitting directly above a half-written draft. The chip it replaced navigated too, so it isn't a behaviour regression — just a much larger accidental-tap surface next to unsaved text.

dmnyc:
Thanks — checked each of these against the code; all six were real. Fixed in adc3601:

1. **Data-saver/privacy gate.** Confirmed and fixed — this was the real bug. Both call sites now check `settings.showUrlPreview()` and fall back to the plain chip / URL header. You're right about the cause: the gate lived in `LoadUrlPreviewDirect`'s caller, so pulling `rememberUrlPreviewState` out bypassed it.
2. **Touch target.** Now an `IconButton` (the repo's existing idiom) rather than a 14dp `clickable`. Worth noting your point about the nesting made this worse than a plain miss — a near-miss fell through to the card's `combinedClickable` and navigated instead of doing nothing.
3. **Icon shown unconditionally.** Gated on `onCardClick != null`, so note-body link cards are visually unchanged and the change is confined to the new external-scope case.
4. **`staticCompositionLocalOf`.** Done.
5. **Prefetch.** Left out for now — agreed it's a real consistency gap, but it means touching `collectWarmTargets()` in shared every-feed code, which felt out of scope for this PR. Happy to do it here if you'd prefer, otherwise I'll follow up separately.
6. **Nits.** Loading/error header now uses `MaxWidthWithHorzPadding` to match the card's inset, and the icon uses `Size14Modifier`. Left the 4dp spacer literal moot — the `IconButton` provides its own spacing, so it's gone.
7. **Composer banner.** Left as-is per your note that it isn't a regression.

Verified on-device: icon opens the browser, card body opens the thread, and the two no longer collide.

davotoula:
Recently I've started reviewing any clonker output in a second new session... And it always, ALWAYS finds bugs that the original session didn't see.

Bit scary to be honest.

dmnyc:
image

PR #3827, opened 2026-07-31
Listen only while foregrounded, and fix the meter that measures it

davotoula:
I was got the "excess resources" on main builds twice. Going through the report... some false positive in data (correcting it), some possible resource drains.

## Issue

The in-app Resource Usage Report showed **7.13 hours of "location listening" against 9.1 seconds of app foreground**. That reads as GPS pinned on around the clock, and it was the leading suspect for the same report's 11–19 percentage points of background battery drain.

Measuring on a Pixel 9a (Android 17) turned up two separate problems — and the alarming one turned out not to be what it looked like.

**1. The meter was measuring the wrong thing.** `location.ms` counted how long a *subscription existed*, not how long anything listened. It reported 10.7 h across two days for something the OS had activated for 8 h 26 m across twenty-one.

**2. The request was genuinely wasteful, but only while foregrounded.** `LocationFlow` iterated `locationManager.allProviders` and registered on *every* one — passive, network, fused **and** gps (the last at `HIGH_ACCURACY`) — each at `MIN_TIME = 10s` / `MIN_DISTANCE = 100m`, to compute a **5 km** geohash. That subscription was then held for 45% of device uptime, because `Account` builds ~30 top-nav filter flows shared `SharingStarted.Eagerly` on the account scope, and `AccountSettings.defaultProductsFollowList` ships as `TopFilter.AroundMe` — so at least one of them always subscribes, for every user who has granted the permission. No opt-in involved.

**3. What it was *not*: a background battery drain.** The manifest declares only `ACCESS_COARSE_LOCATION` — no `ACCESS_BACKGROUND_LOCATION`, and no service with `foregroundServiceType="location"` — so Android suspends location delivery whenever the app is backgrounded. That is structural and applies to every user. Over 21 days of device uptime the OS granted Amethyst roughly **11 minutes** of background-active location, and `dumpsys location` with the app backgrounded read `gps provider: ProviderRequest[OFF]`.

The headline figure was a metric artefact. Fixing the metric first is what made everything after it measurable.

## How it was fixed

**`LocationState` gates on *permission × foreground*.** Backgrounded, it emits nothing — `stateIn` keeps the last value, so the ~60 call sites that read `account.live*FollowLists.value` synchronously keep seeing the last known geohash. Leaving the foreground is delayed by a 5 s grace period so a brief app switch doesn't tear down and rebuild the registration; returning is immediate. `Loading` is emitted only when no `Success` is cached, because `AroundMeFeedFlow.convert` maps anything that isn't `Success` to an empty geotag set — without that guard the gate would flash the "Around Me" feed empty on every return to foreground.

Switching the *consumers* to `WhileSubscribed` instead was considered and rejected: those ~60 synchronous `.value` reads would then silently serve a stale or initial value whenever no collector happened to be active. That's a correctness regression, not a battery fix.

**`LocationFlow` registers on one provider.** An ordered ladder picks it by *existence* and permission compatibility — deliberately not by `isProviderEnabled`, since a registration on a disabled provider goes live by itself when the user enables location from the quick-settings shade without leaving the app. Each rung falls through on `SecurityException`; if none registers, the flow throws. Two request profiles replace the single pair: **60 s / 500 m** for the 5 km geohash, 10 s / 100 m retained for the building-level flow.

**The listening hook moved inside `LocationFlow`** and is now paired from both ends. It fires only after a successful `requestLocationUpdates` (previously an `onStart`, which fired on *collection* even when nothing registered), and is released from a `try`/`finally` rather than `awaitClose` — the last-known-location seed can throw and unwind before `awaitClose` is ever reached, which would leak the registration and pin the refcount for the life of the process.

**`RefCountedSession`** refcounts the ledger session so that when both location flows are listening and one stops, the other's measurement segment stays open.

`LocationFlow` now takes a `LocationManager` rather than a `Context`, which is what makes the registration behaviour unit-testable.

## Testing done

**27 unit tests** across 5 files (JUnit + MockK + kotlinx-coroutines-test), covering the gate, the ladder, the registration/release pairing, and the full `LocationState → LocationFlow → RefCountedSession → ledger` composition using the production `locationSource` and a real `RefCountedSession`.

Every test was **mutation-tested** — deliberately break the implementation, confirm the test fails, restore, confirm it passes. That found **three tests passing vacuously**, including the headline `doesNotListenWhileBackgrounded`; two were fixed and one is documented in-file as inherently non-discriminating rather than left looking like coverage. Three separate test shapes had to be replaced because they passed against a broken implementation — two defeated by `callbackFlow` channel buffering, one because `removeUpdates` fires a line before the call being mutated.

**On device**, using the `benchmark` variant (`initWith(release)` — R8-minified with the shipping proguard rules) on a Pixel 9a / Android 17, installed alongside an untouched release client:

| | before | after |
|---|---|---|
| registrations, foreground | 4 (passive, network, fused, gps) | **1** (fused) |
| request profile | `@+10s0ms HIGH_ACCURACY`, `minUpdateDistance=100.0` | `@+1m0s0ms BALANCED`, `500.0` |
| registrations, backgrounded | 4, held while `mWakefulness=Dozing` | **0** |

Both clients sampled at the same instant, both backgrounded and running: the release client held four registrations — all `{bg, na} … (inactive)`, meaning the OS had already suspended delivery but the registration was still held, which is exactly the quantity `location.ms` counts — while the patched client held none.

A **clean install** (uninstall first, so the account, ledger and permission grant all start empty) then exercised the paths a dirty install can't reach:

- **No feed flash on returning to foreground.** Granted the permission, set a feed to Around Me, backgrounded for 10 s, reopened — geohash present immediately, feed populated, no blank frame. `dumpsys` shows why: a fix is delivered **1–6 ms** after each re-registration.
- **Both request profiles are live and distinct in production.** Geohash screens produce `@+10s0ms` / `minUpdateDistance=100.0`, and the aggregate's `min/max interval` moves from `60s/60s` to `10s/60s`.
- **The refcount holds under concurrent flows.** The coarse registration came up and survived **four complete precise-flow cycles** untouched, with both live simultaneously during the first. Opening and closing geohash chats never closes the "Around Me" registration.
- A sub-grace-period app switch produced **no** teardown/rebuild pair, confirming the 5 s grace.

**Ledger invariant**, from accumulated in-app report data:

| | `location.ms` | `app.fgms` | ratio |
|---|---|---|---|
| release client (before) | 25,660,172 | 9,147 | **2,805×** |
| patched, full day with permission granted | 2m10.8s | 1m56.9s | **1.12×** |

The 13.9 s excess is the grace period and sits inside the `location.ms ≤ app.fgms + 5s × transitions` allowance. The OS's own aggregate over four foreground/background cycles reads total 2m33.542s / active 2m33.456s / foreground 2m33.531s — all within 90 ms, against the release client's 9d14h20m / 8h26m / 8h14m. Registration-held time now equals foreground time, which is what makes the counter meaningful.

`compilePlayDebugKotlin`, `testPlayDebugUnitTest` and the full `test` suite are green. `lintPlayDebug` reports only pre-existing `UnrememberedMutableState` errors in an androidTest file this branch doesn't touch. `FUSED_PROVIDER` (API 31, `minSdk` 26) draws an `InlinedApi` warning rather than `NewApi`, confirming the compile-time-constant inlining the ladder's KDoc relies on.

## Impact on performance

**Foreground:** four simultaneous provider registrations become one, and the poll relaxes from 10 s / 100 m to 60 s / 500 m. At 120 km/h a 5 km geohash cell takes 2.5 minutes to cross, so the coarser profile has no observable effect on the feed. On the pre-API-31 path the ladder also removes three guaranteed-to-throw `getLastKnownLocation` binder calls — each building a `SecurityException` and an eagerly interpolated log line — on every flow start, which means on every return to foreground.

**Backgrounded:** the registration is released entirely rather than held for 45% of device uptime.

**Diagnostics:** `location.ms` goes from 2,805× `app.fgms` to 1.12×, and now reconciles with `dumpsys location` to within seconds.

**Battery — stated plainly, because the original report implied more than the data supports.** This is not a large battery win. Android was already suspending background location delivery, so the registrations being released were mostly idle. Over 7 days on the test device, location listening totalled **5m 18s** against 596 percentage points of background battery drain. Location was never a meaningful contributor, and this change does not claim to fix that drain. The real consumers visible in the same report are an always-on service running ~23.9 h/day, relay churn (3,732 relay-hours over 7 days at a 71.9% dial-failure rate), and an HTTP interceptor misattributing foreground/background time — all separate issues.

One caveat on the fixed counter: `location.ms` still measures *registration held while foreground*, not radio-on time or energy. A `network` registration at 60 s costs almost nothing; a `gps` one at 10 s costs a great deal, and the counter cannot tell them apart. Recording the chosen provider would close that, and is left as a follow-up.

## Possible pre-existing fix, unverified

Below API 31, `gps`, `passive` and `fused` required `ACCESS_FINE_LOCATION`; only `network` accepted coarse. Amethyst holds coarse only, so the previous `allProviders` loop should have thrown `SecurityException` on three of the four providers, meaning location may be **non-functional on Android 8–11** today. The new ladder is correct either way and this branch may therefore fix it, but I could not verify it without an API ≤ 30 device, so the claim is deliberately not made. The related "registrations leak" concern *is* disproved: `allProviders` yields `passive` first, so the throw would precede any registration.

## Follow-ups (not in this PR)

- `preciseGeohashStateFlow` advertises building-level precision that coarse permission makes unattainable — Android fuzzes fixes to roughly a 3 km grid.
- A `LocationResult.Unavailable` state, so a device with no usable provider stops being told "No Location Permissions" when it has them.
- Recording the chosen provider in the ledger so `location.ms` can be read as a cost signal.
- Whether the Products feed should still default to `TopFilter.AroundMe`.
- `GeohashChatScreen` sets the permission flag with a literal `true` instead of the observed value, so a revoked permission is never reflected back.

## License

      original license header.

### DIFF
--- a/amethyst/plans/2026-07-29-location-foreground-gate-impl.md
+++ b/amethyst/plans/2026-07-29-location-foreground-gate-impl.md
@@ -1,0 +1,1457 @@
+# Location Foreground Gate Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop Amethyst holding a location registration when no activity is started, register on one appropriate provider instead of four, and make `location.ms` measure real listening time.
+
+**Architecture:** A three-state gate over *permission × foreground* inside `LocationState` replaces the permission-only gate, so the registration is released whenever the app is backgrounded. `LocationFlow` selects one provider from an ordered ladder instead of shotgunning `allProviders`, and owns the `onListening` hook so accounting cannot start without a live registration. A `RefCountedSession` serialises the meter's refcount with the transition it drives.
+
+**Tech Stack:** Kotlin, kotlinx.coroutines Flow/StateFlow, `android.location.LocationManager` (no Play Services), JUnit 4 + MockK + kotlinx-coroutines-test.
+
+**Design spec:** `amethyst/plans/2026-07-29-location-foreground-gate.md` — read it before starting. This plan implements it; where the two disagree, the spec is wrong and should be corrected.
+
+## Global Constraints
+
+- Module is `amethyst` (Android only). Package root `com.vitorpamplona.amethyst`.
+- `minSdk = 26`, `targetSdk = 37`, `compileSdk = 37` (`gradle/libs.versions.toml:12-14`).
+- Every new `.kt` file starts with the MIT licence header — copy it verbatim from `amethyst/src/main/java/com/vitorpamplona/amethyst/service/location/LocationState.kt:1-20`, changing nothing.
+- Logging uses `com.vitorpamplona.quartz.utils.Log`, never `android.util.Log`. Use the lambda overload (`Log.d("Tag") { "msg $x" }`) when the message interpolates **and there is no throwable**. When a throwable must be logged, use the three-argument form `Log.w("Tag", "message", e)` — the lambda overloads take no `Throwable` (`quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/utils/Log.kt:74-79`), so "converting" such a call to a lambda silently discards the stack trace. That is the exact anti-pattern the repo's `find-non-lambda-logs` skill flags.
+- Never pass `--no-verify` to `git commit`. The pre-commit hook runs `./gradlew spotlessCheck` and nothing else — no compilation — so it passes even at the two points in this plan where the module is temporarily red.
+- No new third-party dependencies. All libraries used here are already declared.
+- Never write a fully-qualified class name inline in a function body — add an `import`.
+- Run `./gradlew spotlessApply` before every commit.
+- Do not push and do not open a PR. Commit locally only.
+- Unit tests live in `amethyst/src/test/java/...`, run with `./gradlew :amethyst:testPlayDebugUnitTest`.
+
+## Naming contract
+
+These names are used across tasks. Use them exactly.
+
+| Name | Defined in | Signature |
+|---|---|---|
+| `RefCountedSession` | Task 2 | `class RefCountedSession(setSessionActive: (Boolean) -> Unit)`, method `fun setActive(active: Boolean)` |
+| `LocationProviderLadder.chooseProviders` | Task 3 | `fun chooseProviders(sdkInt: Int, hasFine: Boolean, exists: (String) -> Boolean): List<String>` |
+| `LocationFlow` | Task 4 | `class LocationFlow(locationManager: LocationManager, sdkInt: Int = Build.VERSION.SDK_INT, hasFine: Boolean = false)`, method `fun get(minTimeMs: Long, minDistanceM: Float, onListening: ((Boolean) -> Unit)? = null): Flow<Location>` |
+| `LocationState` | Task 5 | `class LocationState(context: Context, scope: CoroutineScope, isForeground: StateFlow<Boolean>, onListening: ((Boolean) -> Unit)? = null, locationSource: (Long, Float) -> Flow<Location> = …)` |
+| `LocationState.COARSE_MIN_TIME` | Task 5 | `const val = 60_000L` |
+| `LocationState.COARSE_MIN_DISTANCE` | Task 5 | `const val = 500.0f` |
+| `LocationState.PRECISE_MIN_TIME` | Task 5 | `const val = 10_000L` |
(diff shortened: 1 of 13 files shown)


vitorpamplona:
Good that this report is being helpful :)

davotoula:
Very detailed. 

Data consumption next, let's see if anything can be done about relay storms...

vitorpamplona:
Hope this one can help us better understand and see which subscription is being used where: 

https://github.com/vitorpamplona/amethyst/pull/3832

Feel free to add more tracking elements if needed.

PR #3862, opened 2026-08-05
SyncCoverage: one band interval cannot speak for several kinds

vitorpamplona:
## Summary

A `SyncCoverage` band held ONE `created_at` interval per `(relay, filter)`. For a filter naming several kinds that is a claim no walk can support: ask for `kinds: [0, 30382]`, find profiles going back years and score cards only from last month, and the band records `2020..now` for the pair — so the next run skips that whole interior for **both**, and score cards written inside it are never asked for again. A long-lived kind vouched for a short-lived one, silently.

`Band.spans` is now per kind, so each carries only the evidence actually collected for it. This is a correctness fix, not a feature: the failure mode is lost events with nothing on any log line to say so.

## What keeps the cost where it was

- **`legs()` regroups kinds by the windows they want.** Identical coverage — the common case, and the only case until kinds diverge — collapses back into one ask, so a filter that produced two legs still produces two rather than two per kind. Only a kind whose evidence genuinely differs earns its own.
- **A finished reconcile needs no per-kind evidence and is given none.** Negentropy compares the filter's whole id set in one pass, so it covers every kind in the filter or none. Only the **paged** path changed.
- **Filters naming no kinds** keep a single span under `ALL_KINDS` — the same claim as before, correctly scoped to the case where it is the only claim available.

`record()` gains `observedByKind`, and `SyncCoverage.observe()` accumulates it as events arrive — replacing the pair of hand-rolled `var`s each caller kept and folding the per-event `isPlausible` guard in with it.

A paged walk over a **multi-kind** filter that supplies none earns no band at all, loudly, once. Attributing one interval to every kind is exactly the over-claim being removed, and a band that over-claims skips events silently, which is worse than re-reading them. Single-kind filters are untouched — there the aggregate always was the per-kind answer.

### State file

Gains a per-kind `spans` object and keeps `min`/`max` as the outer edges, so a **rollback** to a binary from before this reads the file and behaves as it always did. A file written *before* this loads its one interval under `ALL_KINDS` — the old, wider claim, kept rather than discarded because discarding it would re-download every upstream's corpus once on upgrade. The first per-kind walk replaces it.

### Downstream note

Callers with multi-kind paged streams must pass `observedByKind` to keep resuming; `MirrorWorker` is migrated in this PR. Anything else on multi-kind paged filters will re-walk from scratch (correct, but not free) until it does the same.

## Test plan

- [x] Ran the relevant modules' tests
- [ ] Manually exercised the change — see "Not done" below

```
$ ./gradlew :quartz:jvmTest :geode:test
BUILD SUCCESSFUL in 5m 32s
```

**All 26 pre-existing `SyncCoverageTest` cases pass unchanged**, which is the evidence single-kind behaviour did not move. Seven new cases cover the multi-kind interior, regrouping, the refusal, the reconcile path, the `ALL_KINDS` restore, and the two review findings below.

**The new behavioural tests were mutation-tested.** A straight revert does not compile (geode uses the new API), so the pre-fix rule was reinstated *in place* — `span = Span(band.minCreatedAt, band.maxCreatedAt)` in `legs()`, and the multi-kind refusal disabled:

```
SyncCoverageTest[jvm] > a multi-kind paged walk with no per-kind evidence earns no band[jvm] FAILED
SyncCoverageTest[jvm] > a long-lived kind no longer vouches for a short-lived one[jvm] FAILED
31 tests completed, 2 failed
```

Both pass with the fix in place.

**One flavour only, deliberately.** Per `CONTRIBUTING-WITH-AI.md` § Build and install both flavours: this is a pure `quartz/` protocol change plus its `geode/` (JVM server) caller. It touches no UI, no manifest, no Play-only dependency, and nothing under `amethyst/`. There is no flavour axis for it to differ along.

## Code review pass

Required by `CONTRIBUTING-WITH-AI.md`. A review pass over the first commit found **two** defects, both fixed in `74145ee` as their own commit rather than folded into the feature commit:

1. Spans for kinds the filter never named were stored as given — inert for `legs()`, but not for `Band.minCreatedAt`, which is what the state file writes as its rollback-compat `min`/`max`. A relay answering with more than it was asked for would push that floor below anything the filter's kinds support, so a rolled-back binary would read the file and over-claim. *This fix, undone through the compatibility path it added.*
2. `observedByKind` on a filter naming **no** kinds was stored per kind, while `legs()` for such a filter reads only `ALL_KINDS`. The band was recorded, persisted, and never consulted — a resume that silently did not resume.

Both live where the new per-kind path meets an old assumption: that `record()`'s input is already scoped to the filter, and that a band's keys are always the filter's kinds. Neither holds once callers supply the map themselves.

**Not done, and worth a maintainer's eye:** the gate asks for that review to come from a *different* agent or model. This one did not — it was a self-review by the same assistant, which the gate exists precisely to distrust. It found two real bugs, so it was not worthless, but it is not the independent pass the rule asks for. Flagging rather than ticking the box.

## Interop suites

`SyncCoverage` decides which `since`/`until` windows to *ask* for. It does not construct, parse, or alter any event.

## AI assistance

Written by Claude Code. Origin: `SyncCoverage` was recently upstreamed from NosFabrica/vespa-relay (#3855), which still carried a fork of it; reconciling the two surfaced this as the one open defect in the shared version. The bug is live in a real multi-kind deployment (`kinds: [0, 10002, 30382]` on a paged stream), not hypothetical.

## License

---
_Generated by [Claude Code](https://claude.ai/code/session_01BbLjD9kvo9X6875VMR1Ghs)_

### DIFF
--- a/geode/src/main/kotlin/com/vitorpamplona/geode/mirror/MirrorWorker.kt
+++ b/geode/src/main/kotlin/com/vitorpamplona/geode/mirror/MirrorWorker.kt
@@ -498,6 +498,12 @@ class MirrorWorker(
                 val syncStartedAt = TimeUtils.now()
                 var seenMin: Long? = null
                 var seenMax: Long? = null
+                // Per KIND as well as in aggregate: one interval for a
+                // multi-kind filter lets a long-lived kind vouch for a
+                // short-lived one, and the band then skips the interior for
+                // both. The aggregate is still tracked because the reconcile
+                // path records against the leg's floor, not per kind.
+                val seenByKind = mutableMapOf<Int, SyncCoverage.Span>()
 
                 fun observe(event: Event) {
                     // Same containment as the live path: even a trusted
@@ -509,6 +515,7 @@ class MirrorWorker(
                             seenMin = minOf(seenMin ?: event.createdAt, event.createdAt)
                             seenMax = maxOf(seenMax ?: event.createdAt, event.createdAt)
                         }
+                        SyncCoverage.observe(seenByKind, event.kind, event.createdAt)
                         handoff.trySendBlocking(event)
                     } else {
                         filtered.incrementAndGet()
@@ -537,6 +544,7 @@ class MirrorWorker(
                     } catch (e: NegentropySyncException) {
                         seenMin = null
                         seenMax = null
+                        seenByKind.clear()
                         // The watchdog matches negentropySync's default rather
                         // than fetchAllPages' shorter one: a paged catch-up
                         // sits behind the same slow upstreams.
@@ -558,6 +566,13 @@ class MirrorWorker(
                         seenMin,
                         seenMax?.coerceAtMost(syncStartedAt),
                         paged = true,
+                        // Capped the same way the aggregate is: one
+                        // future-dated event must not lift a kind's ceiling
+                        // past what was actually asked for.
+                        observedByKind =
+                            seenByKind.mapValues { (_, span) ->
+                                SyncCoverage.Span(span.min, span.max.coerceAtMost(syncStartedAt))
+                            },
                     )
                 } else {
                     val legFloor = leg.since ?: initialSince
--- a/geode/src/main/kotlin/com/vitorpamplona/geode/mirror/SyncCoverageFile.kt
+++ b/geode/src/main/kotlin/com/vitorpamplona/geode/mirror/SyncCoverageFile.kt
@@ -100,8 +100,7 @@ class SyncCoverageFile(
                 root.mapValues { (_, v) ->
                     val o = v.jsonObject
                     SyncCoverage.Band(
-                        o.getValue("min").jsonPrimitive.long,
-                        o.getValue("max").jsonPrimitive.long,
+                        spansOf(o),
                         o["complete"]?.jsonPrimitive?.boolean ?: false,
                         o["fullAt"]?.jsonPrimitive?.long ?: 0L,
                     )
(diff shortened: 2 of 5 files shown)


vitorpamplona:
## Deep audit pass — findings

Pushed `bb95cad`. Three defects fixed, several things checked and found sound, and **one open question about the premise that should probably be settled before merge.**

### Fixed

1. **`record()` read the clock twice per kind.** A 40-kind map took 80 readings, and a span's floor and ceiling were judged against *different* instants — so a span could be accepted at one end and rejected at the other on a tick. One read per call now.
2. **`legs()` published its mutable accumulator.** The same `MutableList` instance was handed to every `Filter` in a group. Immutable in practice, but through a public return value.
3. **The file round trip was asserted for fields, not behaviour.** Three tests now pin it: per-kind spans survive a restart *and still narrow per kind afterwards*; the `ALL_KINDS` sentinel survives its negative key through `toString`/`toInt`; a pre-split file loads as the claim it always was; and `min`/`max` must remain the **outer** edges, or a rolled-back binary skips ground it hasn't covered.

### Checked, sound

`ConcurrentMap.snapshot()` copies, so `export()` can't be mutated under a writer. `Band` is immutable (`widen()` copies its map). `merge()` keeps `old.fullAt`, preserving the re-walk clock. And **`coveringWindow` does not regress** — verified by probe, not reasoning: a paged band gave >1 leg before this change too, and a reconciled band still collapses to one leg and narrows the shared snapshot.

### Known cost

A kind the relay has **no events of** never earns a span, so it gets a full-range leg every cycle — one REQ that returns nothing (`fetchAllPages` breaks on `received == 0` after one page). It self-heals the moment any negentropy reconcile runs, since a reconcile writes a span for every kind in the filter. Only pure-paged (`fetch`-mode, no NIP-77) streams keep paying it. Cheap, but real, and worth knowing about.

### ⚠️ Open question: I could not reproduce the original bug

Auditing `fetchAllPages` to check the cost above, I could not construct the loss mechanism this PR is premised on, and I'd rather say so than let it merge on my say-so.

The pager advances `until = pageMinTs` — **the oldest second seen across the whole filter** — and a relay answering a limited REQ returns the newest N matching events *for the whole filter*, not per kind. So each page appears to cover its window for every kind simultaneously: a sparse kind can't be "crowded out" of a page by a dense one, because any sparse-kind event newer than `pageMinTs` would have been among the newest N. The walk then terminates on `received == 0`, i.e. genuinely drained.

If that reading is right, then for the paged path a single band interval *was* a sound claim for every kind in the filter, and this PR adds per-kind bookkeeping plus the recurring cost above for no correctness gain.

I may well be missing the case. Candidates I couldn't confirm from the code: relays that don't order strictly newest-first under a limit; relays applying per-kind caps or per-kind retention; or a mechanism specific to how a caller composes legs. **You'd know the original observation better than the code does — if you can name the reproduction, I'll add it as a failing test and the premise is settled.** If it turns out the paged path is sound, the honest outcome is to close this rather than carry the complexity.

The two `record`/`legs` defects fixed in `74145ee` are real regardless of how that resolves — they're bugs in this PR's own new code, not the original.


---

PR #3866, opened 2026-08-05
Make the left drawer configurable

davotoula:
Some users might prefer less features:

Screenshot_20260805-214959 Screenshot_20260805-215014

Adds Settings → Side Menu, a show/hide editor for the left drawer's rows, mirroring the existing Bottom Navigation Bar picker. Sections and item order are fixed; visibility is the only per-account choice.

Both screens now render `DrawerSections` — the same list the drawer itself renders — so a destination added to a section appears in the drawer and in its settings screen with no extra work, and `DrawerSectionsTest` fails the build if a new `NavBarCatalog` id isn't filed into a section.

The preference stores what is **hidden**, not what is shown. That is what lets a destination shipped in a later release appear for every account with no migration; storing the visible list would freeze each account's drawer at the moment they first opened the screen. It persists as enum names rather than the enum itself, so an unknown id from a newer client is dropped on read instead of throwing and taking the whole synced-settings blob down with it.

Settings is the one row that cannot be hidden — it is the route back to this screen. It renders a locked "Always on" badge rather than a toggle that ignores taps. Drawer chrome (profile header, relay status, account switcher, version footer) stays fixed; those aren't catalog rows and have no id to store.

Also folds the two picker screens onto shared UI in `NavPickerUi.kt` (`CatalogRow`, `ExpandedKeys`), replaces `categoryIcon()`'s string-resource `when` with a declared `icon` field on `NavBarCategory`, and adds `FAVORITE_ALGO_FEEDS` to the You section — the one catalog destination no drawer section listed, surfaced by the new coverage test.

Per-account, synced across devices over the NIP-78 app-specific data event.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -150,6 +150,7 @@ import com.vitorpamplona.amethyst.service.relayClient.reqCommand.nwc.NWCPaymentF
 import com.vitorpamplona.amethyst.service.uploads.FileHeader
 import com.vitorpamplona.amethyst.ui.actions.NewMessageTagger
 import com.vitorpamplona.amethyst.ui.navigation.bottombars.BottomBarEntry
+import com.vitorpamplona.amethyst.ui.navigation.bottombars.NavBarItem
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.EventProcessor
 import com.vitorpamplona.quartz.buzz.threading.buzzThread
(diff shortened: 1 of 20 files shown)


vitorpamplona:
Woudln't a Switch toggle make more sense instead of this butttons? YOu can also ask the AI to make the UI, "modern, beautiful and exciting" like some of the other parts of settings you might want to copy. That prompt always works for me :)

davotoula:
I like the View button but happy to switch to switch if you prefer. Maybe better for consistency?

vitorpamplona:
let's keep it as is, then

PR #3868, opened 2026-08-05
fix(media): stop rendering non-media NIP-94 files as video

davotoula:
Before:
Screenshot_20260805-223323

After:
Screenshot_20260805-223020 Screenshot_20260805-223004

What was wrong

A NIP-94 kind-1063 file header announcing a webxdc app (m application/x-webxdc, a .xdc zip bundle) rendered as a video player that buffered forever. Same for any archive, installer, or other non-media blob — and for PDFs, which had a working viewer that this path never reached.

Why

The kind-1063 renderers classified content with a single boolean: mimeType.startsWith("image/") || isImageUrl(url). Anything that wasn't an image fell through an unconditional else into MediaUrlVideo, so ExoPlayer was handed a zip. There was no isVideo test at all. The inline-URL path (RichTextParser.createMediaContent) already classified properly and returned null for unknown types, so the same URL pasted into a kind-1 note rendered as a link while the kind-1063 announcement rendered as a player.

How it's fixed

Adds RichTextParser.classifyMedia(url, mimeType): MediaContentKind? — MIME first, URL extension as fallback, and null when no viewer can render the blob. createMediaContent now delegates to it so the two paths can't drift; a test asserts they agree.

Kind-1063 renderers route the four cases through FileHeaderEvent.toMediaContent(), with unknown types landing on a new FileAttachmentCard (icon, alt text, application/x-webxdc · 17 MB, tap to open) built entirely from the event's own tags — no network round-trip to rediscover a type the event already declared. PDFs now reach PdfPreviewCard.

The NIP-71 video renderers keep defaulting to the player, since kind 21/22 asserts the type itself; that default is now explicit rather than a fallthrough.

Not covered: encrypted file headers (kind 15) have the same defect but can't be fixed until MediaContentModels gains an EncryptedMediaUrlPdf.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/FileAttachmentCard.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/FileAttachmentCard.kt
@@ -1,0 +1,142 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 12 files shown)


vitorpamplona:
hah.. funny thing.. now that we have a browser running in a separate process, we can render the result of that browser into the post itself.. we can probably render x-webxdc inline.

davotoula:
That would be pretty awesome... But I don't think it would work for all file header types.

Want me to apply browser override for xdc. Doom inside amethyst? 😂

vitorpamplona:
if you want to spend tokens with it, otherwise I can merge and work on that later

davotoula:
Yes please, let's merge the fix now.

Smaller pr if we want to go browser route later.



--
David Kaspar

PR #3873, opened 2026-08-07
feat(concord): staff-held control_root write-gates the Control Plane (CORD-02 §2)

alexgleason:
## Summary

Brings Amethyst's Concord implementation up to the merged spec change
[concord-protocol/concord#15](https://github.com/concord-protocol/concord/pull/15)
(CORD-02 §2, *Write-Restricted Streams*): the Control Plane's stream key
splits, so its signer keypair now derives from a new staff-held
`control_root` while every member holds only the derived `control_pk` to
subscribe, verify, and read under the unchanged `community_root`-derived
read key. Previously any member could sign wraps at the Control Plane
address and flood the one plane every member must sync in full; possession
of the secret is a spam gate, never authority — every edition is still
judged by its sealed actor's rank in the owner-rooted Roster.

Delivery follows the spec's three paths: the secret rides the staff-making
Grant itself (`control_wrap`), syncs across a holder's own devices in the
kind-13302 Community List, and rolls with the `community_root` at every
Refounding. Legacy pre-split epochs stay readable, and a legacy community
upgrades to the split as a side effect of its next base rotation.

## Test plan

Platform: Linux (Pop!_OS, 6.x), JDK 21, `./gradlew` from a clean tree.

### Feature test plan

**Automated** — `./gradlew test` (all modules) green. Concord-specific
counts from `build/test-results/jvmTest`, including the new suites
`ControlPlaneSplitTest` (7), `ControlRootWrapTest` (7),
`ControlRootRotationTest` (5):

```
:quartz:jvmTest   ControlPlaneSplitTest        tests=7  failures=0 errors=0 skipped=0
                  ControlRootWrapTest          tests=7  failures=0 errors=0 skipped=0
                  ControlRootRotationTest      tests=5  failures=0 errors=0 skipped=0
                  ConcordCommunityListTest     tests=16 failures=0 errors=0 skipped=0
                  AuthorityResolverTest        tests=23 failures=0 errors=0 skipped=0
                  ConcordRefoundingTest        tests=4  failures=0 errors=0 skipped=0
                  ConcordCommunityFactoryTest  tests=3  failures=0 errors=0 skipped=0
                  (+ 20 further concord suites, all 0 failures)
:commons:jvmTest  ConcordModerationTest        tests=5  failures=0 errors=0 skipped=0
                  ConcordActionsTest           tests=4  failures=0 errors=0 skipped=0
                  ConcordSubscriptionPlannerTest tests=6 failures=0 errors=0 skipped=0
                  (+ 10 further concord suites, all 0 failures)
```

**End-to-end on the wire** — two `amy` identities against a loopback
`geode` relay (`ws://127.0.0.1:7448`), driving the real publish path:

1. `amy concord create --name CordTwo` — genesis mints the `control_root`
   beside the `community_root`.
2. `amy concord invite` → `amy concord join <url>` as a second identity.
3. Member attempts a Control Plane write; staff performs the same write.

Stored state after the join — the owner holds both keys, the member holds
the address only, and both share one `community_root`:

```
ALICE (owner/staff)                     BOB (member, joined by invite)
controlPk   3a8e0511…1998d4c7           controlPk   3a8e0511…1998d4c7
controlRoot fb41060e…6ad613d0           controlRoot ""
root        bdcab410…2d1e25cc           root        bdcab410…2d1e25cc
```

The member's Control write is refused rather than missigned (exit 1):

```
$ amy --json concord role <community> spammer 5 BAN
{"error":"forbidden","detail":"this account holds no control_root for the
 community, so it cannot publish Control Plane editions (CORD-02 §2) —
 ask a staff member to grant you a Control-writing role"}
exit=1
```

Staff's writes land, and the fold seats them:

```
$ amy --json concord role <community> moderator 5 BAN
{"role_id":"960c2dfe…fe6a8747","name":"moderator","position":5,"published_to":["ws://127.0.0.1:7448/"]}
$ amy --json concord grant <community> <bob-hex> 960c2dfe…fe6a8747
{"member":"f9bda1e7…1b948cf2","roles":["960c2dfe…fe6a8747"],"published_to":["ws://127.0.0.1:7448/"]}
$ amy --json concord roles <community>
{"roles":[{"id":"960c2dfe…","name":"moderator","position":5,"permissions":"16"}],
 "grants":[{"member":"f9bda1e7…","rank":5,"roles":["moderator"]}],"banned":[]}
```

**Wire-format verification** (`nak` against the same relay) — all four
Control Plane wraps (metadata, `#general`, the role, the Grant) are
authored by the `control_root`-derived `control_pk`, and there is no second
plane at the legacy `community_root`-derived address:

```
$ nak req -k 1059 ws://127.0.0.1:7448 | jq -r .pubkey | sort | uniq -c
      4 3a8e05118d17e222e4506b7a8e006a293f1318a969c47aaafed7a71d1998d4c7

$ nak req -k 1059 -a 3a8e0511…1998d4c7 ws://127.0.0.1:7448 | jq -c '{id, pubkey, tags: [.tags[][0]]}'
{"id":"4f97ae1770c97775…","pubkey":"3a8e05118d17e222…","tags":["p"]}
{"id":"18560ba9a0eeb689…","pubkey":"3a8e05118d17e222…","tags":["p"]}
{"id":"378fbcc80df2ea0b…","pubkey":"3a8e05118d17e222…","tags":["p"]}
{"id":"e44d27270996dd83…","pubkey":"3a8e05118d17e222…","tags":["p"]}
```

That author equals the owner's stored `controlPk` exactly, and the outer
wraps carry only the ephemeral `p` tag (no version tag, no new outer tag —
CORD-02 Appendix B). `ControlPlaneSplitTest` asserts the complementary
negative that can't be shown from outside the keys: the split address is
never equal to the legacy address, a member-signed wrap does not open at
the plane, and wrapping without the write key throws instead of
missigning.

### Regression test plan

Touch points, the failure each would show, and what was verified:

- **Legacy (pre-split) communities** — could become unreadable, since the
  plane's address moved. Failure mode: an existing community's Control
  Plane folds empty and every role/ban disappears. Verified by
  `ControlPlaneSplitTest.aLegacyEpochStaysReadableAfterTheSplitExists` and
  `heldSecretsSelectTheViewOfAnEpoch`: with no `control_pk` held, the
  legacy `concord/control` derivation is still address, signer, and read
  key at once, and reading a legacy epoch does not consult the split.
- **Refounding / rekey blobs** — width change (72 → 104/136) could break
  every existing rotation. Failure mode: retained members decode nothing
  and are silently severed. Verified by `ControlRootRotationTest` and
  `ConcordRefoundingTest`: the 72-byte legacy base blob is still honored on
  read (never minted anew), member blobs carry the pk, staff blobs the
  secret, and a staff pair whose secret doesn't derive to the delivered pk
  is refused.
- **Community List merge across devices** — could inherit stale key
  material across epochs. Failure mode: a device adopts a prior epoch's
  `control_pk` and subscribes to a dead address. Verified by
  `ConcordCommunityListTest` (16): control material is filled only on a
  *same-epoch* merge, never inherited across epochs, and the unrelated
  `invite_ref` stranded-recovery anchor still survives every merge branch.
- **Anti-rollback floor / historical epochs** — prior-epoch Control Planes
  are folded to rebuild the floor. Failure mode: floor rebuilt from nothing
  and superseded editions replay. Verified by `ConcordRollbackFloorTest`
  and `ConcordCommunitySessionTest`: a banked epoch is re-subscribed at the
  `control_pk` remembered with it (a split address can never be
  re-derived), legacy held epochs still derive theirs.
- **NIP-42-gated relays** — `streamKeys()` now contributes the Control key
  only when we hold the signer. Failure mode: a member's control
  subscription starves and channels go empty. Verified by
  `ConcordSubscriptionPlannerTest` (6) and exercised live above: the member
  identity folds the plane fine over an unauthenticated relay. **Not**
  verified against a relay that actually gates the plane's REQ on stream-key
  AUTH — I don't have one; a member cannot answer that challenge by
  construction (it needs the secret), which is the write-restriction
  working as designed, but the resulting member-side behaviour on such a
  relay is untested. Flagging rather than guessing.
- **Moderation verbs (Android + amy)** — every verb now needs the write
  key. Failure mode: a rank-qualified staffer whose secret hasn't arrived
  yet crashes out of the envelope instead of no-opping. Verified by
  `ConcordModerationTest` (5) and the live `forbidden` refusal above; the
  Android path returns `false` through `controlKeysForWrite` and logs,
  and the UI hides the affordance on the same predicate.
- **Both flavours** — `./gradlew :amethyst:assemblePlayDebug
  :amethyst:assembleFdroidDebug` → `BUILD SUCCESSFUL in 3m 11s`. No
  Google-proprietary touch points in this diff; the change is
  flavour-irrelevant, both built anyway.
- **Static analysis** — `./gradlew :amethyst:lintPlayDebug` →
  `BUILD SUCCESSFUL in 13m 24s`, no new findings.
- **Formatting** — `./gradlew spotlessApply` → no files reformatted.

### Not verified — please weigh before merging

- **No on-device run, no screenshots, no `installPlayBenchmark`.** No
  Android device or emulator was available on this machine (`adb devices`
  empty). The Android surface of this diff is two `canWrite` conjunctions
  in `ConcordChannelListScreen.kt` that *hide* existing affordances plus
  the `AccountConcordActions` write guards; nothing new is drawn and no
  reflection-touched code path is added, so I'd expect the R8 build to be
  unaffected — but that is reasoning, not a test, and CONTRIBUTING asks for
  the screenshots regardless. Happy to run these if a maintainer wants them
  before review.
- **Multi-device sync of the new state was not exercised on two real
  devices.** The `control_root` now rides the kind-13302 Community List
  (CORD-02 §8). The merge is covered by unit tests, not by two live
  clients.

## Interop suites

- [x] N/A — no listed suite covers Concord. This change cannot affect
      MLS/Marmot state, NIP-17 DM envelopes, decoded audio, MoQ-lite, or
      QUIC wire bytes; it touches only the Concord packages in `quartz/`,
      `commons/`, `cli/`, and the Concord screens in `amethyst/`. In place
      of a suite, the bespoke `amy` ↔ `amy` ↔ `geode` end-to-end above
      exercises the real publish/subscribe path and the bytes on the wire.

## Cross-client compatibility

This changes how Amethyst addresses an existing plane, so it is worth
being precise:

- **Against the spec.** Implements merged `concord#15`, including the
  review follow-up (`bbc67b6`): `control_root` syncs in the Community List,
  and CORD-04 §3's staff bit list is treated as normative — `STAFF_BITS` is
  exactly `MANAGE_ROLES, MANAGE_CHANNELS, MANAGE_METADATA, BAN,
  CREATE_INVITE, PIN_MESSAGES`, with a comment forbidding extension without
  a spec amendment. `PIN_MESSAGES` claims frozen bit 11.
- **Older Amethyst builds.** A pre-split build reads a split community's
  Control Plane only at the legacy address, which is empty after an
  upgrade, so it sees no metadata/roles/bans. This is the spec's intended
  one-way upgrade (CORD-02 §5 "the first compliant base rotation upgrades a
  Community to the split"), and CORD-06 §3 explicitly forbids mirroring
  editions to the legacy address to appease stale readers — the mirror
  would re-open the member-writable surface the split closes. Communities
  created by older builds keep working until their next base rotation.
- **Other clients.** Armada (Soapbox) implements the same change — the
  spec PR links its proposal. No other client implements Concord today.
- **Reading direction.** A split-aware build reads legacy communities and
  legacy epochs unchanged, so upgrading Amethyst first is safe.

## Known gap

`amy` has no counterpart to Android's `drainConcordStaffGrants`, so a
member promoted to staff *through the CLI* stays read-only until the key
reaches them another way. The adoption logic currently lives in
`AccountConcordActions` (Android). Extracting it into `commons/` so the CLI
can adopt too is a natural follow-up; I left it out to keep this PR to the
spec change. Called out so it isn't mistaken for the write gate misfiring.

## AI assistance

Per `CONTRIBUTING-WITH-AI.md` § *Code review pass*, the diff went through a
separate review pass under a different prompt from the one that wrote it,
and `./gradlew lint` on top; the tests above were re-run afterwards.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -3563,6 +3563,9 @@ class Account(
                 refreshConcordChannelIndex()
                 // A revision also bumps when a base-rotation rekey lands; adopt ours if present.
                 runCatching { concord.drainConcordRekeys() }.onFailure { Log.w("Concord", "rekey drain failed", it) }
+                // A promotion to staff delivers the Control Plane write key inside the Grant
+                // itself (CORD-04 §3), so the fold that seats the role is also when it arrives.
+                runCatching { concord.drainConcordStaffGrants() }.onFailure { Log.w("Concord", "staff grant drain failed", it) }
                 // A rotation we were *excluded* from produces no rekey to drain, so it can only be
                 // found by re-resolving the invite link we joined through. Rate-limited internally.
                 runCatching { concord.recoverStrandedConcordCommunities() }.onFailure { Log.w("Concord", "stranded recovery failed", it) }
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountConcordActions.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountConcordActions.kt
@@ -24,20 +24,28 @@ import com.vitorpamplona.amethyst.commons.actions.ConcordActions
 import com.vitorpamplona.amethyst.commons.actions.ConcordModeration
 import com.vitorpamplona.amethyst.commons.actions.ConcordSubscriptionPlanner
 import com.vitorpamplona.amethyst.commons.model.concord.ConcordChannel
+import com.vitorpamplona.amethyst.commons.model.concord.ConcordCommunitySession
 import com.vitorpamplona.amethyst.commons.viewmodels.ReplyMode
 import com.vitorpamplona.amethyst.ui.screen.loggedIn.chats.publicChannels.concord.concordChannelLastReadRoute
+import com.vitorpamplona.quartz.concord.cord02Community.ConcordCommunityList.withControlRoot
 import com.vitorpamplona.quartz.concord.cord02Community.ConcordCommunityListEntry
 import com.vitorpamplona.quartz.concord.cord02Community.ConcordCommunityListEvent
 import com.vitorpamplona.quartz.concord.cord02Community.HeldRoot
 import com.vitorpamplona.quartz.concord.cord02Community.ImagePointer
 import com.vitorpamplona.quartz.concord.cord03Channels.ChannelChat
 import com.vitorpamplona.quartz.concord.cord04Roles.ChannelEntity
+import com.vitorpamplona.quartz.concord.cord04Roles.ConcordJson
 import com.vitorpamplona.quartz.concord.cord04Roles.ConcordPermissions
+import com.vitorpamplona.quartz.concord.cord04Roles.ControlEntityKind
+import com.vitorpamplona.quartz.concord.cord04Roles.ControlRootWrap
+import com.vitorpamplona.quartz.concord.cord04Roles.GrantEntity
 import com.vitorpamplona.quartz.concord.cord04Roles.MetadataEntity
 import com.vitorpamplona.quartz.concord.cord04Roles.RoleEntity
 import com.vitorpamplona.quartz.concord.cord05Invites.CommunityInvite
 import com.vitorpamplona.quartz.concord.cord05Invites.InviteBundleStatus
 import com.vitorpamplona.quartz.concord.cord05Invites.InviteRelayDictionary
+import com.vitorpamplona.quartz.concord.crypto.ConcordKeyDerivation
+import com.vitorpamplona.quartz.concord.crypto.ControlPlaneKeys
 import com.vitorpamplona.quartz.concord.crypto.GroupKey
 import com.vitorpamplona.quartz.concord.envelope.ConcordStreamEnvelope
 import com.vitorpamplona.quartz.nip01Core.core.Event
@@ -138,6 +146,10 @@ class AccountConcordActions(
                 ownerSalt = community.ownerSalt.toHexKey(),
                 root = community.communityRoot.toHexKey(),
                 rootEpoch = community.rootEpoch,
+                // The creator is the founding staff member (CORD-02 §2): it keeps the write
+                // secret and publishes only the derived pubkey to everyone else.
+                controlPk = community.controlPkHex,
+                controlRoot = community.controlRoot.toHexKey(),
                 relays = relayUrls,
                 name = name,
                 addedAt = TimeUtils.now() * 1000,
@@ -168,6 +180,9 @@ class AccountConcordActions(
                 rootEpoch = entry.rootEpoch,
                 name = entry.name,
                 relays = entry.relays,
+                // The joiner can never derive the Control Plane address, so the bundle carries
+                // it (CORD-05 §1). Null on a legacy community, which has none to carry.
+                controlPk = entry.controlPk,
             )
         val minted = ConcordActions.mintInviteLink(base, invite, TimeUtils.now(), entry.relays)
 
@@ -245,6 +260,9 @@ class AccountConcordActions(
                 ownerSalt = bundle.ownerSalt,
                 root = bundle.communityRoot,
                 rootEpoch = bundle.rootEpoch,
+                // Read access to the Control Plane, never write (CORD-05 §1). Absent = the
+                // community is still pre-split, so we fold it at the legacy address.
+                controlPk = bundle.controlPk,
                 relays = bundle.relays,
                 name = bundle.name,
                 addedAt = TimeUtils.now() * 1000,
@@ -451,6 +469,24 @@ class AccountConcordActions(
     // every client's AuthorityResolver, so a call by someone who doesn't outrank the
     // target is simply dropped on fold. Owner-authored calls always take effect.
 
+    /**
+     * The Control Plane keys for a moderation write, or null when this account cannot
+     * publish there: on a split epoch only `control_root` holders can mint a wrap that
+     * verifies at the plane's address (CORD-02 §2), and wrapping without the secret
+     * throws rather than missigning. Rank and key possession can diverge — a freshly
+     * promoted staffer writes only once their `control_wrap` is adopted (CORD-04 §3),
+     * and the UI gates on rank — so every moderation verb no-ops through this check
+     * instead of crashing on a rank-gated action.
+     */
+    private fun controlKeysForWrite(session: ConcordCommunitySession): ControlPlaneKeys? {
+        val cp = session.controlPlaneKeys()
+        if (!cp.canWrite) {
+            Log.w("Concord") { "Control write refused for ${session.entry.id}: control_root not held at epoch ${session.entry.rootEpoch} (CORD-02 §2)" }
+            return null
+        }
+        return cp
+    }
+
     /** Grant [member] exactly [roleIds] (empty list revokes their roles). */
     suspend fun grantConcordRole(
         communityId: String,
@@ -459,7 +495,23 @@ class AccountConcordActions(
     ): Boolean {
         val session = account.concordSessions.sessionFor(communityId) ?: return false
         if (!account.isWriteable()) return false
-        val wrap = ConcordModeration.grant(account.signer, session.controlPlaneKey(), communityId.hexToByteArray(), member, roleIds, session.controlEditions(), TimeUtils.now(), owner = session.entry.owner)
+        val cp = controlKeysForWrite(session) ?: return false
+        // A Grant that first makes its member staff must deliver the control_root in the same
+        // edition (CORD-04 §3) — grantWithStaffDelivery attaches the pairwise wrap when the
+        // roles carry a Control-writing bit and we hold the secret to hand over.
+        val wrap =
+            ConcordModeration.grantWithStaffDelivery(
+                actor = account.signer,
+                controlPlane = cp,
+                communityId = communityId.hexToByteArray(),
+                member = member,
+                roleIds = roleIds,
+                current = session.controlEditions(),
+                createdAt = TimeUtils.now(),
+                owner = session.entry.owner,
+                controlRoot = session.entry.controlRoot?.hexToByteArray(),
+                epoch = session.entry.rootEpoch,
+            )
         publishConcordWrap(session.entry, wrap)
         return true
     }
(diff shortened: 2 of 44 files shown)


vitorpamplona:
tested or do I need to test?

davotoula:
@alexgleason can you ask your clonker for a manual test plan if you have the session still available please?

alexgleason:
I didn't test it on a real device yet. If Concord chats still work then it should be good. Should also test creating a community.

davotoula:
> I didn't test it on a real device yet. If Concord chats still work then it should be good. Should also test creating a community.

I meant for you to ask your clonker for a manual test plan that I can run on a device myself. 

I find that clonker that just did the code can write a decent manual test plan...

alexgleason:
@davotoula — here you go. Written by the assistant that wrote the diff, so
treat the "expect" lines as claims to falsify rather than gospel.

**Already covered, don't bother redoing:** `./gradlew test` (0 failures; 27
Concord suites in `quartz`, 13 in `commons`, incl. the three new ones),
`assemblePlayDebug` + `assembleFdroidDebug`, `lintPlayDebug`, and an `amy` ↔
`amy` end-to-end over a loopback `geode` where a member's Control Plane
write is refused and staff's lands, with `nak` confirming every kind-1059
wrap sits at the `control_root`-derived `control_pk` and nowhere else. What
follows is strictly the device half.

---

## Manual test plan — CORD-02 §2 write gate

**What you're really testing.** Rank and key possession are now two
separate things. Before this PR, holding a moderation role was enough to
publish to the Control Plane. Now you also need the `control_root`, which
arrives inside the Grant that promotes you. Most of the plan is about the
window where those two disagree, and about not breaking communities that
predate the split.

**Setup.** Two accounts, A (owner) and B (member). Two devices is ideal;
one device with account-switching covers everything except M6. Any relay —
a loopback `geode` (`./gradlew :geode:installDist`, then `geode --port
7448`) is easiest if you want to point `nak` at the same socket and read
the wire.

Worth having open: `adb logcat -s Concord`. The refusal path logs
`Control write refused for <community>: control_root not held at epoch N`.
If that line appears for an account that *should* be staff, that's a real
bug — please capture it.

### M1 — Owner keeps everything (baseline, must not regress)

1. A: Concord → create a community.
2. A: channel list → confirm the "+" (new channel) and the metadata edit
   (pencil) are both present.
3. A: create a channel, rename it, edit the community name/icon.

*Expect:* unchanged from `main`. The creator mints the `control_root` at
genesis, so the owner is staff from the first second. Anything missing here
means the gate is over-firing.

### M2 — Fresh member has rank-less UI

1. A: mint an invite link. B: join it.
2. B: open the community's channel list.

*Expect:* no "+", no pencil. Long-press one of A's messages — no "Ban", no
"Make admin".

*Caveat:* this looks identical to `main`, because B has no roles either
way. It's a baseline for M3, not evidence on its own.

### M3 — Promotion delivers the key (the actual new path)

1. A: long-press a message from B → **Make admin**.
2. B: keep the app foregrounded through a fold cycle; background →
   foreground forces the revision tick if you're impatient.
3. B: reopen the channel list.

*Expect:* B's mod affordances appear — but only after the Grant lands *and*
its embedded key is decrypted and derive-checked, not the instant the role
becomes visible. A brief window where B shows as admin in the members list
yet still has no "+" is **correct**, not a bug. What would be a bug is that
state persisting indefinitely.

4. B: create a channel. A: confirm it appears.

*Expect:* B's edition verifies at the plane for A. If B's channel never
reaches A, the delivered key derived to the wrong address — that's the
core failure mode of this PR, so logcat please.

### M4 — Ban + Refounding rolls both keys

1. A: invite a third account C; C joins and posts.
2. A: long-press C's message → **Ban**.

*Expect:* C's messages vanish for A and B immediately. The epoch bumps; A
and B keep reading and writing across the rotation with no re-invite. B
(staff) must still create a channel *after* the rotation — that's the
staff-sized rekey blob delivering the new secret. B able to read but not
write post-rotation means the staff blob didn't arrive.

3. C: confirm C sees no new messages after the ban.

### M5 — Legacy community still works (highest-risk regression)

Needs a `main` build first, so it's the fiddliest — and the one I'd least
want skipped, since it's what could break existing users.

1. Install `main`. Create a community, add a channel, post a few messages.
2. Install this PR **over** it (don't wipe data).

*Expect:* community still opens, metadata and channels intact, history
readable, still writable. A pre-split community keeps its old single-key
plane until someone rotates it.

3. Now ban somebody in that community.

*Expect:* the rotation upgrades it to the split as a side effect, and
everything still works afterwards. If you're watching the relay,
`nak req -k 1059` shows the new epoch's wraps under a *different* author
than the old epoch's.

### M6 — Multi-device staff sync (needs the second device)

1. Sign account A in on a second device.

*Expect:* the second device can moderate too. The write key rides the
kind-13302 Community List, which is what keeps an owner's second device
from being a read-only spectator. This was a late catch in the spec review,
so it's worth checking explicitly rather than assuming.

### M7 — R8 / minified build

```
./gradlew :amethyst:installPlayBenchmark
```

Sign in, open a Concord community, moderate something.

*Expect:* identical behaviour. I don't believe this diff adds a reflection
surface — no new ViewModel factory, no new serialization entry point — so I
rate it low-risk, but it's cheap and I couldn't run it.

### M8 — Screenshots

Light + dark: owner's channel list (affordances present), a plain member's
(absent), and the members screen. Device model + Android version.

---

**One case I can't reach at all,** flagged so it isn't found in the field: a
relay that gates the plane's REQ on NIP-42 AUTH *as the stream key*. A
regular member can no longer answer that challenge — it needs the secret
they deliberately don't hold. That's the write restriction working as
designed, but it means member-side read behaviour on such a relay is
genuinely untested. If you run relays like that, try it first.

**Known gap, not a regression:** `amy` has no counterpart to the app's
staff-grant adoption, so a staffer promoted via the CLI stays read-only
until the key reaches them another way. That logic lives in
`AccountConcordActions` today; hoisting it into `commons/` is the
follow-up. Worth knowing because in CLI testing it looks exactly like the
write gate misfiring.

davotoula:
Currently in a rabbit hole adding automated on device testing to Amethyst so I don't have to run the manual test plan myself 😄 

How urgent is this PR?

alexgleason:
Without it channels on Concord can fork. It's a core protocol change we already added to Armada affecting all new communities. As long as Concord chats still seem like they're working and it doesn't crash, I'd merge it.

davotoula:
Ok, let me code review the common parts quickly

davotoula:
Nothing in the diff seem to reach non-Concord Amethyst functionality 👍 ...

Clonker had some feedback for you:

- RekeyPayload widths are a one-way interop break. decode() now accepts 72/104/136 bytes, but base rotations are minted at 104/136. An older client only accepts 72, so a member on an old build is silently stranded by any new Refounding. The stranded-recovery path presumably covers it, but it's a rollout ordering constraint, not a free change.

- Possible repeated kind-13302 republish in drainConcordStaffGrants. On success it calls sendMyPublicAndPrivateOutbox(concordChannelList.follow(entry.withControlRoot(...))), but the local guard is entry.controlRoot != null — which only becomes true once the published list round-trips back through LocalCache. adoptConcordRoot has an explicit adoptedConcordRotations dedupe set for exactly this; the grant drain has no equivalent. If a revision tick lands before the round-trip, it re-publishes.

vitorpamplona:
Found a few major bugs.. but ready to merge

alexgleason:
Thank you very much!

PR #3889, opened 2026-08-09
quartz: make a paged walk terminate — floor the cursor at 0, stop when a relay ignores it

vitorpamplona:
## Summary

`fetchAllPages` could not terminate against a relay that does not honour `until`. Found in production on `wss://purplepag.es`, which holds twelve `kind 10002` events stamped `created_at = 0` and treats `until <= 0` as *no* `until` — so the page below them comes back with its five hundred **newest** events. None of those matches the filter's own `until`, so the page delivers nothing, which read as "the boundary second is too dense to page", stepped one second lower, and asked the identical unanswerable question again. Two guards at the points where the cursor moves: the cursor floors at epoch 0, and a relay that answers above the cursor is `UNPAGEABLE` rather than something to step past.

## The bug, measured

Against the live relay, before the change: **~5.5 pages a second, 500 events fetched and discarded on each, an EOSE on every single page**, `until` marching one second further negative every time, for as long as the process ran. A cold walk pulled 1,490,010 real events in ~10.8 minutes and then never returned — so the caller's coverage was never recorded and the next boot re-walked all of it.

```
page  3204 until=1500000001 recv= 500 new= 499   ← real history, healthy
page  3205 until=0          recv= 500 new=   0   ← trap
page  3206 until=-1         recv= 500 new=   0
page  3316 until=-111       recv= 500 new=   0  EOSE 0.20s
```

Not a rate limit: ~1,000 consecutive pages drew no `NOTICE`, no `CLOSED`, no throttling, and a flat 0.18s per page.

Negative `until` is worth keeping off the wire on its own too. What the five relays actually do with `{"kinds":[0,10002],"until":-1}`:

| relay | answer |
|---|---|
| profiles.nostr1.com | `CLOSED: ERROR: bad req: error parsing until` |
| indexer.coracle.social | `NOTICE: bad req: std::get: wrong index for variant`, then never EOSE |
| user.kindpag.es | same — NOTICE, then silence |
| directory.yabu.me | same — NOTICE, then silence |
| purplepag.es | drops the bound, serves its newest page |

So: a killed subscription, a full idle timeout burned per page, or the opposite end of the relay from the one asked for.

## What changed

- **The cursor floors at zero.** `created_at` is unsigned, so nothing can exist below epoch 0: a cursor that would step under it has reached the bottom of the time axis and the walk is `DRAINED`. `until = 0` is still asked — it is a legal query and the boundary re-fetch for epoch-stamped events — only going *below* it ends the walk. Applied on the advance path too, not just the step: `pageMinTs` is an event's own `created_at`, so one relay serving a negative timestamp drives the cursor under zero, and clamping rather than stopping would not help — such an event never equals the boundary, so it dodges the dedup and returns on every page.

- **A relay that ignores the cursor is `UNPAGEABLE`.** When a page delivers nothing and every event it received was *newer* than the `until` it asked for, the relay is not paging and stepping one second lower just repeats the question. `aboveBoundary == received` is what tells this apart from a genuinely dense boundary second, whose events are *at* the boundary rather than above it. `UNPAGEABLE` is deliberate and conservative — it proves nothing about what the relay holds, so no coverage claim can be built on a page the relay never really answered.

The second guard is the structural fix. **Giving the filter a `since` does not substitute for it**: the step path decrements `until` without regard to `since`, so a cursor-ignoring relay still walks from the window floor down to 0 — up to ~1.5 billion pages. A `since` only helps when the relay honours it, and then only because the empty page arrives as a drain.

Related prior work in this area: #3489 (paging correctness + dedup) — a different problem: that one was about *which* events a walk returns, this one about whether it ends.

## Test plan

- [x] Ran `./gradlew :quartz:spotlessApply` — repo is formatted
- [x] Ran `./gradlew :quartz:jvmTest` (paging + negentropy suites)
- [x] Manually exercised the change against the live relays that found it

### Feature test plan

Four scripted tests in `NostrClientFetchAllPagesDrainTest` (JVM, real clock, existing `ScriptedClient` harness), plus a live probe.

**Fails before, passes after** — the required regression bar. Reverting only `NostrClientFetchAllPagesExt.kt` to `main` and re-running:

```
NostrClientFetchAllPagesDrainTest > aCursorSteppingUnderTheEpochDrainsInsteadOfGoingNegative FAILED
NostrClientFetchAllPagesDrainTest > anEventStampedBeforeTheEpochCannotPinTheWalk           FAILED
NostrClientFetchAllPagesDrainTest > aRelayThatIgnoresTheCursorEndsTheWalkInsteadOfSteppingForever FAILED
```

With the fix restored, all 9 tests in that class pass.

**Live relays**, through the patched client — `CursorTerminationProbe`, opt-in and asserting nothing:

```
./gradlew :quartz:jvmTest --tests "*.CursorTerminationProbe" -PprodRelayBench=1 -i
```

```
Does a paged walk TERMINATE? kinds [0, 10002], from 1600000000 down
  purplepag.es             UNPAGEABLE (266 event(s), drained=false)
                             1 page(s), 266 event(s), 2408ms, lowest until=0
  user.kindpag.es          DRAINED (0 event(s), drained=true)
  directory.yabu.me        DRAINED (0 event(s), drained=true)
  profiles.nostr1.com      DRAINED (0 event(s), drained=true)
  indexer.coracle.social   DRAINED (0 event(s), drained=true)
```

purplepag.es: **1 page, 2.4s, cursor never below 0**, where the same walk previously ran 244 pages to `until = -243` and had not ended at 45s.

### Regression test plan

Touch points, failure mode, verification:

- **The dense-boundary-second step-past** — the one path that could genuinely break, since it reaches the same `delivered == 0` branch as the new guard; if the guard misfired, any walk crossing a dense second would stop early and under-report. **This path had no test at all**, so one is added (`aBoundarySecondDenserThanAPageIsStillSteppedPast`) — verified: a second whose head is all the relay's cap will return still steps strictly past to `boundary - 1` and drains below it, with `downloaded` still showing the documented unreachable-tail loss. The guard cannot misfire by construction: a dense second's events are `== boundary`, so `aboveBoundary` stays 0 while `received` is non-zero.
- **Ordinary paged walks** — could end early or lose events. Verified: the 5 pre-existing tests in the drain suite, the 3 idle-timeout tests and the pool test all pass unchanged, and the four control relays in the live probe return identical `DRAINED (0 events)` results before and after.
- **`search`-filter walks (existing `UNPAGEABLE`)** — could be conflated with the new `UNPAGEABLE`. Verified: untouched code path; the new one is reached only from inside the `delivered == 0` branch, which a search-only walk breaks out of earlier.
- **Coverage consumers of `drained` (`SyncCoverage`)** — a new `DRAINED` at the epoch could let a caller claim history it never walked. Verified by argument rather than test: `created_at` is unsigned, so "nothing below epoch 0" is true unconditionally, and it is the same claim the old code reached one page later via an empty EOSEd page at `until = -1` on a well-behaved relay. `NostrClientNegentropySyncTest` (14 tests) passes unchanged.
- **Windowed negentropy sweeps that fall back to paging** — their windows carry `since >= 2020-01-01`, so they never reach the epoch; but their fallback is `fetchAllPages`, so before this change a cursor-ignoring peer would walk them down to 0 too. Verified: this change is what bounds them; no behaviour change for a peer that pages normally.

**Not verified, stated rather than omitted:** I did not build or install either Android flavour. This is a pure `quartz/` protocol change with no Android imports and no UI surface, which `CONTRIBUTING-WITH-AI.md` § *Build and install both flavours* allows if the reason is stated — this is that statement. Nor did I run a release-minified (`installPlayBenchmark`) pass: no reflection, serialization or ViewModel surface is touched.

## Interop suites

The one wire-visible effect is that a negative `until` is no longer emitted in a `REQ`, which none of the listed suites (MLS/Marmot, NIP-17 DM, audio rooms, MoQ-lite, QUIC) exercises — none of them pages by `created_at` cursor.

## AI assistance

Per `CONTRIBUTING-WITH-AI.md`, one gate is **not** met and I am flagging it rather than dropping it silently: § *Code review pass before opening the PR* asks for a review by a different agent or model than the one that wrote the diff, and no second model reviewed this. What it did get instead: a self-audit that found and closed the untested step-past path above, a fails-before/passes-after check on every new test, and independent empirical verification against five live relays. Happy to run a second-model pass before review if you'd like.

The change does not touch any of the § *Don't touch without an issue first* areas (signer/KeyStore, release pipeline, NIP direction) — it introduces no kind, tag or NIP interpretation, only stops emitting a value NIP-01 never defined. No existing issue or open PR covers it (searched); it was found by a downstream consumer of `fetchAllPages`, and the matching fix on that side is NosFabrica/vespa-relay#88.

## License

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/relay/client/accessories/NostrClientFetchAllPagesExt.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip01Core/relay/client/accessories/NostrClientFetchAllPagesExt.kt
@@ -134,6 +134,28 @@ data class PagedFetchResult(
  * cap, so a larger value is clamped to the same page). Stepping past at least keeps
  * the download progressing to older events instead of stalling forever.
  *
+ * **Two guards keep that step from becoming a walk that never ends**, both learned
+ * from a relay in production rather than from reasoning:
+ *
+ *  - **The cursor floors at zero.** `created_at` is an unsigned timestamp, so nothing
+ *    can exist below epoch 0: a cursor that would step under it has reached the bottom
+ *    of the time axis and the walk is [PagedFetchResult.End.DRAINED]. `until = 0`
+ *    itself is still asked — it is a legal query, and the boundary re-fetch for events
+ *    stamped at the epoch — it is only going *below* it that ends the walk. This also
+ *    keeps a negative `until` off the wire, which relays disagree violently about:
+ *    measured across five, one CLOSEs the subscription with a parse error, three
+ *    answer a `NOTICE` and then never EOSE, and one drops the bound and serves its
+ *    NEWEST events.
+ *  - **A relay that ignores the cursor is [PagedFetchResult.End.UNPAGEABLE].** If a
+ *    page delivered nothing and every event it received was NEWER than the `until` it
+ *    asked for, the relay is not paging at all, and stepping one second lower just
+ *    asks the same unanswered question again. That is exactly how the first guard's
+ *    relay behaves — it treats `until <= 0` as no `until` — and without this the walk
+ *    ran ~5.5 pages a second, 500 events fetched and discarded on each, EOSE on every
+ *    one, for as long as the process lived. UNPAGEABLE is deliberate and conservative:
+ *    it proves nothing about what the relay holds, so no coverage claim can be built
+ *    on a page the relay never really answered.
+ *
  * A `search` ([Filter.search]) filter is the exception: NIP-50 results are ranked by
  * relevance, not `created_at`, so paging one by a `until` cursor is meaningless — it
  * would silently turn a top-N search into a time-walk, and never terminate against a
@@ -259,6 +281,14 @@ suspend fun INostrClient.fetchAllPages(
         val boundary = until
         var received = 0
         var delivered = 0
+
+        /**
+         * Events that came back NEWER than the `until` this page asked for — which an
+         * honest relay never sends. Counted because it is the only way to tell a relay
+         * that ignored the cursor apart from a boundary second too dense to page: both
+         * deliver nothing, and only one of them can be fixed by stepping past.
+         */
+        var aboveBoundary = 0
         var pageMinTs = Long.MAX_VALUE
         val idsAtPageMin = HashSet<HexKey>()
 
@@ -289,6 +319,9 @@ suspend fun INostrClient.fetchAllPages(
                         // early) or an unsafely published `idsAtPageMin`.
                         try {
                             received++
+                            // Before the dedup return, so it is counted for every event
+                            // the page received, not just the ones that reach the match.
+                            if (boundary != null && event.createdAt > boundary) aboveBoundary++
                             // Drop a boundary-second event we already delivered on an
                             // earlier page (the inclusive re-fetch returns it again).
                             if (boundary != null && event.createdAt == boundary && event.id in seenAtBoundary) return
@@ -414,6 +447,35 @@ suspend fun INostrClient.fetchAllPages(
             // are resolved by stepping strictly past it. `boundary` is null only on
             // the first page, which has no dedup and so can't be all-duplicate.
             val step = boundary ?: break // first page, all-duplicate: impossible, and `end` stays UNPAGEABLE
+
+            // The relay is not honouring `until`: every event it sent was NEWER than
+            // the cursor this page asked for. Stepping past cannot help — the next
+            // page repeats the same ask one second lower and gets the same answer,
+            // forever. Measured on a live relay (purplepag.es, which treats
+            // `until <= 0` as no `until` and answers with its newest page): ~5.5
+            // pages a second, 500 events fetched and discarded on each, `until`
+            // marching one second further negative every time, an EOSE on every
+            // single page, for as long as the process ran. This is the ONE reading
+            // that ends it, and it is safely conservative — UNPAGEABLE proves
+            // nothing about what the relay holds, so no coverage claim is built on
+            // a page the relay never actually answered.
+            if (aboveBoundary == received) {
+                end = PagedFetchResult.End.UNPAGEABLE
+                break
+            }
+
+            // Below the boundary there is nothing left to ask for: `created_at` is an
+            // unsigned timestamp, so no event can exist under epoch 0 and a cursor
+            // stepping past it has reached the bottom of the time axis. Ending here
+            // rather than sending `until = -1` also keeps a value off the wire that
+            // relays disagree violently about — measured across five: one CLOSEs the
+            // subscription with a parse error, three answer a NOTICE and then never
+            // EOSE (so every page burns a whole idle timeout), one drops the bound
+            // and serves its newest events.
+            if (step <= 0L) {
+                end = PagedFetchResult.End.DRAINED
+                break
+            }
             until = step - 1
             seenAtBoundary = HashSet()
             continue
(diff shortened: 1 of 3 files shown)


vitorpamplona:
CI note: `test-geode` is red on this PR with one failure, and I do not believe it belongs to this diff. Evidence, rather than an assertion:

```
ConcordPlaneKeyDeletionTest > aPlaneKeyHolderCannotDeleteTheChannelsHistory FAILED
    java.lang.AssertionError at ConcordPlaneKeyDeletionTest.kt:164
149 tests completed, 1 failed
```

- **It cannot reach this change.** The diff touches exactly one main-source file, `NostrClientFetchAllPagesExt.kt`. That test never calls the paging API — it uses its own `query()` helper built on raw `client.subscribe` plus an EOSE channel, and mentions neither `fetchAllPages`, `PagedFetchResult` nor `drained`. `geode` is the relay server and does not call the paging accessory at all.
- **It passes on this branch.** Full `./gradlew :geode:test` locally: **149 tests, 0 failures** — the same test count CI ran, so the same set. Also 3/3 on `--tests '*ConcordPlaneKeyDeletionTest*' --rerun`.
- **It passes on `main`** at `e7d71d2` locally, so this is not a base-branch breakage I could point at either.

Best guess at the mechanism: the assertion is inside a `runBlocking` whose helpers are wall-clock bounded (`query()` uses `withTimeout(5000)`, and the test does `publishAndConfirm` immediately followed by a read), and it shares a runner with `MirrorSyncThroughputTest` (1,000,000 events) and `MirrorSyncLossTest` (50,000). A loaded runner is the obvious way that gets tight. I have not tried to prove that, and I would rather flag it than quietly re-run until green.

I could not `rerun-failed-jobs` yet — the workflow still has jobs in progress (`test-and-build-android`, `test-quartz-ios`, Windows desktop), so the API returns `403 This workflow is already running`. I will re-run the failed job once the run finishes and report what the second attempt does. If it fails again on the same test, I will stop treating it as flake and dig in properly.

---

vitorpamplona:
Update — I reproduced the `test-geode` failure and it is a timing flake in the test, unrelated to this diff. Correcting my earlier comment on one point: I said I could not reproduce it, and that was only true on an idle machine.

**Reproduced** by starving the CPU (`taskset -c 0`, i.e. one core), which is what a loaded runner looks like: **1 failure in 4 runs**, identical signature. On a normal 4-core run it was 0/8, and the full 149-test suite passed.

**The actual assertion** — the CI console only prints the exception location, so this came from the local XML report:

```
java.lang.AssertionError: the relay accepts the event itself expected:<true> but was:<false>
  at ConcordPlaneKeyDeletionTest.kt:181
```

Line 181 is the publish, not the deletion semantics the test exists to pin:

```kotlin
val deletion = planeSigner().sign(DeletionEvent.build(history, createdAt = now + 10))
assertEquals(true, client.publishAndConfirm(deletion, setOf(relayUrl)), "the relay accepts the event itself")
```

`publishAndConfirm` defaults to `timeoutInSeconds = 15` and returns `false` for **both** "the relay said no" and "the relay never answered in time" — `.any { it.value.accepted }` collapses `PublishResult.NO_RESPONSE` into the same `false`. The three seeding `publishAndConfirm` calls just above it pass in the same run, so a deterministic policy rejection of the kind-5 is ruled out; timing is the only thing that differs. The test is 2 days old (`fb7c710a`, 2026-08-08), so it has had little CI exposure.

**Suggested fix, if you want one** — not the timeout on its own, but making the failure self-diagnosing: assert through `publishAndConfirmDetailed` / `publishAndCollectResults` so the message says `no response within timeout` instead of `expected:<true> but was:<false>`. Whoever hits this next then reads the cause off the report rather than bisecting for it.

**I have not fixed it here.** It is in a different module, has nothing to do with the paging cursor, and the choice between "raise the timeout" and "assert on the reason" is yours. Happy to send it as a separate PR — say the word and I will.

For this PR specifically: `lint`, and all three `build-desktop` targets are green; `test-and-build-android` and `test-quartz-ios` are still running. I could not re-run the failed job yet — the workflow still has those two in flight, so `rerun-failed-jobs` returns `403 This workflow is already running`. I will re-run `test-geode` as soon as the run completes.

---

PR #3890, opened 2026-08-10
feat(desktop): per-visible metadata + reaction loading via shared finders in commons

nrobi144:
## Summary

Desktop currently loads user metadata (kind 0) and note interactions (reactions/zaps/reposts/replies) with a load-once, viewport-batch path that misses non-feed avatars and never unsubscribes. This PR adopts Android's composition-scoped per-visible model on Desktop: metadata and interactions load **only for on-screen users/notes** and tear down when they scroll off. To share the logic, the per-user metadata finder and the per-note event finder are moved from `amethyst/service/relayClient/reqCommand/{user,event}/` into `commons/relayClient/{user,event}/` behind a narrow `UserFinderAccount` seam, then wired into Desktop.

This is built **on top of** the recent subscription-observability work (`AccountScopedQuery`, `SubPurpose`/`ExplainedFilter`): the moved finders keep emitting the same purpose-tagged filters and preserve per-account attribution — they just source the attribution pubkey from the narrow seam (`UserFinderAccount.userFinderPubkeyHex`) instead of the full `Account`, so the finder query states no longer need to carry `Account` and could drop `AccountScopedQuery`. Android behaviour is unchanged (typealias shim files keep every existing call site compiling; `UserObservers`/`EventObservers` stay in amethyst).

Draft: the automated gates below are all green and the branch merges cleanly onto `main`, but the full manual Desktop pass is still in progress — see the test plan.

### Feature test plan

Automated (all green on this branch):
- `./gradlew :commons:jvmTest` — incl. `EventFinderFilterAssemblyTest` (batched-filter assembly + the `ICacheProvider.checkGetOrCreateUser` seam).
- `./gradlew :commons:verifyKmpPurity` — the moved finders are iOS-pure in commonMain.
- `./gradlew :amethyst:compilePlayDebugKotlin` and `:amethyst:compileFdroidDebugKotlin` — both flavours compile.
- `./gradlew :desktopApp:compileKotlin` and `:desktopApp:run` — app launches, logs in, and loads (OutboxDispatcher / WoT active).
- `./gradlew spotlessApply` — formatted.

Manual (Desktop, macOS — **in progress**, full sheet in `desktopApp/plans/2026-08-10-shared-metadata-loading-v2-manual-testing.md`):
- Scroll Home → visible authors' avatars/names populate; scrolling doesn't REQ the whole list at once.
- Open a never-viewed profile / thread / people-search / DM / notifications → avatars + names populate per on-screen user.
- Reaction/zap counts populate for visible posts.
- Active Relay Subscriptions screen still attributes `PROFILE_METADATA` / `RELAY_LISTS` / `ENGAGEMENT` / `REFERENCED_EVENTS` REQs to the account (verifies attribution survived dropping `AccountScopedQuery`).

### Regression test plan

- **Android metadata/reaction loading** — the finders moved out from under it — verified: `:amethyst:compilePlayDebugKotlin` green via the `UserFinderShims`/`EventFinderShims` typealiases; `UserObservers`/`EventObservers` unchanged. On-device manual pass pending.
- **`Account` (added `UserFinderAccount` impl)** — could break account construction — verified: compiles both flavours; getters are thin snapshot reads over existing relay-list state.
- **Other callers of the moved loader functions** (thread view, hashtag feed, global search, channel/NWC sub-assemblers) — repointed at the commons functions + threaded `LocalCache` — verified: they compile; `filterMissingEvents`/`potentialRelaysToFindEvent` behaviour preserved. Manual thread/hashtag/search "load an uncached note" pass pending (sheet D5).
- **`SingleSubNoEoseCacheEoseManager` move** — its two other callers must keep attribution — verified: `ChannelLoaderSubAssembler` uses a new amethyst `AccountScopedSingleSubNoEoseCacheEoseManager` subclass; `NWCPaymentWatcher` never attributed (behaviour preserved).
- **Account switch** — could leak cached metadata/REQs across accounts — pending manual (sheet D2).
- **iOS/KMP source-set discipline** — new commonMain code must stay pure — verified: `verifyKmpPurity` green; `EOSEAccountFast` moved to commons with a KMP lock (no `synchronized`).
- **Release-minified build** — not run; accepting the risk: the change introduces no new reflection/serialization/ViewModel-factory paths (it's relay subscription-assembly + Compose wiring). Flag for a reviewer to require `installPlayBenchmark` if desired.

## Interop suites

- [x] N/A — change can't affect wire bytes / decoded audio / MLS state / DM envelopes. The finders assemble kind-0 / kind-7 / referenced-event REQ filters; `SubPurpose`/`ExplainedFilter` tags are client-side only and never sent. No DM/MLS/audio/QUIC paths touched.

## AI assistance

Built with Claude Code. The commons extraction was reconciled against the current `main`'s `AccountScopedQuery`/`SubPurpose` refactor; per-file rationale is in `desktopApp/plans/2026-08-06-shared-metadata-loading-upstream-reconciliation-plan.md`.

## License

### DIFF
--- a/.claude/skills/relay-client/SKILL.md
+++ b/.claude/skills/relay-client/SKILL.md
@@ -94,6 +94,32 @@ class MetadataFilterAssembler(
 
 Assemblers stay pure — no state, no I/O. They're the composition seam: `FeedMetadataCoordinator` takes a list of visible notes and assembles a single metadata filter covering every referenced pubkey.
 
+## Per-visible loading — the canonical entry points (`observeUser*` / `observeNote*`)
+
+Prefer these over hand-rolled "load metadata for this list" calls. They are the shared,
+KMP way to load data **only for what's on screen** — a composable subscribes while it is in
+composition and unsubscribes ~30s after it leaves (or the app backgrounds). Both live in
+`commons/relayClient/`:
+
+- **Per user** (`relayClient/user/`): `observeUserInfo/Picture/Banner/AboutMe/Name(user)`
+  each open a composition-scoped `UserFinderFilterAssemblerSubscription(user)` **and** return
+  reactive `State`. Metadata (kind 0 + relay lists) loads for on-screen users only, coalesced
+  into one batched REQ per relay for the whole visible set.
+- **Per note** (`relayClient/event/`): `EventFinderFilterAssemblerSubscription(note)` loads a
+  note's interactions (reactions / zaps / reposts / replies) while it is composed. Android's
+  `observeNote*` display observers layer on top of the same subscription.
+
+Both read front-end-provided CompositionLocals — `LocalUserFinder` / `LocalUserFinderAccount`
+(reused by the event finder) / `LocalEventFinder` — provided once near the composition root
+(Android `AppModules`, Desktop `Main.kt` via its subscriptions coordinator). The account seam
+is the narrow `UserFinderAccount` (snapshot relay-hint getters), NOT the fat `IAccount`.
+`error()` defaults mean these must never be reached from a composition without a relay client
+(e.g. the Android `:napplet` sandbox).
+
+The load-once, viewport-batch path (`FeedMetadataCoordinator.loadMetadataForNotes` /
+`loadMetadataBatched`) is superseded for foreground loading; `MetadataPreloader` remains only
+as an optional off-screen background warmer.
+
 ## Preloaders
 
 `MetadataPreloader` is the "I need metadata for 200 pubkeys, but don't melt my CPU or the relay" path. It uses `MetadataRateLimiter` (token bucket) to throttle bulk fetches and group them into relay-friendly chunks.
--- a/.github/workflows/smoke-test-desktop.yml
+++ b/.github/workflows/smoke-test-desktop.yml
@@ -92,13 +92,29 @@ jobs:
           chmod +x scripts/relax-deb-libicu.sh
           scripts/relax-deb-libicu.sh desktopApp/build/compose/binaries/main-release/deb/*.deb
 
+      # Mirrors the same step in create-release.yml so this job exercises the
+      # exact .deb release ships. libskiko-linux-arm64.so has libEGL.so.1 in
+      # DT_NEEDED and jpackage does not scan lib/app/ for Depends, so without
+      # this the arm64 app dies at startup with
+      #   UnsatisfiedLinkError: libEGL.so.1: cannot open shared object file
+      # See scripts/add-deb-libegl-dep.sh for the full rationale.
+      - name: Add libegl1 dep to arm64 .deb
+        run: |
+          set -euo pipefail
+          chmod +x scripts/add-deb-libegl-dep.sh
+          scripts/add-deb-libegl-dep.sh desktopApp/build/compose/binaries/main-release/deb/*.deb
+
       - name: Install .deb
         run: |
+          # Installed via apt (not `dpkg -i`) so the .deb's declared Depends are
+          # actually resolved — that is what pulls in libegl1 on the arm64
+          # runner, which does not ship it preinstalled.
+          #
           # jpackage's post-install script runs xdg-desktop-menu which fails
           # on CI runners ("No writable system menu directory"). The files are
           # extracted successfully; only the menu registration fails. Allow the
-          # dpkg error, then verify the binary was actually installed.
-          sudo dpkg -i desktopApp/build/compose/binaries/main-release/deb/*.deb || true
+          # install error, then verify the binary was actually installed.
+          sudo apt-get install -y ./desktopApp/build/compose/binaries/main-release/deb/*.deb || true
           echo "Installed files:"
           dpkg -L amethyst | head -30
           # Fail if the binary wasn't actually extracted
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
@@ -32,7 +32,6 @@ import com.vitorpamplona.amethyst.commons.connectedApps.signers.NostrSignerPermi
 import com.vitorpamplona.amethyst.commons.connectedApps.signers.NostrSignerPermissionStore
 import com.vitorpamplona.amethyst.commons.defaults.Constants
 import com.vitorpamplona.amethyst.commons.defaults.DefaultIndexerRelayList
-import com.vitorpamplona.amethyst.commons.defaults.DefaultSearchRelayList
 import com.vitorpamplona.amethyst.commons.marmot.MarmotManager
 import com.vitorpamplona.amethyst.commons.model.IAccount
 import com.vitorpamplona.amethyst.commons.model.buzz.BuzzRelayDialect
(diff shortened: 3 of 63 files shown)


davotoula:
From my clonker to your clonker:

Findings

1. Behavior drift: filterMissingEvents search-relay fan-out (low severity, real)

commons/…/relayClient/event/loaders/FilterMissingEvents.kt

- key.account.searchRelayList.flow.value.forEach { … }   // base
+ key.account.searchRelays().forEach { … }               // PR

Account.searchRelays() is (trustedRelayList.flow.value + searchRelayList.flow.value.ifEmpty { DefaultSearchRelayList }).toSet() — a getter shaped for the outbox finder, not for this call site.

Tracing the two components: SearchRelayListState.flow already falls back to DefaultSearchRelayList internally (SearchRelayListState.kt:61), so the ifEmpty contributes nothing; and TrustedRelayListState.flow defaults to emptySet().

Net effect: no change for users without a NIP-51 trusted-relay list. For users who have one, every missing-event REQ in the per-note event finder now additionally fans out to their trusted relays. Benign in direction, but unintended — it's a side effect of getter reuse, not a decision.

Fix: add a narrow searchOnlyRelays() to UserFinderAccount for this call site, or say in the PR that the widening is deliberate.

2. Latent runtime trap: staticCompositionLocalOf { error(…) }

LocalUserFinder, LocalUserFinderAccount, LocalEventFinder throw when unprovided. Android never provides them, and today only Desktop-only UserSearchCard reads them — so nothing breaks now. But any future commons composable calling the no-arg observeUser* / EventFinderFilterAssemblerSubscription(note) and rendered on Android (or worse, in the :napplet process, which has no relay client at all) fails at runtime with no compile-time signal. The KDoc acknowledges this ("Reading it without a provider is a programming error") but nothing enforces it. Worth a guard — either have Android provide them, or a test that asserts no Android-reachable commons composable reads them.

3. Redundant/misleading ifEmpty in Account.searchRelays()

searchRelayList.flow.value.ifEmpty { DefaultSearchRelayList } — .flow already applies that fallback. Harmless, but the UserFinderAccount KDoc's "with the default fallback applied" is true only by accident, which is how #1 slipped through.

4. Pre-existing bug on a line this PR touches — FilterByEvent.kt:46-52

note.replyTo?.forEach { parentNote ->
    if (parentNote !is AddressableNote && note.event == null) {
        potentialRelaysToFindEvent(LocalCache, note).ifEmpty { default }.forEach { relayUrl ->
            add(relayUrl, note.idHex)
        }
    }
}

The loop binds parentNote, but all three uses in the body are the outer note — so parent notes are never actually fetched; the child's id is just re-added. Confirmed present on upstream/main, so not introduced here, but the PR edits exactly these lines. The analogous loop in FilterMissingEvents.kt shadows the name (forEach { note -> … }) and is correct, which is probably how the divergence survived. Cheap to fix while in the file.

5. Test coverage gaps

The new EventFinderFilterAssemblyTest is good but covers the wrong layer for this PR's risk:

- It tests the low-level filterMissingEvents(Map<Relay, Set<id>>) overload and the checkGetOrCreateUser seam — not the filterMissingEvents(cache, keys) overload where finding #1 lives. A test asserting which relays a stub UserFinderAccount contributes for a missing event would have caught it.
- Nothing pins the UserOutboxFinderSubAssembler union equivalence — the largest mechanical move in the PR, verified here only by reading.
- Nothing asserts AccountScopedSingleSubNoEoseCacheEoseManager still attributes ChannelFinderQueryState, which is the entire reason that subclass exists.

6. Style

EventFinderFilterAssemblyTest.kt uses an inline fully-qualified name:

override fun getOrCreateAddressableNote(key: Address): com.vitorpamplona.amethyst.commons.model.AddressableNote = error("unused")

The project's CLAUDE.md forbids inline FQNs in function bodies/signatures — should be an import. (The FQNs on the right-hand side of the typealias declarations in the shim files are fine; that's the one place the language requires them.)

Verdict

The Android-side extraction is careful and, on everything I could check, behavior-preserving — the attribution seam, the relay-tier union, and the lock swap all hold up. Only finding #1 is a genuine (if narrow) behavior change, and #4 is a free fix on a line already being touched. The bigger structural concern is #2: the error(…) CompositionLocals put a runtime-only failure mode into commonMain where Android can reach it, and the PR's own defense is a doc comment.

nrobi144:
Thanks for the careful review — genuinely useful catches. Addressed in `eaba40f2a6`:

**1. Search-relay fan-out widening (real).** You're right, `searchRelays()` was the wrong getter here — it's shaped for the outbox finder (trusted ∪ own search), and reusing it silently widened every missing-event REQ to trusted relays. Added a narrow `UserFinderAccount.searchOnlyRelays()` that returns exactly the pre-extraction `account.searchRelayList` read (own search list, no trusted union), implemented on `Account` and `DesktopIAccount`, and switched `FilterMissingEvents` to it. Behaviour is now identical to base.

**2. `error()` CompositionLocals reachable from Android.** Agreed this was the weakest spot — a doc comment isn't enforcement. Went with your "have Android provide them" option: `AppNavigation` now provides `LocalUserFinder`/`LocalUserFinderAccount`/`LocalEventFinder` at the logged-in root from `accountViewModel.dataSources()` + `.account`. Android's own UI still uses the `AccountViewModel` overloads, but any shared composable that reaches the logged-in tree via the no-arg overloads is now safe. (`:napplet` renders neither feed nor user composables, so it's unaffected; and the provide is additive — it can't crash, unlike the `error()` default.)

**3. Redundant `ifEmpty` in `Account.searchRelays()`.** Confirmed `SearchRelayListState.flow` already applies `DefaultSearchRelayList` (`normalizeSearchRelayListWithBackup`), so the `ifEmpty` was dead — removed it and its now-unused import, and reworded the `UserFinderAccount` KDoc so "with the default fallback" isn't accidentally-true.

**4. `parentNote` shadowing in `FilterByEvent`.** Nice catch — the loop bound `parentNote` but the body used `note`, so parent notes were never fetched. Fixed to use `parentNote` (`.event`, `potentialRelaysToFindEvent`, `.idHex`). Confirmed the analogous `FilterMissingEvents` loop shadows correctly, so no change needed there.

**5. Test coverage.**
- Added the test at the layer where #1 lived: `filterMissingEvents(cache, keys)` with a stub `UserFinderAccount` asserts a missing event fans out to `searchOnlyRelays()` + the follow/mine/search default, and **not** the trusted relay that only `searchRelays()` would contribute — this now fails if the wrong getter is reused.
- The `UserOutboxFinderSubAssembler` relay-tier union and the `AccountScopedSingleSubNoEoseCacheEoseManager` attribution are still verified by reading rather than a test — happy to add those if you'd like; the union is a straight substitution over the same `UserFinderAccount` getters, and the attribution subclass is a one-line `accountPubKeyOf` override, but I take the point that "the entire reason the subclass exists" deserves a pin.

**6. Inline FQN in the test.** Fixed — imported `AddressableNote`.

All gates green after the changes (commons `jvmTest` + `verifyKmpPurity`, `:amethyst compilePlayDebugKotlin`, `:desktopApp compileKotlin`, spotless). Thanks again.

davotoula:
It's a bit funny / scary how a fresh independent session reviews better than a review inside same session/context...

PR #3895, opened 2026-08-11
fix(embed): bring the keyboard back on a tap in an already-focused field

vitorpamplona:
## Summary

In an embedded tab, a text field you had already focused could get stuck with no way to raise the keyboard: dismiss the keyboard (or tab away and come back) and neither tapping the field nor anything else brought it up — you had to tap off the field and tap it again. The page's field never blurred, so no `focusin` fired, and `ime.focus` was the host's **only** keyboard-raising signal. Two signals close the gap: a payload-free `ime.wantkb` doorbell that every tap inside a focused field rings, and an `ime.resync` the host sends when a tab re-activates so it can re-take a field whose focus never moved. A tab left mid-typing comes back with its keyboard; a tab whose keyboard you deliberately dismissed comes back with the caret in place and the page unobstructed.

The page keeps its focus throughout — blurring it on tab-away would fire the page's own blur handlers (validation, autocomplete dismissal, submit-on-blur) for a switch the user never made inside the page, and lose the caret you came back to.

## Test plan

- [x] Ran `./gradlew test` (`:amethyst:testFdroidDebugUnitTest`, plus the pre-commit/pre-push suites — `BUILD SUCCESSFUL`)
- [x] Manually exercised the change — on-device QA by @vitorpamplona; see attribution note below

### Feature test plan

Automated (`tools/ime-test/shim-events.mjs`, added here — drives the shipped shim in headless Chromium and asserts the page→host envelopes):

```
ok    tap an unfocused field announces it and asks for the keyboard
ok    tap the SAME already-focused field still rings the doorbell
ok    the doorbell carries no payload
ok    a host resync is answered with the focused field state
ok    tapping off the field blurs it
ok    a resync with nothing focused answers nothing
ok    a readonly field announces itself as readonly
ok    readonly survives the resync round trip
ok    a tap inside an already-focused contenteditable rings the doorbell
EXIT=0
```

Run against `main`'s shim it fails 7 of the 9, including the exact reported bug — `no ime.wantkb — the keyboard could never come back` — so it fails before the fix and passes after.

On device (@vitorpamplona; commits 37f9c5ad and b7ef983b are the fixes that came out of it):

- Type in a field in a bottom-row embedded tab → dismiss the keyboard → tap the field again → keyboard returns (previously: nothing until you tapped away and back).
- Type in a field → switch tabs → return → keyboard is back, caret where it was.
- Dismiss the keyboard → switch tabs → return → no keyboard, caret still in the field, page unobstructed; tapping the field raises it.
- Focus a `readonly` field → no keyboard, no caret handle, toolbar offers Copy / Select all only.

### Regression test plan

- **Typing / composing in an embedded field** — could regress if a re-focus restarted the input mid-word — verified: `onPageReFocus` short-circuits while the mirror still holds the field, so no `restartInput` and no re-seed; covered by the "doorbell carries no payload" and "already-focused field" cases above, and by on-device typing.
- **Per-tap cost in the page** — the first cut of this change attached the field's editing state (and its caret geometry) to every tap, which doubled per-tap cost on a large field and put up to 40 KB on the wire per tap. Measured in Chromium, interleaved A/B against `main`'s shim, 7 rounds × 100 taps on a 40k-char textarea: `main` 3.79 ms/tap · 149 B/tap, this branch 3.87 ms/tap · 184 B/tap — CPU at parity, payload constant regardless of field size.
- **Selection UI (handles, toolbar, loupe)** — shares `SelectionUiState` — verified: a re-focus carries no geometry, so it cannot pop a handle up on its own; the caret handle still appears on tap via the existing `ime.carettap`.
- **Host-side fields (the browser's address bar)** — could wrongly arm a page-field keyboard restore — verified by construction: the restore mark is `mirroring && keyboardWanted`, both owned by the mirror, so a keyboard belonging to the address bar never arms one.
- **Tab eviction / theme flip / account switch** — the restore mark is per-tab process state that could outlive its tab — verified: cleared in `evict`, `evictAll` and `rebuildAll`.
- **Both flavours** — `:amethyst:assembleFdroidDebug` + `:amethyst:assemblePlayDebug` → `BUILD SUCCESSFUL in 5m 6s` (179 tasks). Not *installed* — see accepted risk below.
- **Release-minified build** — not run. This change touches no reflection surface (no ViewModel factories, no serialization, no expect/actual); it is plain View + Compose code and a JS asset. Flagging rather than omitting.

Accepted risk / not verified by me: I had no device in this session, so the `installPlayDebug` / `installFdroidDebug` / `installPlayBenchmark` halves of the gate are unmet on my side. The on-device passes above are @vitorpamplona's, and the two commits that came out of that QA are on the branch. One host-side behaviour has never been exercised on hardware by anyone: the `TYPE_NULL` mirror for `readonly` fields (last commit) — worth a look on a device with a hardware keyboard before merge.

## Interop suites

## AI assistance

Written by Claude Code; the on-device QA, and the two fixes it produced (keyboard-restore intent tracking, `readonly` handling), are @vitorpamplona's.

Per `CONTRIBUTING-WITH-AI.md` § *Code review pass*: an audit pass ran over the whole diff and found three defects in the AI-authored part, each fixed on the branch rather than folded back into the original commits —

1. A re-focus attached the field's caret geometry, whose measurement mirrors the field into a hidden div and forces a synchronous layout: **2× the per-tap cost** on a 40k-char textarea (3.73 → 7.20 ms), measured.
2. The "am I still this field's mirror" check read `hasFocus()` alone; a focus lingering past `clearFocus()` would have skipped the re-seed and shipped the **previous tab's text** to the page on the first keystroke. Now tracked explicitly.
3. `readonly` was enforced at the Cut/Paste call sites, which a hardware keyboard's Ctrl+V bypasses (it goes straight to `onTextContextMenuItem`) — and `readonly` blocks the *user*, not scripts, so anything reaching the mirror is applied to the page and fires an `input` event no browser would. The mirror is now `TYPE_NULL` for such a field: no InputConnection to type through, while selection and Copy still work.

The gate's *different model* requirement is **not** met — the review was by the same agent that wrote the diff, plus Android Lint. Worth a human eye on the three fixes above.

## License

---
_Generated by [Claude Code](https://claude.ai/code/session_01AC3ambee9KFcvHCS6HRqhS)_


vitorpamplona:
## On-device QA report

**Device:** Samsung Galaxy Tab A7 Lite (SM-T220), Android 14 / SDK 34, arm64.
**Branch tested:** `237ece79d5` (this branch, with `main` merged in — see "Merged main" below).

Two defects found and fixed, both landed as their own commits at the tip per
CONTRIBUTING-WITH-AI § *When on-device QA finds bugs in your own diff*. Task C is
settled with a measurement that rules out **both** candidate fixes. Task D was applied,
measured, and reverted.

---

## Task A — readonly fields

### Found: a readonly field's text was written into the *next* field

Copying from the readonly field and then tapping an editable one wrote the readonly text
into it, with an `input` event no browser would fire:

```
FOCUSOUT readonly
FOCUSIN  empty
INPUT    empty val="readonly" len=8 sel=8..8     <- the page, not the user
```

`1de9c242`'s `TYPE_NULL` closes the **inbound** half — nothing can be typed into a readonly
field's mirror. It does not close the **outbound** half. The mirror still flushes on
selection changes (a long-press select-all emits one, Chrome's collapse-to-endpoint
another) and `ime.set` carries the buffer's text. Host log at the moment of the leak:

```
FLUSH sending={"type":"ime.set","text":"readonly","selStart":0,"selEnd":8,...}
FLUSH sending={"type":"ime.set","text":"readonly","selStart":8,"selEnd":8,...}
```

Delivery is asynchronous, so the flush lands after the page has moved focus and the shim
applies it to whatever field is focused *then*.

Fixed in `be95701618`, two changes because each alone leaves a hole:

- `stateJson()` omits the `text` key entirely while the mirrored field is readonly. The
  shim already treats a missing text as selection-only
  (`var next = (msg.text != null) ? String(msg.text) : prev`), so handles and Copy keep
  working off a synced selection while nothing can be written back.
- `onPageBlur()` no longer clears `fieldReadOnly`, and suppresses its own echo. The flag
  describes the buffer the mirror still holds, which outlives the blur; clearing it up
  front meant the second flush above — scheduled by `clearFocus()` moving the caret, hence
  firing *after* the reset — was computed as if the field were editable. The blur now wraps
  `clearFocus()` in `applyingRemote` and drops the queue. **That half is not
  readonly-specific**: any field's stale buffer could land on its successor; readonly just
  made it visible.

One wrong lead, recorded for the next person: plain `readonly → editable` does **not**
leak. The Copy step is required to reproduce.

### Verified (embedded tab)

| check | result |
|---|---|
| tap readonly → focuses, **no keyboard**, **no caret handle** | pass |
| long-press → handles + toolbar showing **Copy / Select all only** | pass |
| Copy actually reaches the system clipboard | pass (IME clipboard chip shows `readonly`) |
| after the fix: Copy → tap editable → no `INPUT`, field stays empty | pass |
| editable field types normally straight after a readonly excursion | pass |
| `disabled` field raises no keyboard | pass |

### NOT verified — step 3, hardware keyboard

**No Bluetooth/USB-OTG keyboard is attached to this device, and it is not a DeX or
Chromebook.** The Ctrl+V-via-`onTextContextMenuItem` path that motivates `TYPE_NULL` is
therefore **untested on hardware**. I am not reporting it as passing. Note the outbound fix
above is independent of it and closes a hole `TYPE_NULL` alone did not.

---

## Task B — keyboard-restore matrix

6/6 on the harness, then re-run post-merge against a **real site** (brainstorm.world's
search field) rather than only my own page:

| # | case | result |
|---|---|---|
| 1 | type → dismiss → tap same field → keyboard returns | pass |
| 2 | type (kb up) → switch tab → back → keyboard back, caret preserved | pass |
| 3 | dismiss → switch tabs → back → no keyboard, caret kept → tap → returns | pass |
| 4 | address bar (host field) → switch tabs → page must not pop a keyboard | pass (`owner=host-window`) |
| 5 | mid-word composition across a tab switch | pass |
| 6 | `contenteditable` re-tap → keyboard returns | pass |

Case 5 detail: typed `hel` on the soft keyboard (a real composing region, not
`input text`), switched tabs and back, typed `lo` → **`hello`**, no duplication, no loss.

Restore latency after returning to a tab left mid-typing: **~1 s, 5/5 runs** (the resync
round trip).

Method note: keyboard state read from `dumpsys input_method` — `mInputShown` for up/down,
and the non-null `mServedView` to prove ownership (`RemoteImeView` = embed vs
`AndroidComposeView` = host chrome). That is how case 4 is proven rather than assumed.

---

## Task C — "the embed is slower": it is **input routing**, and neither candidate fix applies

`perf.html`, same URL in both hosts, matched n=14 taps:

```
FULL-SCREEN  vis=visible focus=true cores=8 | cpu=48ms layout=0.88ms timer50=50ms raf=61fps
             longtasks=5 | inputDelivery=12.9ms toPaint=39.2ms

EMBED        vis=visible focus=true cores=8 | cpu=51ms layout=0.91ms timer50=50ms raf=61fps
             longtasks=4 | inputDelivery=28.3ms toPaint=38.1ms
```

Walking the decision tree in the README:

- **`vis=hidden` — ruled out.** Both `visible`, `raf=61fps`, `timer50=50ms` exactly. No
  timer clamping, no rAF suspension. **The visibility fix must not be applied.**
- **`cpu` 2–4× — ruled out.** 51 vs 48 ms (+6%). With a nuance worth recording: the cgroup
  difference **is** exactly as predicted —

  ```
  full-screen open:  cpuset:/top-app     schedtune:/top-app
  embed tab open:    cpuset:/foreground  schedtune:/foreground
  ```

  — it simply **does not translate into CPU throughput on this device**. Adding
  `BIND_IMPORTANT` + `setRendererPriorityPolicy` would be a fix that doesn't move the
  numbers, so per the brief it should not be kept. **Not applied.**
- **`layout` — ruled out** (0.88 vs 0.91 ms).
- **`inputDelivery` — this is the gap: 12.9 → 28.3 ms (+119%).** Stable across samples
  (24.2 ms at n=6, 28.3 ms at n=14). `toPaint` is 3% *better* in the embed.

So JS throughput, layout, timers and rAF are all identical, and ~15 ms is added to every
input event before the page sees it. That matches the reported symptom — lag between
action and reaction during normal use, with no page-load involvement — exactly. As
instructed for this branch of the tree: **reporting the numbers, not guessing a fix.**

### Warm parked tabs — and a correction to my own first reading

With a second web app pinned, the first sample taken right after switching looked alarming:

```
cpu=92ms layout=1.69ms raf=4fps      <- brainstorm still LOADING in the background
```

Re-measured once it settled:

```
cpu=39ms layout=0.59ms raf=61fps     <- vs 51ms / 0.91ms / 61fps with ONE tab pinned
```

So parked warm tabs **do not** degrade the active tab in steady state; the degradation is
real but transient, lasting only while a parked neighbour is actively loading. Had I
reported the first number it would have justified the pause/resume work on false evidence.
Mirroring the napplet `MSG_PAUSE`/`MSG_RESUME` may still be worth it for battery/data on
parked tabs — but **not** on the strength of active-tab contention.

### NOT measured

- **`console.log` IPC + recomposition cost.** Needs a purpose-built page that logs in a
  loop; not built. Untested.

---

## Task D — deferred offsets: applied, measured, **reverted**

Lint confirms exactly 5 `UseOfNonLambdaOffsetOverload` in `EmbeddedTabLayer.kt` (191, 249,
314, 331, 450; the constant `.dp` park at 197 is correctly not flagged). I converted all
five to `Modifier.absoluteOffset { IntOffset(...) }` and measured the same protocol
(12 web-tab switches, `dumpsys gfxinfo` framestats):

| | Total | Janky | p50 | p90 | p95 | p99 | Slow UI thread |
|---|---|---|---|---|---|---|---|
| baseline | 46 | 100% | 200 ms | 700 ms | 2200 ms | 4550 ms | 46/46 |
| after (lambda) | 47 | 100% | 200 ms | 500 ms | 2000 ms | 4950 ms | 47/47 |
| **after, run 2 — same build** | 48 | 100% | 150 ms | 200 ms | 200 ms | 250 ms | 48/48 |

The third row is the decisive one: **two identical runs on the same build differ by 20× at
p99** (250 ms vs 4950 ms) and 2.5× at p90. Run-to-run noise dwarfs any baseline↔after
difference, so this measurement cannot detect the change, and the change showed no
improvement.

What is invariant across every run: **100% janky, `Number Slow UI thread` = every frame**,
while GPU time stays at 6–7 ms. That is the documented ~1 s `SurfaceControlViewHost`
reconfigure on the UI thread. Deferring a state read from composition to layout cannot
touch it — and as the brief predicted, the layer still reads `bounds` in composition for
`.size(...)`, `haveBounds` and the overlay geometry, so the recomposition never went away.

**Reverted.** If the five warnings are worth silencing, the honest framing is
code-cleanliness, not performance; the real lever is hoisting each surface's placement into
its own child composable so the layer body stops reading `contentBounds` at all. That is a
larger change and I did not attempt it.

Task B was re-run on the reverted build afterwards (placement modifiers put the surface and
the hidden IME view on screen): 4/4 pass on the real site.

---

## Merged `main` into the branch

The branch predated `112504205c` (bottom-nav web-app tab switching), so with two web apps
pinned, **switching between them crashed** — which blocked the warm-tab and framestats work
until I merged:

```
IllegalArgumentException: No destination with route Route.WebApp/http%3A%2F%2Flocalhost%3A8765%2Fperf.html
is on the NavController's back stack.
```

Merge is clean (`237ece79d5`); the only file both sides touched is `shim.js`, and main's
edits there are in the NIP-5D/relay region, disjoint from the IME code. After merging: 10
web-app tab switches, **0 fatals**, each tab rendering its own site.

---

## Builds and gates

```
> Task :amethyst:installFdroidDebug
Installing APK 'amethyst-fdroid-arm64-v8a-debug.apk' on 'SM-T220 - 14' for :amethyst:fdroid-debug
Installed on 1 device.
BUILD SUCCESSFUL in 13s

> Task :amethyst:installPlayDebug
Installing APK 'amethyst-play-arm64-v8a-debug.apk' on 'SM-T220 - 14' for :amethyst:play-debug
Installed on 1 device.
BUILD SUCCESSFUL in 1m
```

`spotlessApply` run before each commit; `:amethyst:testFdroidDebugUnitTest` green;
`tools/ime-test/shim-events.mjs` exits **0** on this branch (including after the merge) and
**1** against main's shim — I checked the exit codes, not just the printed report.

## Everything I could not verify

- **Hardware keyboard** (Task A step 3) — no Bluetooth/USB keyboard attached to this device.
- **`console.log` recomposition cost** — needs a purpose-built logging page; not built.
- **Perfetto** — not used; framestats only.
- **Release-minified build** — not exercised.

PR #3896, opened 2026-08-11
Declare pointer hints on Report, ChatMessage, Classifieds and ChannelCreate events

vitorpamplona:
## Summary

`PubKeyHintProvider` / `EventHintProvider` / `AddressHintProvider` are quartz's kind-agnostic answer to "what does this event point at" — they let a caller walk an event's references without knowing which tag name a given NIP picked for them (`p` vs `P` vs `member` vs `moderator`). Measured against the 248k-event corpus in `quartz/src/commonTest/resources`, 84 of 403 event classes implement one, and those cover ~95% of all pointer edges. This PR closes the four largest remaining gaps: `ReportEvent` (1984), `ChatMessageEvent` (14), `ClassifiedsEvent` (30402) and `ChannelCreateEvent` (40) — together ~14.6k undeclared edges, taking corpus coverage to ~98.5%.

`ReportEvent` carried by far the most (14,244), and they are the *negative* trust signal that any web-of-trust projection most needs. Its tag classes also predate the modern layout, so they are brought up to the structure used by e.g. NIP-88 polls: `ReportedAuthorTag` implements `PubKeyReferenceTag`, `ReportedEventTag` implements `GenericETag`, all three carry a relay hint, and each gains `parseKey` / `parseId` / `parseAddressId` / `parseAsHint` companions.

**Bug fixed along the way.** NIP-56 predates the convention that slot 2 of a pointer tag is a relay hint — it put the report type there instead — so both layouts are in the wild. The old reader passed slot 2 straight to `ReportType.parseOrNull`, which despite its name never returns null: its `else` branch maps anything unrecognized to `OTHER` (`ReportType.kt:93`). A modern `["p", "<pubkey>", "wss://relay/"]` tag therefore parsed as an `OTHER` report *and* masked the event-level default type. The new shared `ReportTagLayout` disambiguates by shape — a slot that parses as a relay URL is a hint, never a type — and falls back to the event-level default when a tag names none of its own. No report-type code is a relay URL, so the two spaces cannot collide.

**No wire-format change.** `assemble()` still emits the legacy `[name, id, type]` form unless a relay hint is explicitly supplied, because many clients still read the report type out of slot 2. No caller passes a hint yet, so emitted bytes are byte-identical to `main`. (This was a real trap: `arrayOfNotNull` pads *interior* nulls with `""` rather than dropping them, so the naive three-arg `assemble` would have started writing `["p", pk, "", "spam"]` on every report. Caught by `assembleRoundTripsBothLayouts`.)

Also renames `ReportedAuthorTag.pubkey` → `pubKey` to satisfy `PubKeyReferenceTag`, updating the four call sites (`LocalCache`, `CachePruner`, `GrapeRankScore`, `UserReportWarningState`).

### Feature test plan

New `quartz/src/commonTest/.../nip56Reports/ReportHintProviderTest.kt` — 11 tests:

- Legacy layout `["p", pubkey, "nudity"]` → type from slot 2, no hint.
- Modern layout `["p", pubkey, relay, "impersonation"]` → hint from slot 2, type from slot 3.
- Empty hint slot `["p", pubkey, "", "nudity"]` → still reads type from slot 3.
- Blank type slot → falls back to the event-level default.
- `e` and `a` tags follow the same rules; `ReportedEventTag.author` is null (NIP-56 has no author slot).
- `assemble()` round-trips both layouts and asserts the legacy 3-element output byte-for-byte.
- `ReportEvent` exposes every pointer it carries, with hints emitted only where a relay is actually present.
- Completeness invariant: `linkedPubKeys()` ≡ `reportedAuthor()` and `linkedEventIds()` ≡ `reportedPost()`, so the generic and typed views cannot drift apart.
- Malformed pointers (short hex, bare `["p"]`, unparseable address) are dropped.

```
./gradlew :quartz:jvmTest --tests "*ReportHintProviderTest*" --tests "*ModerationResolveReportEvent*"
BUILD SUCCESSFUL in 1m 1s
```

The `OTHER`-masking fix is covered by `relayHintWithoutTypeFallsBackToDefaultNotOther`, which asserts `ReportType.SPAM` (the event-level default) where the old slot-2 logic produced `OTHER`.

### Regression test plan

- **Emitted report tags (wire format)** — the highest-risk item, since `assemble()` feeds every report Amethyst publishes. Verified: `assembleRoundTripsBothLayouts` pins the legacy output to exactly `["p", <pubkey>, "spam"]`. This assertion failed on my first attempt and is what surfaced the `arrayOfNotNull` padding behaviour.
- **Legacy report parsing** — could regress through the rewritten `parse()`. Verified: `legacyLayoutReadsTypeFromSlotTwo`, plus the pre-existing `ModerationResolveReportEventTest` still passes untouched.
- **Report moderation UI paths** (`UserReportCache`, `UserReportWarningState`, `Note.kt`, `CachePruner`) — all read `.type` / `.pubKey`. Semantics unchanged apart from the intended `OTHER` fix; compile-checked after the rename.
- **`pubkey` → `pubKey` rename** — pure rename, no behaviour change, compile-verified across every consuming module.
- **NIP-17 DM envelopes** — `ChatMessageEvent` gains an interface and two read-only methods; `replyTo()`, `build()`, `reply()` and the encryption path are untouched.
- **Full quartz suite** — covers the existing NIP-17 / NIP-28 / NIP-99 tests around the four changed classes:

```
./gradlew :quartz:jvmTest --rerun-tasks
BUILD SUCCESSFUL in 6m
```

Compile verification across every consuming module:

```
./gradlew :amethyst:compilePlayDebugKotlin :commons:compileKotlinJvm \
          :cli:compileKotlin :desktopApp:compileKotlin :nappletHost:compileDebugKotlin
BUILD SUCCESSFUL
```

**Single flavour.** Per `CONTRIBUTING-WITH-AI.md` § *Build and install both flavours*, one flavour is enough for a change that is conclusively flavour-irrelevant. This is a `quartz/` protocol change plus four mechanical call-site renames: no new dependencies, no manifest or ProGuard changes, no Play-only imports, nothing reflection-touched. I compiled the Play variant.

- [x] Ran `./gradlew spotlessApply` — repo is formatted (pre-commit static analysis also passed)
- [x] Ran the relevant module's tests — full `:quartz:jvmTest` green
- [ ] Manually exercised the change

**Not run, and worth a reviewer's attention:** this was developed in a headless container with no device or emulator, so there is no on-device pass — no `installPlayDebug` / `installFdroidDebug`, no `:amethyst:installPlayBenchmark` R8 check, and no manual exercise of the report/moderation UI. The Android surface of this diff is three one-word renames, so I judged the compile check sufficient, but flag it if you'd like an on-device confirmation of the moderation screens before merge.

## Interop suites

- [x] N/A — no wire-format change (see Summary: emitted tag bytes are identical to `main`), and the NIP-17 DM envelope path is untouched. The suites were not run.

## AI assistance

Diff authored by Claude Code. The four target classes were chosen from a measurement over the checked-in 248k-event corpus rather than from inspection; the `OTHER`-masking bug was found while reading `ReportType.parseOrNull` during that work, not reported separately.

## License

https://claude.ai/code/session_015qevGhorGWtF5o5C4vhAnv

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/CachePruner.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/CachePruner.kt
@@ -395,7 +395,7 @@ class CachePruner(
 
         if (noteEvent is ReportEvent) {
             noteEvent.reportedAuthor().forEach {
-                cache.getUserIfExists(it.pubkey)?.reportsOrNull()?.let { reports ->
+                cache.getUserIfExists(it.pubKey)?.reportsOrNull()?.let { reports ->
                     reports.removeReport(note)
                     reports.removeReportNamingUser(note)
                 }
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/LocalCache.kt
@@ -1864,7 +1864,7 @@ object LocalCache : ILocalCache, ICacheProvider, Dao {
         val new = consumeRegularEvent(event, relay, wasVerified)
 
         if (new) {
-            val authorsReported = event.reportedAuthor().mapNotNull { checkGetOrCreateUser(it.pubkey) }
+            val authorsReported = event.reportedAuthor().mapNotNull { checkGetOrCreateUser(it.pubKey) }
             val eventsReported =
                 event.reportedPost().mapNotNull { checkGetOrCreateNote(it.eventId) } +
                     event.reportedAddresses().map { getOrCreateAddressableNote(it.address) }
@@ -1885,7 +1885,7 @@ object LocalCache : ILocalCache, ICacheProvider, Dao {
                 // report can `p`-tag an incidentally-mentioned third party with no type of its own,
                 // and there is no threshold here to absorb that noise the way
                 // `receivedReportsByAuthor`'s hide path does.
-                val explicitlyTyped = event.reportedAuthorsWithOwnType().mapTo(mutableSetOf()) { it.pubkey }
+                val explicitlyTyped = event.reportedAuthorsWithOwnType().mapTo(mutableSetOf()) { it.pubKey }
                 authorsReported.forEach { author ->
                     if (author.pubkeyHex in explicitlyTyped) author.reports().addReportNamingUser(note)
                 }
--- a/cli/src/main/kotlin/com/vitorpamplona/amethyst/cli/commands/graperank/GrapeRankScore.kt
+++ b/cli/src/main/kotlin/com/vitorpamplona/amethyst/cli/commands/graperank/GrapeRankScore.kt
@@ -303,7 +303,7 @@ object GrapeRankScore {
                 dropped++
                 continue
             }
-            builder.addReports(r.pubKey, r.reportedAuthor().map { it.pubkey })
+            builder.addReports(r.pubKey, r.reportedAuthor().map { it.pubKey })
         }
         if (dropped > 0) System.err.println("[graperank] dropped $dropped retracted reports (NIP-09 deletions)")
         return dropped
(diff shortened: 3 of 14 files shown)


davotoula:
Did neo4j almost make it in? Seems like a good fit for any Web of trust networks?

Fond memories of using neo4j and writing cypher 😂😂

vitorpamplona:
This was part of a neo4j store, yes. But too many schema issues yet. :(

PR #3923, opened 2026-08-14
fix: blank parent card above a top-level community post

davotoula:
## Summary

A NIP-22 comment posted to a NIP-72 community answers the community itself, so it has no parent note — but the reply-context resolver rendered an empty card where the parent belongs. The resolver excluded the community with `note.event?.kind != CommunityDefinitionEvent.KIND`, which only holds once the community definition event is cached: an uncached `AddressableNote` has a null `event`, and `null != 34550`, so the empty shell was passed to `ReplyNoteComposition`. The same test also misrouted nested replies inside a community — `replyingToAddressOrEvent()` falls back to the root `A` tag, so it returned the community shell and the real parent comment was never shown.

The fix resolves the parent by *address* kind, which an `AddressableNote` knows without its event, and renders the community itself as the context card instead of leaving the slot blank. The resolver was triplicated across `Text.kt`, `ZapPoll.kt` and `JumpToParentReplyButton`; it now lives in `commons` as `replyingDirectlyTo(note, cache)` with unit tests.

Reported against `nevent1qqs9k0cfqv0f8u0kmlr3azzsgha2l84qh4fgk27fyh9yw6a5pkqmnq` — a mostr-bridged top-level post in the "movies" community, which tags the community both as `A`/`a` and as an `E`/`e` pointing at the definition event's *id*. That id can never resolve, because addressable events are cached by address, not by id.

### What changed

| Area | Change |
|--------------------------------|-----------------------------------------------------------------------------------------------------|
| `commons/model/Note.kt`        | New `kindOrNull()` / `isKind()` — an addressable note knows its kind from the address before its event arrives |
| `commons/ui/note/ParentNote.kt`| New shared `replyingDirectlyTo(note, cache)`, replacing three copies of the same resolution          |
| `quartz/nip72ModCommunities`   | New `CommentEvent.isTopLevelCommunityPost()` and `topLevelCommunityAddress()`; `communityAddress()` now stops at the first community root address instead of parsing every `A` tag into a list |
| `quartz/nip22Comments`         | New `hasDirectKinds()` (`fastAny`-backed) so the top-level check allocates nothing on the render path |
| `amethyst/ui/note/nip22Comments` | New `DisplayCommentScope` owning the community-vs-external-id choice for both call sites, and `DisplayCommunityScope` rendering the card |
| `amethyst` call sites          | `Text.kt`, `ZapPoll.kt`, `ThreadFeedView.kt`, `NoteCompose.kt` (`JumpToParentReplyButton`, `RenderRepost`) now share the resolver |

`RenderRepost` carried the same null hole one function away from the fix and is included.

No new event kind, tag form, or NIP interpretation — this only changes how existing NIP-22/NIP-72 events are rendered.

## Test plan

- [x] Ran `./gradlew spotlessApply` — repo is formatted (`spotlessCheck` green)
- [x] Ran `./gradlew test` (`:quartz:jvmTest`, `:commons:jvmTest`, `:amethyst:testPlayDebugUnitTest` — all green)

Device: **Pixel 9a, Android 17 (API 37)**, running the R8-minified `installPlayBenchmark` build (`com.vitorpamplona.amethyst.benchmark`, `1.13.1-BENCHMARK`).

### Feature test plan

- Open the reported note in the **home/profile feed** → the previously blank slot above the post now shows the community card (avatar + `movies`), and the community definition is fetched lazily at render time — the real avatar and name load rather than falling back to the bare `d` tag.
- Tap the post → **Thread screen** → the same card renders above the master note, via `ThreadFeedView`'s own call site.
- Card is tappable → navigates to the community screen.

Screenshots (dark theme): feed row, profile Notes tab, and Thread view — attached below.
Screenshot_20260814-111946 Screenshot_20260814-113934

### Regression test plan

- **Reply-context card on ordinary NIP-10 threads** — the extracted `replyingDirectlyTo` serves every note with a parent; a mistake here would blank or mis-target every reply card — verified: unit-tested (`ParentNoteTest.nestedReplyInsideACommunityStillResolvesItsParent`, `noteInsideAChannelIsNotOfferedAsAParent`), and the refactor is behaviour-preserving by construction (each call site's pre-existing asymmetry is retained — `JumpToParentReplyButton` still applies its extra channel filter, which the shared fallback deliberately does not).
- **Nested replies inside a community** — previously resolved to the community shell instead of the real parent; this PR fixes that, so it must not regress in the other direction — verified by unit test; **not yet exercised on-device** (see gaps below).
- **Community screen feed** — now publishes its own scope via `LocalCurrentExternalScope`, so rows don't repeat a card for the community they are already inside — **not yet exercised on-device**.
- **URL / hashtag / geohash comments** — `DisplayCommentScope` consolidated the external-id branch; a mistake would drop the URL preview chip — verified: the external-id path is unchanged apart from being moved, and `LocalCurrentExternalScope` has no provider above a thread destination, so honouring it at both call sites is behaviour-identical; **not exercised on-device**.
- **R8-minified build** — the whole manual plan above ran on `installPlayBenchmark`, not a debug build, so no keep-rule gap in the new Compose code.
- **KMP source sets** — new code in `commons/commonMain` and `quartz/commonMain` must stay Android-free — verified: no `android.*` imports, and `:quartz:compileKotlinIosSimulatorArm64` builds clean.
- **Render-path allocation** — this code runs per note in scrolling feeds — verified: the new predicates use the `fast*` operators (`fastAny`, `fastFirstNotNullOfOrNull`) per the quartz raw-array convention; the cheap kind check now gates the address parse, so non-community comments allocate nothing.

### Regression tests for the bug

Both suites were mutation-checked — reverting the fix makes them fail, so they are not passing for the wrong reason:

| Suite | Reverting | Result |
|------------------------------------|---------------------------------------------|--------------|
| `commons ParentNoteTest` (5 tests) | resolver back to `event?.kind` only         | 4 of 5 fail  |
| `commons ParentNoteTest` (5 tests) | `kindOrNull()` back to `event?.kind` only   | 2 of 5 fail  |
| `quartz CommunityCommentScopeTest` (6 tests) | n/a — new protocol predicate      | pass         |

### Known gaps

Stated explicitly rather than omitted:

- **F-Droid flavour not built.** Only `:amethyst:compileFdroidDebugKotlin` was compiled (clean); `installFdroidDebug` was not run. The change adds no Google-proprietary touch points — it is Compose + `quartz`/`commons` logic only — so it should be flavour-irrelevant, but the build was not exercised end-to-end.
- **Light-theme screenshots not captured** — dark only.
- **On-device coverage is partial.** Ordinary NIP-10 reply cards were confirmed working on device after the resolver extraction, alongside the community path. Zap-poll reply cards, reposts of community posts, URL/hashtag-scoped comments, and the community screen's own feed rest on unit tests rather than a click-through.
- **`ThreadFeedView` now honours `LocalCurrentExternalScope`** where it previously did not. This is behaviour-identical today because neither provider (`UrlScreen`, `CommunityScreen`) is an ancestor of a thread destination, but it is a structural assumption about the navigation graph, not something a unit test can pin. If a thread view is ever nested under either screen, the scope card would silently disappear.

And the ### Regression tests for the bug table, which is now understated:

### Regression tests for the bug

Every suite was mutation-checked — reverting the relevant code makes them fail, so none are passing for the wrong reason:

| Suite | Reverted change | Result |
|----------------------------------------------|---------------------------------------------------|-------------|
| `commons ParentNoteTest` (7 tests)            | resolver back to `event?.kind` only               | 4 fail      |
| `commons ParentNoteTest` (7 tests)            | `kindOrNull()` back to `event?.kind` only         | 2 fail      |
| `quartz CommunityCommentScopeTest` (9 tests)  | `isTopLevelCommunityPost()` back to the kind-only heuristic | 2 fail |
| `quartz CommunityCommentScopeTest` (9 tests)  | `communityAddress()` scan order reversed          | 1 fail      |

`communityAddress()` is additionally pinned against its pre-rewrite implementation as a reference oracle across eight tag shapes, because seven call sites outside this branch depend on it being unchanged.

## Interop suites

This PR only reads existing tags and changes rendering; no event is constructed or serialized differently.

## AI assistance

Per `CONTRIBUTING-WITH-AI.md` § *Code review pass before opening the PR*, the diff was reviewed by agents other than the one that wrote it, and the tests plus the on-device plan were re-run afterwards:

| Pass | Outcome |
|--------------------------------|--------------------------------------------------------------------------------------------|
| `/simplify` (4 parallel agents: reuse, simplification, efficiency, altitude) | ~12 distinct findings; the substantive ones fixed in `b03af02` — see below |
| `/kotlin-review` (kotlin-reviewer agent) | 0 CRITICAL, 0 HIGH, 1 MEDIUM (a local shadowing the function it calls) — fixed in `1532707` |
| `./gradlew :amethyst:lintPlayDebug` | Exit 0; 0 new findings in touched files (the 4 hits there are pre-existing) |

Review findings acted on: the null-hole fix was generalized onto the model (`kindOrNull`) instead of staying a private community-specific helper, which is what let `RenderRepost` be fixed too; the duplicated scope dispatch was collapsed into one composable; three allocation sites on the feed render path were removed; an over-large `CommunityScaffold` extraction was reverted in favour of the `UrlScreen` precedent; and a test that passed for the wrong reason (the early return fired before it reached the branch it claimed to cover) was rewritten so the address-kind path is genuinely exercised.

Findings deliberately **not** acted on, with reasons:

- `CommentEvent.tagsWithoutCitations()` already decides "is this the root of a community post" from addresses, while the new `isTopLevelCommunityPost()` decides it from kind tags. Unifying them is the deeper fix, but it changes what enters `Note.replyTo` for every community comment, with blast radius into feed filters and thread assembly — that deserves its own PR with its own testing, not a quiet rider on a bug fix.
- `ReplyContext.from()` (the desktop reply-label resolver) has no community gate at all and shows "Replying to @owner" for these posts — the desktop twin of this bug. Left alone because it is a desktop behaviour change that was not verified here. **Worth a follow-up issue.**
- Outside the community screen a top-level community post now names the community twice: the `/n/<community>` chip in the header and the new context card. Left as-is; suppressing the chip for top-level community posts is a one-condition change if maintainers prefer it.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteCompose.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteCompose.kt
@@ -67,6 +67,7 @@ import com.vitorpamplona.amethyst.commons.ui.note.QuietMark
 import com.vitorpamplona.amethyst.commons.ui.note.RenderCashuMint
 import com.vitorpamplona.amethyst.commons.ui.note.RenderFedimint
 import com.vitorpamplona.amethyst.commons.ui.note.RenderMintRecommendation
+import com.vitorpamplona.amethyst.commons.ui.note.replyingDirectlyTo
 import com.vitorpamplona.amethyst.commons.ui.state.produceCachedStateAsync
 import com.vitorpamplona.amethyst.model.AddressableNote
 import com.vitorpamplona.amethyst.model.LocalCache
@@ -1995,14 +1996,7 @@ fun JumpToParentReplyButton(
 
     val parentNote =
         remember(baseNote) {
-            val noteEvent = baseNote.event as? BaseThreadedEvent ?: return@remember null
-            val direct =
-                noteEvent
-                    .replyingToAddressOrEvent()
-                    ?.let { accountViewModel.getNoteIfExists(it) }
-                    ?.takeIf { it.event !is CommunityDefinitionEvent && LocalCache.getAnyChannel(it) == null }
-            val resolved = direct ?: baseNote.replyTo?.lastOrNull { it.event !is CommunityDefinitionEvent }
-            resolved?.takeIf { LocalCache.getAnyChannel(it) == null }
+            replyingDirectlyTo(baseNote, LocalCache)?.takeIf { LocalCache.getAnyChannel(it) == null }
         } ?: return
 
     ClickableBox(
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteCompose.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/NoteCompose.kt
@@ -67,6 +67,7 @@ import com.vitorpamplona.amethyst.commons.ui.note.QuietMark
 import com.vitorpamplona.amethyst.commons.ui.note.RenderCashuMint
 import com.vitorpamplona.amethyst.commons.ui.note.RenderFedimint
 import com.vitorpamplona.amethyst.commons.ui.note.RenderMintRecommendation
+import com.vitorpamplona.amethyst.commons.ui.note.isCommunityDefinition
 import com.vitorpamplona.amethyst.commons.ui.note.replyingDirectlyTo
 import com.vitorpamplona.amethyst.commons.ui.state.produceCachedStateAsync
 import com.vitorpamplona.amethyst.model.AddressableNote
@@ -1708,7 +1709,7 @@ fun RenderRepost(
     accountViewModel: AccountViewModel,
     nav: INav,
 ) {
-    note.replyTo?.lastOrNull { it.event !is CommunityDefinitionEvent }?.let {
+    note.replyTo?.lastOrNull { !it.isCommunityDefinition() }?.let {
         NoteCompose(
             it,
             modifier = Modifier,
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/nip22Comments/DisplayCommentScope.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/ui/note/nip22Comments/DisplayCommentScope.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2025 Vitor Pamplona
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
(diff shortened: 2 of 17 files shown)


vitorpamplona:
I am confused about this one... What is it trying to fix? The screenshots show the /n/movies AND a reply quote to Movies in the note. Was the point to add the inner card? Was the outer /n/movies not enough?

davotoula:
Sorry didn't provide screenshot of before the fix. Here it is:
Screenshot_20260814-172627

`nevent1qqs9k0cfqv0f8u0kmlr3azzsgha2l84qh4fgk27fyh9yw6a5pkqmnq`

PR #3934, opened 2026-08-16
Add support for ipfs:// URIs using public gateways (dweb.link & ipfs.io)

besoeasy:
## Summary

Adds support for `ipfs://` URIs across URL detection, media classification, and media loading pipelines. Unresolved `ipfs://<CID>` links in notes, avatars, and NIP-94 file metadata are resolved on the client side to public HTTP gateways (`https://dweb.link/ipfs/` primary, `https://ipfs.io/ipfs/` fallback), ensuring Nostr events maintain immutable, decentralized content identifiers while leaving gateway routing to the client.

## Test plan

- [x] Ran `./gradlew spotlessApply` / `./gradlew spotlessCheck` — repo is formatted
- [x] Ran `./gradlew :commons:jvmTest :quartz:jvmTest` (including new `IpfsGatewayResolverTest`, `UrlParserTest`, and `MediaUrlContentExtTest` cases)

### Automated Test Output:
```text
> Task :spotlessCheck UP-TO-DATE
> Task :quartz:spotlessCheck
> Task :commons:spotlessCheck
> Task :amethyst:spotlessCheck
BUILD SUCCESSFUL in 2m 45s

> Task :quartz:jvmTest
> Task :commons:jvmTest
BUILD SUCCESSFUL in 9m 24s
35 actionable tasks: 35 executed
```

### Manual Testing Notes & Screenshots:
Tested notes and NIP-94 events referencing `ipfs://` image and video URIs. Verified that:
1. `ipfs://<CID>/<path>` is properly detected by `UrlDetector` and parsed into `withScheme` by `UrlParser`.
2. Media items bridge through `IpfsGatewayResolver` to `https://dweb.link/ipfs/` for Coil image rendering and ExoPlayer video streaming.
3. Profile avatars with `ipfs://` URIs load through `bridgeProfilePictureUrl`.

## Interop suites

*(Reason: This change is strictly a client-side URI resolution and media-loading bridge for `ipfs://` schemes. It does not alter Nostr wire protocol formats, audio room pipelines, MLS states, or DM envelope cryptography.)*

## AI assistance

## License


vitorpamplona:
I am not sure if this is right... 

ipfs.io and dweb.link are not robust enough for production systems handling hundreds of requests per second (our current state). These public gateways are run on a best-effort basis with no service level agreements (SLAs), and they enforce strict rate limits. Relying on them for high-traffic apps risks throttling, downtime, and slow response times as well as outright ban. 

Public endpoints are intended for testing, demos, and light community use, not heavy production workloads.

besoeasy:
@vitorpamplona I agree but we have to start somewhere.

I am alone working on this IPFS thing side. A little encouragement and help would be nice, since depending on blossom doesn't feel right.

Please see this, 
note1dzd4prget0dtu58kdksh5sfcleqfrqyc4satxma77lnt6cusntvq839mud

Here is proof IPFS is workable gupt.app - it doesn't use public gateway but Direct p2p connection from browser using webrtc - I use a npmjs package.

My biggest fear is anyone can upload CASM and take down my blossom server. See every country has its own laws. I work with people from India and other countries and running a blossom server. Is 110% guarantee of going to jail.

try these app :
Gupt.app
0xchat

They both have support for IPFS and work flawlessly sometimes we see a delay of 1-4 seconds but we never had a problem.

https://apps.umbrel.com/app/originless
https://originless.gupt.app/

besoeasy:
@vitorpamplona you are a developer with good history of experiment and implementation, if any anyone can introduce this. It is you. 

As I said, we might not agree on which which way is best for uploads? But we can agree that we are stuck on just one technology and most of the servers don't work. I was frustrated earlier today because I tried to upload almost 60 MB to primals blossom server. 

Do me a favour, Please give the https://gupt.app a try, I am sure that you will love it.

+ I am not asking much. I am just asking you to allow IPFS:// parsing - regarding gateways we can allow user to use its own ipfs gateway or they can simply just fall back to the public ones.

besoeasy:
@vitorpamplona Thank you for the honest feedback — you are right that depending on `dweb.link` / `ipfs.io` alone is not something I would want to build a production client on either. My earlier comments muddled the ask, so let me restate it clearly, because I think the value got lost.

## What I am actually asking for (smaller than it looks)

**1. Just accept `ipfs://` as a valid URI scheme.** The parsing/detection parts of this PR (`UrlParser`, `UrlDetector`, `MediaUrlContentExt`) are tiny and zero-risk — they only treat `ipfs://<CID>` the way we already treat `blossom:`. Nothing is fetched at parse time, no wire format changes, no new dependencies. The gateway resolution only happens when media is actually rendered.

**2. A small LRU cache with a ~3h TTL for resolved IPFS content.** This directly addresses your "hundreds of requests per second" concern. IPFS CIDs are **content-addressed and immutable** — the same CID means the exact same bytes for every user, forever. That makes IPFS the single most cache-friendly scheme in all of Nostr:

- With a per-client LRU cache (3h), each device fetches each unique CID **at most once every 3 hours**. A viral image seen by a million users is ~1 fetch per device per 3h, not per scroll, not per view.
- Amethyst already has Coil's memory + disk caches — the IPFS fetch plugs straight into them, so this is a small change, not a new subsystem. I'm happy to implement it in this PR to make the gateway concern concrete.

**3. Let users choose their gateway.** Default to a public gateway, but expose a setting so users can point at their own IPFS node (`http://localhost:8080`) or a self-hosted Originless node. Communities that care will run their own, which is exactly how this scales without leaning on public gateways.

## Why IPFS instead of (or alongside) Blossom

Blossom is a great idea, but as a *server* it has three structural problems that IPFS does not:

|  | Blossom / central HTTP server | Originless (IPFS) |
|---|---|---|
| **Setup** | Needs a domain, TLS cert, open public port | Zero domain / NAT-ready. One `docker run` line, or 1-click Umbrel deploy ([apps.umbrel.com/app/originless](https://apps.umbrel.com/app/originless)) |
| **Operator liability** | Server is a direct HTTP host serving content at `https://domain/hash` — the operator is legally on the hook for whatever is uploaded | Files are content-addressed (`ipfs://CID`) and served by the P2P swarm; the operator is a peer, not a web origin. Nothing points back at their domain |
| **Link resilience** | If the server dies or is taken down, every link breaks instantly | Once a CID is propagated to the swarm it survives even if the origin node goes offline |

That legal point is not theoretical for me. I collaborate with people in countries where hosting anything remotely questionable on a domain you control is a near-guarantee of trouble. A Blossom server is a domain + a public origin you own — it is a liability. Originless is a peer in a swarm: **no domain, no setup, no exposed origin**, and once a file is propagated it stays alive even if your node shuts down. The software already exists and is working in production — [gupt.app](https://gupt.app) and 0xchat ship IPFS support and run fine; occasionally a 1–4s delay, but never a hard failure.

## The bottom line

I'm not trying to replace Blossom overnight, and I agree the upload path can be debated. All I'm asking for in this PR are two small, safe things:

1. Accept `ipfs://` as a URI scheme so these links stop being plain text.
2. Resolve them through a configurable gateway (public by default, user-overridable) plus the 3h LRU cache so we are not hammering anything.

Please give [gupt.app](https://gupt.app) a try — I think it will change your mind about how far IPFS has come on the client side. If you'd like, I'll update the PR with the LRU cache implementation to make the gateway concern concrete.

besoeasy:
@vitorpamplona Thank you for the honest feedback — you are right that depending on `dweb.link` / `ipfs.io` alone is not something I would want to build a production client on either. My earlier comments muddled the ask, so let me restate it clearly, because I think the value got lost.

## What I am actually asking for (smaller than it looks)

**1. Just accept `ipfs://` as a valid URI scheme.** The parsing/detection parts of this PR (`UrlParser`, `UrlDetector`, `MediaUrlContentExt`) are tiny and zero-risk — they only treat `ipfs://<CID>` the way we already treat `blossom:`. Nothing is fetched at parse time, no wire format changes, no new dependencies. The gateway resolution only happens when media is actually rendered.

**2. A small LRU cache with a ~3h TTL for resolved IPFS content.** This directly addresses your "hundreds of requests per second" concern. IPFS CIDs are **content-addressed and immutable** — the same CID means the exact same bytes for every user, forever. That makes IPFS the single most cache-friendly scheme in all of Nostr:

- With a per-client LRU cache (3h), each device fetches each unique CID **at most once every 3 hours**. A viral image seen by a million users is ~1 fetch per device per 3h, not per scroll, not per view.
- Amethyst already has Coil's memory + disk caches — the IPFS fetch plugs straight into them, so this is a small change, not a new subsystem. I'm happy to implement it in this PR to make the gateway concern concrete.

**3. Let users choose their gateway.** Default to a public gateway, but expose a setting so users can point at their own IPFS node (`http://localhost:8080`) or a self-hosted Originless node. Communities that care will run their own, which is exactly how this scales without leaning on public gateways.

## Why IPFS instead of (or alongside) Blossom

Blossom is a great idea, but as a *server* it has three structural problems that IPFS does not:

|  | Blossom / central HTTP server | Originless (IPFS) |
|---|---|---|
| **Setup** | Needs a domain, TLS cert, open public port | Zero domain / NAT-ready. One `docker run` line, or 1-click Umbrel deploy ([apps.umbrel.com/app/originless](https://apps.umbrel.com/app/originless)) |
| **Operator liability** | Server is a direct HTTP host serving content at `https://domain/hash` — the operator is legally on the hook for whatever is uploaded | Files are content-addressed (`ipfs://CID`) and served by the P2P swarm; the operator is a peer, not a web origin. Nothing points back at their domain |
| **Link resilience** | If the server dies or is taken down, every link breaks instantly | Once a CID is propagated to the swarm it survives even if the origin node goes offline |

That legal point is not theoretical for me. I collaborate with people in countries where hosting anything remotely questionable on a domain you control is a near-guarantee of trouble. A Blossom server is a domain + a public origin you own — it is a liability. Originless is a peer in a swarm: **no domain, no setup, no exposed origin**, and once a file is propagated it stays alive even if your node shuts down.

**And the gateway concern goes away almost entirely:** [gupt.app](https://gupt.app) doesn't hammer public gateways at all — it uses [`@helia/verified-fetch`](https://github.com/ipfs/helia-verified-fetch), which fetches the file **directly P2P over WebRTC** (libp2p) and only falls back to a public gateway if the direct peer-to-peer fetch fails. The gateway is the last resort, not the primary path. 0xchat ships IPFS support the same way and both run fine in production; occasionally a 1–4s delay on a cold CID, but never a hard failure.

### Proof it works — I posted this today

[note1dzd4prget0dtu58kdksh5sfcleqfrqyc4satxma77lnt6cusntvq839mud](https://iris.to/note1dzd4prget0dtu58kdksh5sfcleqfrqyc4satxma77lnt6cusntvq839mud) — a NIP-68 kind 20 post with **7 full-resolution photos (4640×6960)** uploaded to IPFS in one go, every one carrying `ipfs://` CIDs, gateway URLs, SHA-256 hashes, dimensions and blurhashes in the `imeta` tags:

```
["imeta","url https://dweb.link/ipfs/QmRioC4HUs3XhQ8kymC735A8jWiiZcQ8QTzBMuy1NWy1Ef",
 "fallback https://ipfs.io/ipfs/QmRioC4HUs3XhQ8kymC735A8jWiiZcQ8QTzBMuy1NWy1Ef",
 "fallback ipfs://QmRioC4HUs3XhQ8kymC735A8jWiiZcQ8QTzBMuy1NWy1Ef",
 "m image/jpeg", "x 8c5d96ef...", "dim 4640x6960", ...]
```

All 7 uploads succeeded on the first try. I genuinely tried to do the equivalent on a Blossom server earlier today (uploading ~60 MB to Primal's server) and it **failed repeatedly — sometimes files uploaded, sometimes they failed** with no way to tell which until the event was already signed. That flaky, opaque upload path is exactly why I don't want to build on Blossom.

## Showcase: real apps already running on IPFS today

Neither of these is a demo — both are live production apps shipping to real users, with IPFS/Originless as their storage backend, and neither has had storage problems:

- **[0xchat](https://0xchat.com/)** — a Nostr client with IPFS media support built in. Media uploads and loads through IPFS in normal daily use without issues.
- **[gupt.app](https://gupt.app)** — an encrypted messaging app that stores all media on IPFS. It fetches directly P2P via `@helia/verified-fetch` over WebRTC, falling back to a gateway only when a direct peer fetch fails.

Both have been running this way for a while now. The only imperfection is an occasional 1–4s delay on a cold CID the very first time — after that it's instant. No upload failures, no dead links, no "server down" screens. That's the reliability I want Amethyst users to have, and the infrastructure to deliver it is already out there, today, not something that needs to be built.

## A respectful ask for your help

I'll be honest about the trade-off: IPFS tooling can be irritating at times — it's young, and parts of the developer experience are rough. But its ecosystem is **hundreds of times larger than Nostr's**, and its infrastructure is far more robust than the handful of Blossom servers we currently depend on. And importantly, `ipfs://` is **already specified in the NIPs** — this PR isn't inventing anything, it's simply letting Amethyst read URIs that are already part of the spec.

What I'm trying to do, with your help, is bring that `ipfs://` handling into Amethyst — so we can stop cramming three fallback URLs into every `imeta` tag, and start leaning on an ecosystem that has already worked through the growing pains Nostr is still going through. You've been the person who introduced so many of Amethyst's best experiments; if anyone can bring this in, it's you.

## The bottom line

I'm not trying to replace Blossom overnight, and I agree the upload path can be debated. All I'm asking for in this PR are two small, safe things:

1. Accept `ipfs://` as a URI scheme so these links stop being plain text.
2. Resolve them through a configurable gateway (public by default, user-overridable) plus the 3h LRU cache so we are not hammering anything.

Please give [gupt.app](https://gupt.app) a try — I think it will change your mind about how far IPFS has come on the client side. If you'd like, I'll update the PR with the LRU cache implementation to make the gateway concern concrete.

vitorpamplona:
I am good with the support for IPFS, the only problem is fixing in those servers that were never made for production apps. If all of our users start hammering the same servers, this might not end up well. We run into the same issue with Unified Push notifications where people think most of the servers are free, but they only serve ~300 notifications a month for all the user's apps. After that, everything is blocked, the user doesn't receive anything and thinks we are the problem. 

We still have about 30,000 installs running at least once a week, which is small, but already big enough to be annoying to a cheap media server. Maybe we should have user-chosen IPFS servers just like we have with blossom so that each user can deploy their own IPFS proxy.

besoeasy:
Well I knew this was a problem that is why I made : 

https://github.com/besoeasy/Originless

```bash
docker run -d \
  --name originless \
  --restart unless-stopped \
  -p 3232:3232 \
  -e STORAGE_MAX=100GB \
  -e NOSTR_NPUBS=npub1...,npub1... \
  -v originless-data:/data \
  -v originless-archive:/archive \
  ghcr.io/besoeasy/originless:latest
```

It allows user to upload, file management, garbage collection, and even fetches what content you post on your NOSTR and pins it locally on your pc. 

The key part: since IPFS redirects requests to the node with the highest availability of files, public gateways will simply forward the request to your local node instead of serving the file themselves. That way you do not even have to expose it — your files stay reachable through any public gateway while living on your own machine.

1 click deploy via https://apps.umbrel.com/app/originless

vitorpamplona:
Right the redirect from the big nodes to your local machine is still quite heavy right? Like if 4000 people are looking at the same image at the same time because it was just posted by... say... Odell.. then every Amethyst user will ping the same server at the same time, which will all redirect to your server at the same time, which will have to manage 4000 uploads to each person's phone at the same time. 

Besides killing the central server with too many redirects, isn't that very slow?

besoeasy:
> Right the redirect from the big nodes to your local machine is still quite heavy right? Like if 4000 people are looking at the same image at the same time because it was just posted by... say... Odell.. then every Amethyst user will ping the same server at the same time, which will all redirect to your server at the same time, which will have to manage 4000 uploads to each person's phone at the same time.
> 
> Besides killing the central server with too many redirects, isn't that very slow?

Quite the opposite. The more people request the same file, the better it gets replicated across the network.

Each gateway uses an LRU cache, so the most frequently requested files are cached and served from many locations. Similar to torrents, files are split into small pieces that can be fetched from across the network. When multiple Amethyst users request the same file, those pieces end up being cached on multiple nodes as well, which improves availability and reliability.

In practice, popular content is the easiest and fastest content to serve because it already exists in many caches throughout the network.

The harder case is content that's rarely accessed. A file that's only fetched once a day is more likely to require routing and a cold start before it can be served.

vitorpamplona:
> The more people request the same file, the better it gets replicated across the network.

The app doesn't request from these other locations; it's only two. It focuses all of our traffic on only two servers. Those 2 servers need to manually forward the request every single time to these other nodes. It's a light task, but it's still a massive load if everybody decides to use those two servers. That's why they also block and limit requests to them directly from production apps.

besoeasy:
Here is what I want from you (vitor):

**Support `ipfs://` with a user-chosen gateway, falling back to public ones.**

Concretely:

1. Accept `ipfs://<CID>` as a valid URI scheme in the parsing/detection layers (`UrlDetector`, `UrlParser`, `MediaUrlContentExt`) — tiny and zero-risk.
2. Ask the user which IPFS gateway they want to use (their own node/proxy by preference), defaulting to public gateways (`dweb.link`, then `ipfs.io`) only as a fallback — same pattern as the Blossom server setting we already have.

This keeps traffic off the public gateways for users who self-host, while still making `ipfs://` links work out of the box for everyone else.

**If you want to test this setup**, my repo is here: https://github.com/besoeasy/Originless

It's a one-line docker deploy that gives you upload, file management, garbage collection, Nostr-based pinning — and it now ships its own gateway enabled by default (can be disabled). Running example: https://originless.gupt.app, in use by https://gupt.app

Happy to rework this PR to add the gateway setting + fallback chain if you're good with the direction.

vitorpamplona:
> Happy to rework this PR to add the gateway setting + fallback chain if you're good with the direction.

Please do, in a similar way we do with Blossom. You should also pick an event kind to be the standard way each user picks their central/preferred/paid IPFS resolver, as we do with Blossom servers.

vitorpamplona:
In fact, you might not need any changes to Nostr clients if you make a Blossom-compatible upload interface to your IPFS project. Then every Nostr app would already use IPFS with your own gateway as http://.

besoeasy:
Re: picking an event kind for resolver preferences — I'd like to push back on that one, because I think it quietly reintroduces the problem we're trying to avoid:

**1. Publishing gateways in events = hardcoding gateway URLs again.**
The whole value of `ipfs://CID` is that the identifier is location-independent — that's the core of the IPFS spec. As soon as we standardize stuffing `https://some-gateway/ipfs/...` strings into replaceable events (and eventually into client UX around them), we've rebuilt link rot with extra steps: gateways die, domains expire, and now that string lives in signed event history.

**2. It's mostly bloat.**
A gateway preference is only meaningful to *its owner* — my gateway is typically behind NAT/Tor, rate-limited to me, or a paid endpoint keyed to my account. Replicating that list across relays so other apps can read a URL they can't (or shouldn't) use adds bytes to the event graph without buying anyone interoperability.

**3. The simpler design covers both cases:**

- Events contain **only** `ipfs://CID`
- Client resolves through a **single local setting**: user-set gateway, falling back to a short default list shipped with the app
- No new event kind needed for v1

**4. "Production grade" is solved by software, not by which public gateway we pick.**
Users who need real throughput run their own resolver — this already exists off-the-shelf:

- [ipfs/rainbow](https://github.com/ipfs/rainbow) — the official standalone gateway implementation, with caching
- [besoeasy/Originless](https://github.com/besoeasy/Originless) — same idea plus upload/pinning management, one docker command

Both give a product-grade HTTP gateway on localhost/home server/Umbrel; the user points the setting at it. People who prefer renting point it at any commercial gateway. The app stays out of the hosting business entirely — exactly like relay configuration, just one URL.

**5. The part that keeps getting missed: IPFS is hosting *without domains*.**
I feel like the Blossom framing has us stuck comparing two HTTP server products, but that's not what IPFS is. Blossom fundamentally requires an HTTP origin — a domain, DNS, a TLS certificate, an open port. That's precisely the infrastructure burden most people can't or don't want to carry, and why most never self-host anything. IPFS doesn't need any of it: nodes speak libp2p directly, content is addressed by CID, transfers are peer-to-peer and hole-punch through NAT — no domain registration, no cert renewals, no port forwarding, no exposed origin to attack. Gateways are just an optional HTTP bridge bolted on for browsers; they're not the storage layer. A user's node can live behind a home router with zero public footprint and still serve and replicate content across the swarm. That's not a workaround — that's the actual design goal of IPFS, and it's something no domain-based protocol can offer.

If that direction works for you, I'll rework this PR to: keep the scheme detection (`UrlDetector`/`UrlParser`/`MediaUrlContentExt`) as-is, and reduce `IpfsGatewayResolver` to read the user setting with the small default chain — dropping the hardcoded dweb.link/ipfs.io primaries.

besoeasy:
@vitorpamplona Let me zoom out, because I think we're debating gateway capacity when the decision actually in front of you is much smaller — and the stakes much bigger.

**What every Nostr client does today:** media links in events are HTTP URLs pointed at someone's server. When nostr.build dies, throttles, or bans — every note that ever referenced it loses its images, permanently, because the URL *is* signed history. We've normalized a design flaw: **Nostr has no way to reference content independently of where it happens to be hosted right now.**

**That's the hole `ipfs://CID` fills.** Not "another upload protocol" — the first scheme that lets an event say *what* without saying *where*. Location stops being part of the record and becomes a per-user runtime choice. Everything underneath — self-hosted nodes, public gateways, commercial CDNs — becomes swappable infrastructure behind a stable identifier.

**And this is exactly the lesson Nostr already taught us with relays.** We never baked relay URLs into events. Events carry content; relay selection lives in client settings; anyone can run one; nodes come and go and the network survives. Media is the last major piece still violating that rule. `ipfs://` support just applies the same architecture to blobs.

**On your capacity concern — I'm not dismissing it, I'm saying it isn't yours to solve:**
- Yes, phones will fetch through HTTP bridges today. *Which* bridge is a user setting, exactly like relays.
- Defaults exist only so day-one UX works; they're not an endorsement. Users who care swap in Rainbow on a home box, Originless, or a paid provider.
- Popular CIDs get replicated and cached across every participating node and gateway automatically — demand for viral content *grows* the serving infrastructure around it, rather than us having to pre-allocate it.

**Cost/benefit from your chair:** worst case, Amethyst gains a URI scheme almost nobody uses — a few hundred lines, inert. Best case, **Amethyst is the first major client whose users' media outlives every individual host.** The rework stays deliberately boring: accept the scheme, resolve through one local setting, short default fallback chain. No new event kind, no protocol claim, nothing hardcoded anywhere.

The bigger picture isn't "add IPFS support." It's: **stop letting event history depend on other people's uptime.**

besoeasy:
@vitorpamplona Some context about where I'm coming from, because I don't think I've made it clear in this thread.

I'm a computer programmer by profession — but I haven't touched a mobile app in about ten years. So treat everything I say here as coming from a *user* of Amethyst, not someone telling you how to build Android apps. Your side of the house is yours.

**Why I care about this at all:** I am sick and tired of censorship. What I personally want is simple: host every picture on my Nostr profile in true, original quality — and trust no host except myself. That's the entire motivation. Not token counts, not specs for their own sake.

**Now the part you really need to hear, because your country probably doesn't make it obvious:**

Here in India, all it takes is someone uploading an *edited* picture of Hindu or Christian gods and filing an FIR at a police station. Courts and ISPs move fast after that — and what gets banned is **the domain**. Not one post. Not one user. The whole domain.

So picture `blossom.primal.net` blocked nationwide overnight. In that moment, **1.4 billion people lose access to every piece of Nostr media that lives behind that one URL** — including notes written by people who have nothing to do with India, because their events pointed at a hostname that a magistrate in Delhi decided is contraband. One FIR. Everyone's media. Gone for a fifth of humanity.

And India isn't special. Turkey does it, Pakistan does it, Russia does it, courts in "liberal" democracies do it with DNS seizures. Just because you live somewhere that works by the book doesn't mean the whole world does — and even the book gets rewritten. **Any design where media resolves through one mandated hostname carries a built-in kill switch, and you don't get to choose who holds it or when they pull it.**

That is why "just use Blossom" doesn't settle this for me. It's not about which server product is nicer. It's that:

1. **There must never be one solution.** Multiple independent paths to the same content — different gateways, different operators, different countries — is the only real defense. A CID gives me that; a hardcoded `https://host/media/...` never will.
2. **The ecosystem is converging into a monoculture**, and honestly it makes me sick watching the entire Nostr world line up behind whatever fiatjaf and hzrd149 publish this quarter. They're talented, but Nostr exists precisely because we refused to let a small set of people own the rails. Media rails shouldn't be owned either — not by them, not by public gateways, not by anyone.

All I'm asking Amethyst for is neutrality at the identifier level: accept `ipfs://CID` in events, resolve through a gateway the user controls, ship a few defaults so nobody notices anything. If my government kills a gateway tomorrow, I change one setting — my pictures, and everyone else's, are still there.

That's the bigger picture. It's not a feature request; it's why some of us can't afford your worst case.

vitorpamplona:
So, did you make the setting? 

Also, you should research how Blossom works. While it is true that the full http URL goes directly into the post, Nostr clients update that URL with the new servers that the user is using when that one URL fails automatically. They just pick a new server from each user's Blossom Server list event kind and try there instead. Blossom also has a blossom: url that works exactly as IPFS addresses. 

That's why I said you can just follow the same pattern. Instead of using the same IPFS resolvers all the time, like in this PR (which just defeats the purpose of IPFS as a decentralized host provider), each user can set up their resolver and let everybody else use their resolver to find each user's media.

besoeasy:
@vitorpamplona Fair correction on the client-side fallback behavior — I take the point that Blossom clients don't just die on a dead URL when they can retry against the author's server list.

But here's what I see in practice: **there are only 2-3 main Blossom servers carrying almost everything.** And yes, I understand Blossom content is addressed by SHA256 — but here's the part that matters:

**that content has to be duplicated *manually* to every new server, today.**

A blob exists on exactly the server(s) where someone explicitly uploaded or mirrored it (BUD-04 `/mirror` is still a deliberate, per-server action). If server #1 vanishes and I never bothered mirroring my library to servers #2 and #3 — which nobody does until the fire starts — then the hash in my old notes points at bytes that no longer exist anywhere reachable. The SHA256 makes the reference verifiable; it doesn't make the content *available*. Availability in Blossom is a function of how diligent each user was about manual duplication.

IPFS flips exactly that one property:

- **Zero duplication needed for reachability.** Any gateway/resolver can pull a CID from the swarm on demand, serve it, and cache it. My content becomes fetchable through your resolver without me ever uploading to you.
- **Popularity replicates itself.** The more people view a viral image, the more nodes cache it — redundancy grows *with* demand instead of depending on users pre-mirroring to a list of servers they picked years ago.
- **New gateways join with full historical access instantly** — compare that to a new Blossom server, which starts empty and only gains content one manual upload/mirror at a time.

So the difference between us isn't "hash-addressed vs not" — both are. It's **"replication by diligence" vs "replication by physics."** That's why I keep pushing on `ipfs://` specifically: the identifier is only half the value; the swarm providing the bytes is the other half, and it's the half Blossom structurally lacks.

Happy to be proven wrong on the ecosystem numbers — but look at which servers actually serve Nostr media today and tell me that's not 2-3 eggs in one basket.

*(Full disclosure: this comment is my best effort drafted with AI assistance — I gave the arguments, it polished the words. If anything reads off, feel free to tweak it; the points stand either way.)*

besoeasy:
Tell my AI agent to do a read only implementation as of now !

we can do uploading part later

vitorpamplona:
> that content has to be duplicated manually to every new server, today.

It's not manual; most new servers mirror the contents from all the others they can find already. 

> we can do uploading part later

What do you mean? Full solution or nothing. Otherwise, what's the point of this if nobody can use?

besoeasy:
what i meant was, i will do it step by step, to have your inputs while i do it.

Part 1. Content Resolve 
Part 2. Allow ipfs:// & https:// support

besoeasy:
@vitorpamplona here is small testing that you can do 

see this nsec17pfeksxqg4v83n7cg6naydtyd753knzhl3dsx7tejm5u6gx6vsuqjt24ly account, try your best to replicate the images across blossom.

i uploaded 8 pictures. only 2 went thru via blossom, while older direct http upload implementation was more reliable.

image
image
image

can you imagine failing with one 12 MB image ?  that what blossoms been doing. 


meanwhile my raspberry pi 3 - uploaded to IPFS without problem and is relaying it  https://bafybeihvp5mu2auaynij6uyzbzzvax2od5ebvoyw2jm3uj7exruowcl7lq.ipfs.inbrowser.link/

image

vitorpamplona:
I am not sure what you are trying to test. 

Are you paying for those servers or are you running them like you run your ipfs node? 

Otherwise, usually they have lots of caps for free users... Kinda similar to how ipfs nodes delete things at random.

besoeasy:
@vitorpamplona take a look boss

Screenshot_20260827-062556_Amy Debug

https://njump.to/nevent1qqs2zv3f3m6wpslrwqqh8nadhq7xlefr3hcyrcdjhuce8el9wrjxu4gpr4mhxue69uhkummnw3ez6vp39eukz6mfdphkumn99e3k7mf0qgsrdw72sy8z46hcjwgtlc645f5d5vwsxzjf7lz8ynuar2nyara0elqrqsqqqqqpq2je7k

https://github.com/user-attachments/assets/0cee57c3-5876-4155-bca9-ad119e23b8a8


https://github.com/user-attachments/assets/c9f3fe57-506a-485e-beb9-c3550aec68e0

commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/UrlParser.kt
```diff
@@ -109,6 +110,8 @@ class UrlParser {
                             collectGroupLinks(content, url.originalUrl, groupLinks)
                         } else if (url.originalUrl.startsWithAny(blossomScheme)) {
                             blossom.add(url.originalUrl)
+                        } else if (url.originalUrl.startsWithAny(ipfsScheme)) {
```
vitorpamplona, line 113: startsWithAny runs twice in this construction

besoeasy:
@vitorpamplona it works better then I expect, everything smooth, i tested local and clearnet nodes.

10/10

amethyst/src/main/java/com/vitorpamplona/amethyst/model/nipB7Blossom/BlossomServerListState.kt
```diff
@@ -82,6 +82,17 @@ class BlossomServerListState(
 
     fun mergeServerList(blossom: List<String>?): List<ServerName> = blossom?.map { ServerName(host(it), it, ServerType.Blossom) }?.ifEmpty { DEFAULT_MEDIA_SERVERS } ?: DEFAULT_MEDIA_SERVERS
 
+    /**
+     * Extracts Originless (or compatible IPFS-native) server URLs from the
+     * configured server list. These double as IPFS path-gateways because they
+     * expose `/ipfs/{cid}`.  Used by the IPFS resolver to try the user's own
+     * node before falling back to public gateways.
+     */
+    fun originlessGateways(): List<String> =
```
vitorpamplona, line 91: Are originless blossom compatible? If yes, why do you need to make the distinction? If not, why is this inside the Blossom Server List?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/hls/HlsBlobUploaderFactory.kt
```diff
@@ -61,7 +61,7 @@ object HlsBlobUploaderFactory {
         context: Context,
     ): HlsBlobUploader =
         when (server.type) {
-            ServerType.Blossom -> {
+            ServerType.Blossom, ServerType.Originless -> {
```
vitorpamplona, line 64: HLS doesn't work with IPFS, right? You can't change the contents of the file fast enough via IPFS for streaming to work.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
```diff
@@ -471,7 +471,7 @@ class UploadOrchestrator {
             return when (server.type) {
                 ServerType.NIP95 -> uploadNIP95(finalUri, compressed.contentType, null, null, context)
                 ServerType.NIP96 -> uploadNIP96(finalUri, compressed.contentType, compressed.size, alt, contentWarningReason, server.baseUrl, null, null, account, forcedSigner, context)
-                ServerType.Blossom -> uploadBlossom(finalUri, compressed.contentType, compressed.size, alt, contentWarningReason, server.baseUrl, null, null, account, forcedSigner, context)
+                ServerType.Blossom, ServerType.Originless -> uploadBlossom(finalUri, compressed.contentType, compressed.size, alt, contentWarningReason, server.baseUrl, null, null, account, forcedSigner, context)
```
vitorpamplona, line 474: Why do you need to separate server types if both use the blossom API?  Wouldn't you just add the Originless as a blossom uploader/downloader?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/pdf/PdfFetcher.kt
```diff
@@ -53,22 +64,33 @@ object PdfFetcher {
...
             try {
-                val request =
-                    Request
-                        .Builder()
-                        .url(url)
-                        .get()
-                        .build()
+                var lastFailure: IOException? = null
+                var downloaded = false
+                for (candidate in candidates) {
```
vitorpamplona, line 69: Why do you have multiple gateway resolvers? I thought the point of IPFS is that you only needed one and that one communicates with ever other node.

vitorpamplona (commented):
I am a little confused about this reuse of blossom but separating urls in the app from the blossom servers? Why not just keeping only one list of "Blossom" servers and your server just replies as if it was a Blossom request? From my understanding, it would be a simple URL redirect.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/components/pdf/PdfFetcher.kt
besoeasy: Multiple is for all the servers user may have added. its user defined list not a default from our side

amethyst/src/main/java/com/vitorpamplona/amethyst/model/nipB7Blossom/BlossomServerListState.kt
besoeasy: No originless is not blossom compatible 

Because one uses SHA256 while ipfs have its own CID & CIDv2 

Nevertheless the different https:// and IPFS://

and i wanted originless to be opt in - we do not want to scare people away with complex setups. 

this is what we followed for 0xchat.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/hls/HlsBlobUploaderFactory.kt
besoeasy: Not quite. HLS and IPFS can work together, but the key is that you generally don't mutate one IPFS file as the stream progresses.

[1. tomeshnet/ipfs-live-streaming ](https://github.com/tomeshnet/ipfs-live-streaming) — probably the best example

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
besoeasy: the reason 

Blossom - sha256 based uses http protocol

Originless - CiDv1 & CIDv2 based uses ipfs://
(http is just a compatibility layer) 

if content is indetified thru different mechanisms, they shouldn't be same

besoeasy:
@vitorpamplona re: why Originless shouldn't just be folded into the Blossom server list — the core issue is that content is identified through fundamentally different mechanisms:

**Blossom** — content-addressed by **SHA-256 hash**. The URL is `https://server/sha256hex`. The reference is tied to a specific server's HTTP origin. If that server goes down, clients fall back to other servers in the user's Blossom list, but the content must have been explicitly uploaded/mirrored there first.

**Originless (IPFS)** — content-addressed by **CID (v1/v2)**. The reference is `ipfs://CID`. Any node in the IPFS swarm can resolve a CID on demand — no prior upload to that node is needed. The HTTP gateway URL is just a compatibility layer for clients that can't speak libp2p directly; the real identifier is the CID, not the domain.

These are two different content-addressing schemes with different resolution semantics. Blossom says "fetch this exact bytes from this server (or fall back to the user's other servers)." IPFS says "fetch this exact bytes from *anywhere in the swarm* that has it." The resolution path, fallback logic, and failure modes are structurally different — putting Originless into the Blossom list would conflate two distinct protocols and force IPFS resolution through a Blossom-shaped code path that doesn't match how IPFS actually works.

That said, I hear your point about keeping the code simple. The goal here is:
1. `ipfs://CID` is parsed and recognized (not treated as plain text)
2. Resolution goes through the user's configured IPFS gateways (their own node first, then defaults)
3. The two systems coexist — Blossom for SHA-256 HTTP uploads, IPFS/CID for content-addressed immutable storage

vitorpamplona:
Of course you can merge.. You almost did it in the PR; you are basically reusing the entire upload stack for Blossom. The you can reply the url with an IPFS handle and use that in the text. But as you know that will only work in clients that implement IPFS... you will need to change about 100 clients today to display your image. Or you can write the url directly in the user's own IPFS originless gateway. That will make it open in all clients.. they don't need to implement IPFS to see the image.  

Blossom is just an API (upload/download/mirror) specification. You can follow the same API and implement IPFS storage in the backend. It's literally the same thing: one API path to upload and a gateway to download. The inner details on how the picture is saved is not speced out on Blossom.

besoeasy:
@vitorpamplona valid points, but a clean implementation is better.

i will open a clean PR. since this one went into discussions too much. please check the code once and merge.

vitorpamplona:
Don't you want to keep going, the code is not ready yet. If you say that the API is not compatible with Blossom, then you cannot use the Blossom server state on the Account, which saves the url on the Blossom server list event. Otherwise other apps will expect a full Blossom API and not work well with our user's list.

besoeasy:
@vitorpamplona fixed here https://github.com/vitorpamplona/amethyst/pull/3997

PR #3937, opened 2026-08-17
fix: honour "always log in" for relays the account does not use itself

vitorpamplona:
Split out of #3936 so it can be reviewed on its own — it is unrelated to the memory work there.

## The bug

`RelayAuthResolver` gated the `ALWAYS` policy behind `isFirstParty`:

```kotlin
RelayAuthPolicy.ALWAYS -> if (inputs.isFirstParty) ALLOW else fallThrough(inputs)   // -> ASK
```

So **"Always log in"** only auto-authenticated relays the account had its *own* reason to be on. Any relay reached only through somebody else's traffic — a followed author's outbox, another logged-in account — fell through to a prompt.

With the outbox model dialling 250+ relays and no stored per-relay decisions yet, that is **one prompt per third-party relay** on a fresh install. It is invisible on an established install only because those per-relay decisions have already accumulated.

## Why it is a bug and not a privacy feature

Narrowing to "only the relays I actually use" is exactly what the **"decide per relay"** option (`CUSTOM` + `RelayAuthCustomToggles`) exists to express. Applying the same narrowing to `ALWAYS` left the user no way to say *"just authenticate everywhere"* — and contradicted both the enum's own KDoc:

```kotlin
/** Authenticate with every relay that requests it. */
ALWAYS,
```

and the setting's user-facing description, `relay_auth_policy_always_desc` → **"Every relay that asks."**

## The change

`ALWAYS` is now unconditional. The first-party gate stays on `CUSTOM`, where it belongs.

Precedence is otherwise untouched: blocked relays (kind 10006) and explicit per-relay overrides are still resolved *before* the policy, so nothing the user has already decided is reopened.

## Tests

The old behaviour was pinned by a test that documented it as intentional:

```kotlin
fun nonFirstPartyNeverAutoAuthsUnderAlwaysPolicy() {
    // "Always log in" is a statement about the relays this account uses. ...
    assertEquals(RelayAuthVerdict.ASK, resolve(inputs(policy = ALWAYS, isFirstParty = false)))
}
```

Replaced with `alwaysPolicyAuthsEvenWhenNotFirstParty` (ALLOW either way), plus a new `customPolicyStillRequiresFirstParty` so the gate stays covered where it does apply.

`RelayAuthPermissionLedger.kt:122` is the only caller of `RelayAuthResolver.resolve`, so there is no duplicate policy logic to keep in sync.

## Verification

- `:commons:jvmTest` green; amethyst `AuthDecisionResolverTest` + `RelayAuthVenueCoverageTest` green.
- On device: a fresh install that previously produced a stream of "«relay» asks who you are" dialogs now launches straight into the feed with none.

### DIFF
--- a/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/relayauth/RelayAuthResolver.kt
+++ b/commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/relayauth/RelayAuthResolver.kt
@@ -92,11 +92,14 @@ data class RelayAuthInputs(
  *      else fall through
  * 4. Fall-through → [RelayAuthVerdict.ASK] when the purpose is known, otherwise DENY.
  *
- * Both automatic grants in step 3 additionally require [RelayAuthInputs.isFirstParty]: an account
- * never reveals its identity *without being asked* on a relay it has no reason of its own to be on.
- * That is what keeps a bystander account off a relay only another account uses. It deliberately does
- * not suppress the question — a non-first-party challenge we can explain falls through to ASK, so
- * "decide per relay" means the user decides rather than a silent denial they never see.
+ * The [RelayAuthPolicy.CUSTOM] grant additionally requires [RelayAuthInputs.isFirstParty]: under
+ * "decide per relay" an account never reveals its identity *without being asked* on a relay it has no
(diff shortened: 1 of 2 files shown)


davotoula:
Nice one!

PR #3939, opened 2026-08-17
Mute public chats

davotoula:
Screenshot_20260810-164625

(PS: there is an issue with the amethyst users chat logo, the updated logo kind 41 is on nostr.wine and nowhere else)

## Mute a public chat
Long-press a public chat in Messages, or use the new button in the expanded channel header, and the chat stops driving notifications. The row keeps its place with a notifications-off glyph, and **the messages stay fully visible** — this silences alerts, it does not hide content.

## What muting suppresses

| Signal | Where it is gated |
| --- | --- |
| Row unread bubble | `rowHasUnread` (`RowUnread.kt`) |
| Bottom-nav envelope badge | `AccountViewModel.messagesHasNewItems` |
| Push notifications | `EventNotificationConsumer.notifyChannelMessage` |
| Notifications-tab entries | `NotificationFeedFilter.acceptableEvent` |
| Reactions/zaps/reposts on your messages there | both of the above |

Muting silences the channel completely, including `@`-mentions, matching Signal/Telegram "mute group".

All five decisions route through one pure predicate, `isMutedPublicChatMessage`, so they cannot drift apart.

## Storage

Two representations, and the split is deliberate.

**Local device state** (`AccountSettings.mutedPublicChats`) is the read path. Every consumer reads only this, because it needs no Nostr signer.

**NIP-78** (kind 30078, `d="AmethystSettings"`, beside the existing `pinnedRooms`) carries it across devices. No new event kind.

The split exists because `AppSpecificState` decrypts the settings blob asynchronously, and a push-driven cold start can reach `notifyChannelMessage` before that completes. For a NIP-55 external-signer account that decrypt is an IPC round-trip to Amber that may never complete in the background. A mute that loses that race posts exactly the notification the feature exists to kill, so NIP-78 never gates a read — it only feeds the local mirror.

The synced field is `List<String>? = null`, and the nullable default is load-bearing. `updateFrom` overwrites local state whenever remote differs, and `AppSpecificState` replays the cached backup on every app start. With a non-null `emptyList()` default, an older client that rewrote the blob without the key would clear the user's mutes on *every single launch*. `null` keeps "key absent" distinguishable from an explicit "unmute everything", and `mergeMutedPublicChats` acts on that distinction.

## Also fixed: "Mute thread" destroyed public chats

Verified by execution: `Event.threadRootIdOrSelf()` returns the **channel id** for every NIP-28 message, top-level and reply alike, because the channel *is* the NIP-10 root. So "Mute thread" on a public-chat message wrote the whole channel into the NIP-51 mute list, and `Account.isAcceptable` — which `ChannelFeedFilter` calls — then rejected every message in it.

The channel rendered empty while the Messages row kept listing it, and because the list is published to relays it survived restart, cache clear and reinstall. With every message filtered there was no message left to long-press, so Settings → Security Filters → Muted threads was the only escape, where the entry showed as an unnamed "unknown" thread.

A tester hit this during device testing of this branch and lost posts. The action is now hidden when the note is a `ChannelMessageEvent`; **unmute is still offered** when a legacy mute is already set, so anyone already caught can escape.

## Testing

**1279 unit tests pass.** 11 are new, covering the pure predicate and the whole serialization contract, including both arms of the null-vs-empty distinction.

The `mergeMutedPublicChats` guard is mutation-verified rather than assumed: changing `?: local` to `?: emptySet()` fails exactly one test, `absentKeyLeavesTheLocalMuteSetAlone`, and only that one.

`AccountSettings` cannot be constructed in a JVM unit test — `syncedSettings` is a property initializer that builds a default `AccountSyncedSettingsInternal`, whose language preferences call `Resources.getSystem()`, and there is no Robolectric in the project. Decisions are therefore extracted into pure functions, the same shape as the existing `unreadPrivateChatRoute`.

Verified on a Pixel 9a (Android 17):

- muted → no push, no unread dot, no envelope badge
- muted → messages still fully visible in the chat
- unmute → both dots return for messages that arrived while muted
- "Mute thread" absent on public chats in this build, still present in release

That third item only works because muting never advances the last-read marker.

## Known limitations

**Unmuting does not backfill the Notifications tab.** `NotificationFeedFilter` is an `AdditiveFeedFilter`: it only applies its filter to newly-arriving items and never re-scans the cache when an input changes. Entries suppressed while muted stay suppressed until the tab is refreshed. This is consistent with the rest of the app — unmuting a NIP-51 thread behaves the same — but it is a real gap and is commented at the call site.

**Push notifications suppressed while muted are gone permanently.** Android cannot replay a notification that was never posted, so "unmuting restores everything" holds for both dots and for chat content, not for system notifications.

**Older Amethyst builds can erase the synced copy.** `ignoreUnknownKeys = true` means they do not break, but any settings change republishes the blob without the field. The local copy survives, so only cross-device propagation is lost. `pinnedRooms` already carries the identical exposure.

**Android only.** Desktop never reads kind 30078, so it does not honour mutes. The stored form is a list, so the other Messages row types extend to it without a migration.

## Scope

20 files, +669/-58, `amethyst` module only. No new dependencies, no new event kind. `MaterialSymbols.NotificationsOff` was already in the bundled subset font, so it needs no regeneration — and it is deliberately not `VolumeOff`, which already labels the different "Mute thread" action.

`ChatroomRowUnread.kt` is renamed to `RowUnread.kt` for ktlint's single-class filename rule, after the helper gained a `RowUnread` type carrying an initial value alongside its flow. Without that seed, routing the row through the shared helper turned a `StateFlow` read into a cold-flow collection, and every row painted dotless for one frame before correcting itself.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
@@ -214,6 +214,7 @@ private object PrefKeys {
     const val HAS_DONATED_IN_VERSION = "has_donated_in_version"
     const val DISMISSED_POLL_NOTE_IDS = "dismissed_poll_note_ids"
     const val DISMISSED_CHANNEL_INVITES = "dismissed_channel_invites"
+    const val MUTED_PUBLIC_CHATS = "muted_public_chats"
     const val VIEWED_POLL_RESULT_NOTE_IDS = "viewed_poll_result_note_ids"
     const val PENDING_ATTESTATIONS = "pending_attestations"
 
@@ -650,6 +651,7 @@ object LocalPreferences {
                     putStringSet(PrefKeys.HAS_DONATED_IN_VERSION, settings.hasDonatedInVersion.value)
                     putStringSet(PrefKeys.DISMISSED_POLL_NOTE_IDS, settings.dismissedPollNoteIds.value)
                     putStringSet(PrefKeys.DISMISSED_CHANNEL_INVITES, settings.dismissedChannelInvites.value)
+                    putStringSet(PrefKeys.MUTED_PUBLIC_CHATS, settings.mutedPublicChats.value)
                     putString(
                         PrefKeys.VIEWED_POLL_RESULT_NOTE_IDS,
                         JsonMapper.toJson(settings.viewedPollResultNoteIds.value),
@@ -789,6 +791,7 @@ object LocalPreferences {
                     val hasDonatedInVersion = getStringSet(PrefKeys.HAS_DONATED_IN_VERSION, null) ?: setOf()
                     val dismissedPollNoteIds = getStringSet(PrefKeys.DISMISSED_POLL_NOTE_IDS, null) ?: setOf()
                     val dismissedChannelInvites = getStringSet(PrefKeys.DISMISSED_CHANNEL_INVITES, null) ?: setOf()
+                    val mutedPublicChats = getStringSet(PrefKeys.MUTED_PUBLIC_CHATS, null) ?: setOf()
                     val viewedPollResultNoteIdsStr = getString(PrefKeys.VIEWED_POLL_RESULT_NOTE_IDS, null)
                     val localRelayServers = getStringSet(PrefKeys.LOCAL_RELAY_SERVERS, null) ?: setOf()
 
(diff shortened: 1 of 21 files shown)


vitorpamplona:
Might need to do the same for Buzz / NIP-29 relay groups / Ephemeral, etc. 

But I think right now you only receive the notification if you were tagged in these groups. We could also notify every message from chats that people join... anyway.. much to improve

PR #3944, opened 2026-08-17
perf: scan NIP-19 entities without ICU to stop the cold-start native-heap OOM

vitorpamplona:
## The bug

On a 2.9 GB SM-T220 the **release** build grew to ~1.9 GB RSS during a cold start and was killed by lmkd about 32 s in:

```
lmkd: Reclaim 'com.vitorpamplona.amethyst.benchmark' … to free 1871628kB rss, 375492kB swap
      reason: min2x watermark is breached even after kill
ActivityManager: Process … has died: fg TOP
```

The growth is entirely in the **native** heap. The Java heap plateaus at its 512 MB `largeHeap` ceiling and GCs back down, while native runs away:

| t | TOTAL | Dalvik | **Native** |
|---|---|---|---|
| 9.2s | 309M | 93M | 45M |
| 13.8s | 1137M | 438M | 497M |
| 28.2s | 1884M | 334M | **1372M** |

That's why earlier Java-heap histograms (~350 MB) never explained it. It is also **not images** — Coil's memory cache held 2–11 MB of its 102 MB budget the whole time.

## Cause

`forEachNip19Match` called `Regex.matchAt` once per candidate position. On Android `java.util.regex` is ICU-backed, and `Regex.matchAt` builds a fresh `Matcher` whose `region()` → `reset()` → `MatcherNative.setInput()` **copies the entire input into native memory**. So scanning one note allocated a full native UTF-16 copy of its content **per candidate `n`**, and note content reaches **767 KB** in the tail. The Java `Matcher` object is tiny, so Java-heap-driven GC had no reason to reclaim them promptly and native memory grew unbounded.

heapprofd's top malloc stack is exactly this path:

```
LocalCache.justConsume → updateHintIndexes → TextNoteEvent.eventHints
  → Nip19Parser.parseAll → Regex.matchAt → Matcher.region → Matcher.reset
    → MatcherNative.setInput → libicu_jni MatcherState::updateInput → malloc
```

The file's own KDoc had already recorded the symptom from an earlier pass — *"2,541 of 4,573 live Matchers were running this regex"* — but that pass optimized **speed** (the 9–23× anchoring win) and left the native retention in place.

Confirmed causally before writing the fix, via a kill switch on the same build: with the scan disabled the native heap plateaued at ~257 MB; with it enabled it reached 882 MB by t=18 s and kept climbing.

## The fix

The grammar is prefix + bech32 payload + trailing non-space, so it's matched directly with char compares — no ICU.

Two details worth reviewing:

- **Case folding is ASCII-only on purpose.** `RegexOption.IGNORE_CASE` maps to `Pattern.CASE_INSENSITIVE`, which is ASCII-only unless `UNICODE_CASE` is also set. Kotlin's `ignoreCase = true` is Unicode-aware and would have *accepted inputs the regex rejected* (U+212A KELVIN SIGN folds to `k`).
- **Reusing a single Matcher would not have fixed this** — `region()` re-copies the input on every call. Removing ICU from this path was the only option.

Only the ingest hot path changes. `uriToRoute` (98 call sites), `tryParseAndClean` and `hasAny` still use the regexes: they run on short user input, not per ingested event.

## Measured on device (release codegen, 3 runs)

Native heap RSS:

| | t~9s | t~14s | t~22s | t~28s | t~43s |
|---|---|---|---|---|---|
| before | 45M | 497M | 626M | 1372M | *killed at 31.9s* |
| after | 48M | 127M | 170M | 175M | 172M |

Native now plateaus at **~170 MB**, total RSS falls back to ~500 MB instead of climbing to 1.88 GB, and the process survives past 45 s with **zero lmkd kills**. The feed renders with `@mention` NIP-19 entities resolving correctly.

Note this reproduces **only in release builds** — a debug build only reaches ~337 MB because the interpreter throttles ingest below the rate needed to outrun memory. Measure with `./gradlew :amethyst:installPlayBenchmark`.

## Testing

- New `Nip19ScannerRegexEquivalenceTest` runs **both original regexes** over a shared corpus and requires identical entity lists — targeting the exact-58 boundary, the excluded bech32 chars (`1`, `b`, `i`, `o`), ASCII-only case folding, and what `[\S]*` swallows (Java's `\s` is the six ASCII whitespace chars, so U+00A0/U+2003 belong to the trailing group).
- Mutation-tested: both `58 → 57` and admitting `b` into the alphabet fail the test.
  - Worth calling out — the first version of that test had **no teeth for charset bugs**. Admitting an extra char only produces *invalid* payloads that `parseComponents` rejects on both sides, so the lists matched trivially. The cases that catch it are a **valid** variable-length entity butted straight against an excluded char (`NEVENT + "b"` etc.), where a permissive charset swallows the char, fails the bech32 decode and silently drops a real entity.
- Pre-existing guards still pass: `Nip19ScanTest` (23), `NIP19ParserTest` (37), and commons' `nip19MatchesReferenceScan`.
- Full quartz suite: **4288 tests, 0 failures**.

## Scope

This fixes the dominant native-heap consumer. It does not address the ~190-relay unthrottled connect fan-out that drives ingest volume in the first place — that remains the structural item.

### DIFF
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
@@ -24,6 +24,7 @@ import androidx.compose.runtime.Immutable
 import com.vitorpamplona.quartz.nip01Core.core.HexKey
 import com.vitorpamplona.quartz.nip01Core.core.hexToByteArray
 import com.vitorpamplona.quartz.nip01Core.core.toHexKey
+import com.vitorpamplona.quartz.nip19Bech32.bech32.Bech32
 import com.vitorpamplona.quartz.nip19Bech32.bech32.bechToBytes
 import com.vitorpamplona.quartz.nip19Bech32.entities.Entity
 import com.vitorpamplona.quartz.nip19Bech32.entities.NAddress
@@ -163,17 +164,6 @@ object Nip19Parser {
 
     private val EVENT_VARIABLE_PREFIXES = arrayOf("nevent1", "naddr1", "note1", "nrelay1", "nembed1")
 
-    /** bech32: digits except `1`, letters except `b`, `i`, `o`. ASCII-only, both cases. */
-    private val BECH32_CHARS =
-        BooleanArray(128).apply {
-            for (c in "qpzry9x8gf2tvdw0s3jn54khce6mua7l") {
-                this[c.code] = true
-                if (c in 'a'..'z') this[c.code - 32] = true
-            }
-        }
-
-    private fun isBech32(c: Char): Boolean = c.code < 128 && BECH32_CHARS[c.code]
-
     /**
      * `\S` in the trailing group. Java's `\s` is the six ASCII whitespace chars unless
      * `UNICODE_CHARACTER_CLASS` is set, so U+00A0 and friends count as NON-space here.
@@ -280,7 +270,7 @@ object Nip19Parser {
                         if (!matchesPrefixAt(content, i, prefix)) continue
                         val dataStart = i + prefix.length
                         var dataEnd = dataStart
-                        while (dataEnd < len && isBech32(content[dataEnd])) dataEnd++
+                        while (dataEnd < len && Bech32.isDataChar(content[dataEnd])) dataEnd++
                         // `+` needs at least one payload char.
                         if (dataEnd > dataStart) {
                             end = endOfTrailing(content, dataEnd)
@@ -310,7 +300,7 @@ object Nip19Parser {
         to: Int,
     ): Boolean {
         for (k in from until to) {
-            if (!isBech32(content[k])) return false
+            if (!Bech32.isDataChar(content[k])) return false
         }
         return true
     }
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
@@ -246,7 +246,25 @@ object Nip19Parser {
     ) {
         var i = 0
         val len = content.length
+        // Jump between 'n'/'N' with indexOf — an intrinsified, vectorised char search — instead of
+        // testing every character. Both cases are tracked separately so each is only re-searched
+        // once consumed, amortising to about one indexOf per candidate. Measured ~2x faster on
+        // content with no entity at all, which is 94% of real notes (2588-note production sample),
+        // and ~26% faster on the 767KB mention-heavy tail. Small mention-dense content (a ~700B
+        // note with 2 mentions) is a few microseconds slower; that trade is deliberate.
+        var nextLower = content.indexOf('n')
+        var nextUpper = content.indexOf('N')
         while (i < len) {
+            if (nextLower in 0..<i) nextLower = content.indexOf('n', i)
+            if (nextUpper in 0..<i) nextUpper = content.indexOf('N', i)
+            i =
+                when {
+                    nextLower < 0 && nextUpper < 0 -> return
+                    nextLower < 0 -> nextUpper
+                    nextUpper < 0 -> nextLower
+                    else -> if (nextLower < nextUpper) nextLower else nextUpper
+                }
+            if (i >= len) return
             if (isCandidateAt(content, i)) {
                 var end = -1
 
--- a/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
+++ b/quartz/src/commonMain/kotlin/com/vitorpamplona/quartz/nip19Bech32/Nip19Parser.kt
@@ -133,6 +133,78 @@ object Nip19Parser {
 
     fun hasAny(content: String): Boolean = nip19regex.matches(content)
 
+    // ---------------------------------------------------------------------------------------
+    // ICU-free scanner for the ingest hot path.
+    //
+    // `java.util.regex` on Android is backed by ICU, and `Matcher.region()` -> `reset()` ->
+    // `MatcherNative.setInput()` copies the ENTIRE input into native memory on every call.
+    // `Regex.matchAt` builds a fresh Matcher per call, and [forEachNip19Match] calls it once per
+    // candidate position — so scanning one note allocated a full native UTF-16 copy of its content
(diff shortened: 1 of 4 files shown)


vitorpamplona:
Follow-up commit `c5a6aa2090`: the alphabet check now lives on `Bech32` as `isDataChar`, instead of the scanner carrying its own copy of `ALPHABET`.

`Bech32.map` couldn't be reused for it — it's an `Array<Byte>` (boxed `java.lang.Byte[]`), and this runs per character over whole note contents, so it would unbox on every char. `isDataChar` gets its own primitive `BooleanArray` built from the same `ALPHABET` constants in the existing `init` block, so there's still one source of truth.

Held at parity with the private lookup it replaced, verified in the bytecode:

| form | call site per character |
|---|---|
| plain member | `invokevirtual Bech32.isDataChar` — measured ~1–2% slower, consistently signed across 8/10 benchmark cases |
| `inline` only | `invokevirtual Bech32.getDATA_CHARS()` — the property getter replaced the call |
| `inline` + `@JvmField` | `getstatic; iload; baload` — identical to what the private array produced |

RegexContentBenchmark, medians of 3 runs (ns, lower better):

| case | bytes | before | after |
|---|---|---|---|
| 0 mentions | 4104 | 4473 | 4518 |
| 0 mentions | 68096 | 73398 | 74167 |
| 0 mentions | 767144 | 836400 | 835772 |
| m=120 | 767072 | 1029977 | 1030072 |
| **total** | | **2102097** | **2104899** (+0.1%) |

The 767 KB cases — the tail that caused the OOM — overlap run to run (`[855512, 828137, 836400]` vs `[833456, 839112, 835772]`). The two sub-microsecond cases swing ±80% between repeats of the *same* build, so they carry no signal.

On-device native heap unchanged: plateaus at ~169–182 MB over two more cold starts, zero lmkd kills.

Adds `Bech32DataCharTest` (sweeps the whole BMP against `ALPHABET`). Mutating the now-shared `ALPHABET` fails both it and the NIP-19 equivalence test. Full quartz suite 4292/0.

One note: `PoolEventOutboxScaleTest` flaked once during a parallel full-suite run but passes 3/3 in isolation both with and without these changes — it's a load-sensitive scaling test in relay-outbox code this PR doesn't touch.

vitorpamplona:
**Validated against production data.**

Until now the equivalence corpus was hand-written, which only proves the scanner agrees with the regex on cases someone thought of. So I pulled real note content off public relays with `amy fetch` (kinds 1, 1111, 30023 from relay.damus.io / nos.lol / relay.primal.net) and ran the same differential over it:

```
2588 real contents, max 183973 chars, 117 with entities, 7742 entities parsed, 0 mismatches
```

Both `parseAll` and `parseAllEvents` produced entity lists identical to the original regexes on every one. 7742 entities parsed, so the comparison isn't vacuous.

Reproduce:
```bash
./gradlew :cli:installDist
HOME=/tmp/amyhome cli/build/install/amy/bin/amy fetch --kind 1 --limit 3000 --paginate \
  --relay wss://relay.damus.io,wss://nos.lol,wss://relay.primal.net --timeout 25 --json
# then diff Nip19Parser.parseAll against nip19regex.findAll over each .content
```
(the isolated $HOME matters — otherwise amy picks up an account and stalls on the bunker signer during relay AUTH.)

On-device, release build, feed scrolled: mentions resolve (`@Rabble`, `@Liz Sweigart`, `@Alice Chan`), hashtags and reposts render, no crashes, and no parse-related errors in logcat. The only errors present are push-server 503s and Firebase transport warnings, both unrelated.

The differential harness itself is not in this PR — it needs an external corpus file and a machine-specific path.

vitorpamplona:
**Correction to the PR description, and a further speedup — plus two more offenders found.**

### 1. Correction

The description says the old code allocated a native copy "per candidate \`n\` position per note". That overstates it: `isCandidateAt` requires a full NIP-19 prefix match, so a Matcher was minted **once per NIP-19 prefix occurrence**, not once per `n`. The amplification is still content-length × entity-count — over 2588 real notes that is 7,871 Matchers copying **1,752 MB**, worst single note 62.8 MB — but the mechanism sentence was imprecise.

### 2. `86a9d3b780` — indexOf candidate jumping

`findHashtags` already used the better shape and this scanner didn't: jump between candidates with intrinsified `indexOf` rather than testing every char. Medians of **6** benchmark runs per arm:

| case | bytes | char-loop | indexOf | delta | ranges overlap |
|---|---|---|---|---|---|
| 0 mentions | 608 | 1048.0 | 422.5 | −59.7% | no |
| 0 mentions | 68096 | 75678.0 | 38936.5 | −48.5% | no |
| 0 mentions | 767144 | 867870.5 | 434141.0 | −50.0% | no |
| m=120 | 767072 | 1097200.5 | 808300.5 | −26.3% | no |
| m=5 | 4050 | 12317.0 | 10835.5 | −12.0% | yes |
| m=2 | 698 | 4453.5 | 7211.0 | **+61.9%** | yes |
| **TOTAL** | | 2218165.5 | 1431202.0 | **−35.5%** | |

Every no-match case is ~2× faster with non-overlapping ranges — and that's the path that runs on nearly every event: of 2588 real notes, only 160 contain a NIP-19 prefix, so **94% never leave the scan loop**. The one regression is a ~700 B note with 2 mentions, +2.7 µs absolute with overlapping ranges; accepted deliberately.

Correctness re-verified after the change: 0 mismatches over the same 2588-note production corpus (7742 entities), plus the synthetic corpus and `Nip19ScanTest`.

### 3. Not in this PR: `findHashtags` has the same bug, ~5.5× worse

`ContentHashTags.findHashtags` and `IndexedTags.forEachIndexTag` use the identical `Regex.matchAt`-per-candidate pattern, so each candidate still mints an ICU Matcher that copies the whole content natively. Measured over the same 2588 real notes:

| scanner | Matchers | native bytes copied | worst single note |
|---|---|---|---|
| nip19 (fixed here) | 7,871 | 1,752 MB | 62.8 MB |
| **findHashtags** | **43,626** | **9,639 MB** | **279.6 MB** |
| findIndexTags | 0 | 0 MB | — (legacy `#[0]` form, absent from real data) |

Both are reached per event via `BaseNoteEvent.citedUsers()` / `findCitations()`. This very likely accounts for the rest of the "2,541 of 4,573 live Matchers" the earlier heap dump attributed to the NIP-19 regex — the other ~2,000 were something else.

Happy to fix `findHashtags` the same way in a follow-up. `findIndexTags` shares the pattern but never fires on real data, so it's latent rather than urgent.

(For the record: NIP-29 / `nip29RelayGroups` contains no `Regex` at all.)

PR #3987, opened 2026-08-26
Fix nwc silent refusals

davotoula:
## Summary

Some NIP-47 wallet refusals reached the phone and showed the user nothing at all — no toast, no dialog, no error state. The action simply looked like it had never happened. This was reported from two field trips of a NIP-47 wallet service against Amethyst's built-in NWC wallet, a week apart, same phone, same pairing: `QUOTA_EXCEEDED` on `pay_invoice` and `RESTRICTED` on `list_transactions` were both silent, while a `PAYMENT_FAILED`/`NO_ROUTE` on the same `pay_invoice` method rendered its full text.

Three independent defects produce that symptom. All three are fixed here, and both original cases are verified on-device against a real wallet.

`QUOTA_EXCEEDED` is worth singling out: of the refusals a NIP-47 wallet returns, it is the only one the user can act on themselves — wait for the budget window, or ask for a raise. `RESTRICTED` and `UNAUTHORIZED` need the wallet operator whatever the user is told. So the failure mode was the inverse of what you would choose: the refusals only an operator can fix were shown, and the one the user could act on arrived as silence.

## What was wrong

**1. The zap path had no user-visible timeout.** `NwcSignerState`'s 60-second safety net only dropped the relay subscription. It never retired the pending tracker entry and never told anyone, so a response lost in transit was permanent silence. This turns out to be the actual cause of the reported `QUOTA_EXCEEDED` case — see Testing below.

**2. `WalletTransactionsScreen` never read `walletViewModel.error`.** The ViewModel set it correctly on both the refusal and timeout paths; the view branched on `isLoading`/`isEmpty` only and rendered "No transactions yet" straight over the top of a perfectly good error string.

**3. Consumers matched on `PayInvoiceErrorResponse`, which the deserializer only produces when `result_type == "pay_invoice"`.** NIP-47 does not require a wallet to echo `result_type` on an error, and an error for any other method takes the generic `NwcErrorResponse` branch — so those refusals were dropped without a word. The DVM screen went further and thanked the user for a payment that had just been refused.

## What changed

**Timeouts now reach the user.** The give-up path retires the tracker entry and invokes a new `onTimeout` callback, wired through `AccountZapActions` → `AccountViewModel` to every interactive surface: `ZapPaymentHandler`, `InvoicePaymentDispatcher`, `SendPaymentScreen`, `DvmContentDiscoveryScreen`, and both `V4VPaymentHandler` rails.

`NwcPaymentTracker.cleanup` now returns whether *it* removed the entry. Both the timeout and the response path bottom out in `ConcurrentHashMap.remove`, so exactly one of them ever gets the value — a timeout racing a real response stays quiet instead of overwriting the wallet's own answer. Four new tests pin that race.

**Every refusal is renderable.** Call sites match `IErrorResponseLike` rather than the concrete pay_invoice type, and `errorMessage()` falls back to the code name when a wallet sends `code` without `message` (which NIP-47 permits). The shared helpers live in a new `ui/nwc/NwcResponseMessages.kt`.

The classification deliberately does **not** check for an expected success type. `Nip47ResponseKSerializer` guesses a `result_type`-less success carrying a `preimage` into `PayInvoiceSuccessResponse`, so demanding (say) `PayKeysendSuccessResponse` would report a settled keysend as unreadable — a false failure on a real payment, which is worse than the gap it closes. Since `DecryptCache` funnels every decrypt *and* parse failure into `null`, the model is simply: null is unreadable, `IErrorResponseLike` is a refusal, anything else settled.

**The transactions screen renders its error state** through the existing shared `EmptyState` from `commons`, which wraps the description in a `SelectionContainer` — so the wallet's reason is selectable and copyable for reporting.

**One log line meant two things.** `LocalCache` checked the payment tracker before checking whether it had already processed the event, so a relay replaying a cached kind-23195 (which some NWC relays do on every REQ filter change, as `NWCPaymentFilterAssembler` already documents) produced the same "no pending request is registered" warning as a genuinely late response. Both appeared in the field capture and could not be told apart. The duplicate check now runs first.

**Wallet error text is localized.** Nothing new here was hardcoded, but promoting `_error` to a full-screen state gave top billing to English literals that had previously only appeared as a small line on `WalletDetailScreen`. Those now resolve through resources. `PAYMENT_FAILED` turned out to be an exact duplicate of the existing `send_payment_failed` and is gone; the spoofed-reply timeout built its own plural with two inline `if/else` switches and is now a real `<plurals>`.

## Testing

**On-device**, Pixel 9a, benchmark build, against a live NIP-47 wallet service with logcat captured throughout:

| Case | Before | After |
| --- | --- | --- |
| `QUOTA_EXCEEDED` on `pay_invoice`, over per-payment cap | nothing at all | "Your wallet connect provider returned the following error: this connection may not pay more than 30000 msat at once" |
| `RESTRICTED` on `list_transactions` | "No transactions yet" | "Could not load transactions" + "this connection may not use that method" + Refresh |
| Response lost in transit | nothing, forever | "Your wallet did not respond within 60 seconds. The payment may or may not have gone through — check your wallet before retrying." |

The over-limit zap was run twice on the same pairing and produced **both** outcomes: the timeout message on the first attempt, the wallet's actual reason on the second. Same code, same wallet, same limit — so the difference is transport, not client-side dispatch. That resolves the open question in the original report: the intermittent silence was never a rendering bug for that case, it was a lost response that the old client turned into nothing. The same capture shows two relays returning `ERROR: too many concurrent REQs` at the moment of the zap, and the reporter's own wallet-side logs recorded `no relay accepted an NWC response` with 40% of websocket upgrades to one relay failing.

The successful run shows no `NwcSignerState` give-up line and no `LocalCache` warning, which is the correct signature of a clean match — the response arrived, matched, and cancelled the timer.

**Automated:** 4 new tests in `NwcPaymentTrackerTest` (give-up wins, response wins, spoofed-then-give-up, unknown request) and 4 in `ResponseTest` pinning both field-observed payloads plus the no-`result_type` and no-`message` shapes. `:quartz:jvmTest`, `:commons:jvmTest` and `:amethyst:testPlayDebugUnitTest` all pass; play, fdroid and desktop all compile; spotless clean; Android lint reports 0 errors with no finding on any string added here.

## Known gaps, deliberately not addressed

These are pre-existing and out of scope for this fix, but worth recording since they sit next to it:

- **`WalletSendScreen` and `WalletReceiveScreen` still have no timeout.** They hang on "Sending payment…" / "Creating invoice…" indefinitely. Adding one is easy now that the parameter exists, but it is new behaviour rather than a fix to the reported bug.
- **Two timeout mechanisms for one request.** `WalletViewModel` retires its entries at 30s without dropping the subscription, while `NwcSignerState` does both at 60s. They compose only because 30 < 60, and the shorter one can tell the user "timed out" for a request the wallet answers at 40s. Consolidating them needs a product call on which duration is right.
- **`onTimeout` defaults to silence.** The default is still the old behaviour for any caller that omits it. A sealed outcome type would make omission a compile error — `desktopApp` already models this as `PaymentResult { Success | Error | Timeout }` — but that is ~20 call sites across three modules.
- **`errorMessage()` can surface an untranslated code name** such as `QUOTA_EXCEEDED` when a wallet omits `message`. It is wrapped in a localized sentence, and it beats the previous "Error parsing error message". Localizing all 15 `NwcErrorCode` values belongs in `amethyst`, not in protocol code.

## Review notes

Four commits, each self-contained: the fixes, a cleanup pass, findings from a Kotlin review, then the log disambiguation and localization. `desktopApp`'s `processResponse` had the same narrow type check and is fixed alongside; it reported "unexpected response type" instead of the wallet's reason.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountZapActions.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountZapActions.kt
@@ -99,7 +99,7 @@ class AccountZapActions(
 
     suspend fun sendNwcRequest(
         request: Request,
-        onTimeout: (() -> Unit)? = null,
+        onTimeout: () -> Unit = {},
         onResponse: (Response?) -> Unit,
     ) {
         val (event, relay) = account.nip47SignerState.sendNwcRequest(request, onTimeout, onResponse)
@@ -109,7 +109,7 @@ class AccountZapActions(
     suspend fun sendNwcRequestToWallet(
         walletUri: Nip47WalletConnect.Nip47URINorm,
         request: Request,
-        onTimeout: (() -> Unit)? = null,
+        onTimeout: () -> Unit = {},
         onResponse: (Response?) -> Unit,
     ): HexKey {
         val (event, relay) = account.nip47SignerState.sendNwcRequestToWallet(walletUri, request, onTimeout, onResponse)
@@ -138,7 +138,7 @@ class AccountZapActions(
     suspend fun sendZapPaymentRequestFor(
         bolt11: String,
         zappedNote: Note?,
-        onTimeout: (() -> Unit)? = null,
+        onTimeout: () -> Unit = {},
         onResponse: (Response?) -> Unit,
     ) {
         val (event, relay) = account.nip47SignerState.sendZapPaymentRequestFor(bolt11, zappedNote, onTimeout, onResponse)
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountZapActions.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountZapActions.kt
@@ -99,18 +99,20 @@ class AccountZapActions(
 
     suspend fun sendNwcRequest(
         request: Request,
+        onTimeout: (() -> Unit)? = null,
         onResponse: (Response?) -> Unit,
     ) {
-        val (event, relay) = account.nip47SignerState.sendNwcRequest(request, onResponse)
+        val (event, relay) = account.nip47SignerState.sendNwcRequest(request, onTimeout, onResponse)
         account.client.publish(event, setOf(relay))
     }
 
     suspend fun sendNwcRequestToWallet(
         walletUri: Nip47WalletConnect.Nip47URINorm,
         request: Request,
+        onTimeout: (() -> Unit)? = null,
         onResponse: (Response?) -> Unit,
     ): HexKey {
-        val (event, relay) = account.nip47SignerState.sendNwcRequestToWallet(walletUri, request, onResponse)
+        val (event, relay) = account.nip47SignerState.sendNwcRequestToWallet(walletUri, request, onTimeout, onResponse)
         account.client.publish(event, setOf(relay))
         return event.id
     }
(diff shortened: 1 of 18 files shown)


vitorpamplona:
Doesn't this get too annoying? I don't want to repeat the Relay Login fiasco we just had. :(

davotoula:
Fair point but not the same. The fix is in two places:

If a zap times out it shows a helpful error message rather than just silently gone. 

In wallet section, transactions. If you have disabled nwc privilege "show transactions" then it shows a message (not a pop-up) rather than the existing "no transactions".

PR #3994, opened 2026-08-26
fix(lists): stop deleted lists showing their UUID in the feed filter picker

### LINKED ISSUE #3949: Deleted lists continue to display in Home view menu
Deleted lists continue to display in the "Select an option to filter the feed" menu on the Home view. The deleted lists display as 32 character hex string (e.g., UUID set of 8, 4, 4, 4, 12 characters separated by dashes).

**To Reproduce**
1. Create new list in "My Lists"
2. Delete the list.
3. Navigate to Home screen
4. Open top menu, "Select an option to filter..." to select the feed (e.g., Feeds, Relays, Locations, Lists)
5. Deleted list remains as UUID

**Device**
 - Android Version: Android 17
 - App Version: v1.13.1
 - App Flavour: FDROID

greenart7c3:
Fixes #3949

## Summary

Deleting a NIP-51 people list (kind 30000) or follow pack (kind 39089) left a null-event `AddressableNote` behind, and the persisted per-screen `TopFilter` that still pointed at the address re-created that empty shell on every app start via `getOrCreateAddressableNote`. With no event left, `PeopleListName.name()` fell back to `note.dTag()` — the raw UUID — so the deleted list kept showing in the top-bar feed filter ("Select an option to filter the feed"), even after a restart.

Four changes:

- **`DeletionIndex` (quartz)** — new `hasBeenDeleted(address, pubKey)` overload that checks the NIP-09 index by the `kind:pubkey:dTag` reference, so deletion status can be checked even when the addressable's event is not loaded (empty shell).
- **`LocalCache.hasBeenDeleted(address)`** — exposes that check for addressable shells.
- **`PeopleListsState.existingPeopleListNotes()` / `FollowListsState.existingPeopleListNotes()`** — hide a null-event shell **only when the deletion index shows its address was deleted by its own author** (`it.event != null || !cache.hasBeenDeleted(it.address)`). Shells that are merely not loaded yet stay in the list so the UI flow that opens the picker can still subscribe and fetch them from relays.
- **`deleteFollowSet()` (both states)** — after signing+sending the NIP-09 deletion, reset any persisted `default*FollowList` that still points at the deleted address back to that screen's factory default (new `AccountSettings.resetFeedFiltersPointingTo`), so no dangling filter survives a restart.

## Review history

The first revision filtered out every addressable without a loaded event (`it.event != null`). As pointed out in review, that is not a deletion check — it also removed legitimate not-yet-cached lists, blocking the UI flow that subscribes and loads addressables on demand. This revision only hides shells whose address has a matching NIP-09 deletion record in the `DeletionIndex`; unloaded ones stay loadable.

## Test plan

- `./gradlew :amethyst:compileFdroidDebugKotlin` — clean.
- `./gradlew :quartz:jvmTest :commons:jvmTest` and `./gradlew spotlessCheck :quartz:verifyKmpPurity :commons:verifyKmpPurity` — pass.
- Manual (reproduces #3949): create a list in My Lists → select it in the Home top-bar filter → delete it. Before: the picker keeps an entry whose name is a 32-char UUID, and it survives restarts. After: the entry disappears from the picker immediately and the screen's filter resets to its default (e.g. "Follows" on Home).

Note: substantial AI involvement — this PR was written with the help of an AI coding assistant.

No interop suites needed: the change only affects local cache filtering and per-screen persisted settings, not wire bytes or event handling.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
@@ -788,6 +788,62 @@ class AccountSettings(
     // list names
     // ---
 
+    /**
+     * All per-screen persisted feed filters paired with their factory default.
+     * Deleting a list (NIP-51 people list / follow pack) must reset any screen whose
+     * filter still points at the deleted address — otherwise the screen keeps a
+     * dangling [TopFilter.PeopleList] that re-creates an empty AddressableNote shell
+     * on every start and shows the list's dTag/UUID in the top bar instead of a name.
+     */
+    private val feedFiltersWithDefaults: List<Pair<MutableStateFlow<TopFilter>, TopFilter>> =
+        listOf(
+            defaultHomeFollowList to TopFilter.AllFollows,
+            defaultStoriesFollowList to TopFilter.Global,
+            defaultNotificationFollowList to TopFilter.Selected,
+            defaultDiscoveryFollowList to TopFilter.Global,
+            defaultPollsFollowList to TopFilter.Global,
+            defaultPicturesFollowList to TopFilter.Global,
+            defaultNappletsFollowList to TopFilter.Global,
+            defaultNsitesFollowList to TopFilter.Global,
+            defaultWorkoutsFollowList to TopFilter.Global,
+            defaultGitRepositoriesFollowList to TopFilter.Global,
+            defaultHighlightsFollowList to TopFilter.Global,
+            defaultCalendarsFollowList to TopFilter.Global,
+            defaultProductsFollowList to TopFilter.AroundMe,
+            defaultShortsFollowList to TopFilter.Global,
+            defaultPublicChatsFollowList to TopFilter.Global,
+            defaultLiveStreamsFollowList to TopFilter.Global,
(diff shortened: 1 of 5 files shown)


vitorpamplona:
Wouldn't removing them from the UI block legitimate events from loading when things are not locally cached? Because this is not checking if it has been deleted, it's removing all non-loaded events from the base state class

Because the UI is generally the flow that subscribes to create the REQ with relays and load addressables on demand (generally when the TopFilter list is clicked and opens).

greenart7c3:
Good point — filtering on `event != null` was checking the cache state, not deletion status, and would hide legitimate not-yet-loaded lists exactly as you describe.

Reworked in e0e2e427c4: the picker now only hides a null-event shell when the `DeletionIndex` shows its address (`kind:pubkey:dtag`) was deleted by the list's own author — new `DeletionIndex.hasBeenDeleted(address, pubKey)` overload + `LocalCache.hasBeenDeleted(address)`. Shells that are merely not loaded stay in the list, so the UI flow that subscribes/REQs addressables on demand keeps working and can fetch them from relays. The reset of persisted TopFilters pointing at the deleted address (`resetFeedFiltersPointingTo`) is unchanged.

PR #3997, opened 2026-08-27
IPFS:// support alongside http:// for better content survival. 

besoeasy:
@vitorpamplona  REFERENCE : https://github.com/vitorpamplona/amethyst/pull/3934


vitorpamplona:
The code is not ready yet. If you say that the API is not compatible with Blossom, then you cannot use the Blossom Server State on the Account, which saves the url on the Blossom server list event. Otherwise, other apps will expect a full Blossom API and will not work well with our users' list. You will need to specify your own event kind for your own servers. 

I don't think that is necessary. Your server can just reply as if it were a regular Blossom API and everything would work in all Nostr clients, for both upload and display.

besoeasy:
@vitorpamplona

Agreed on kind 10063: Originless is not Blossom, so those URLs must not go on the user’s Blossom server list. Other clients would expect BUD-01 against that host.

This PR already keeps Originless off 10063 (and off NIP-96 kind 10096). Nodes are a local list on the Originless settings page. Media Servers still publishes only real Blossom URLs.

The leftover is [BlossomServerListState.hostNameFlow](https://github.com/vitorpamplona/amethyst/pull/3997#) feeding the upload picker when Originless uploads are on. That never writes an event, but it should not live in Blossom state. I’ll move that merge out.

A new event kind is not needed. Originless is optional and Amethyst-local.

We are not making Originless speak Blossom. The point of this PR is notes that store ipfs://CID so the file survives any one HTTP host. A Blossom-compatible Originless would put Blossom HTTP URLs in notes and lose that. Upload stays a simple unauthenticated POST /upload; display rewrites ipfs:// through the user’s Originless gateways.

besoeasy:
Screenshot_20260827-081836_Amy Debug
Screenshot_20260827-081932_Amy Debug

Checked, Works as expected, no list publishing

vitorpamplona:
> A new event kind is not needed. Originless is optional and Amethyst-local.

If you are placing IPFS in Nostr notes, every client will need to code a way to load those files, otherwise nobody will use the IPFS server because nobody else can see their post. 

That means, every client will need to code some form of server/gateway list to be used. 

Which means that when users change clients, which happens multiple times a day, they want a single list in all clients. 

Which means Originless is not optional (otherwise, nobody will see the image) and Amethyst-local will make it very annoying for users because they have to set up the same list in every one of their 10+ Nostr clients they use.

besoeasy:
i understand @vitorpamplona i already agree with you. but we have to start slowly. 

> If you are placing IPFS in Nostr notes, every client will need to code a way to load those files, otherwise nobody will use the IPFS server because nobody else can see their post.

i understand boss, but for now, 0xchat and amethyst. works fine, that is good chunk already. 

that is why its opt in, and not a default. we will slowly get der with time on all clients.

but this does no harm, and its optional. plus benefits are huge for people living in INDIA and CHINA

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/utils/ChatFileUploadDialog.kt
```diff
@@ -216,7 +216,9 @@ private fun ImageVideoPostChat(
         )
     }
 
-    if (fileUploadState.multiOrchestrator?.hasCompressible() == true) {
+    if (fileUploadState.selectedServer.type.usesClientMediaCompression &&
```
vitorpamplona, line 219: why did you remove this condition?

amethyst/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
```diff
@@ -865,7 +874,11 @@ object LocalPreferences {
                         async {
                             parseOrNull<List<ClinkDebitWalletEntry>>(clinkDebitWalletsStr)?.mapNotNull { it.normalize() } ?: emptyList()
                         }
-                    val defaultFileServer = async { parseOrNull<ServerName>(defaultFileServerStr) ?: DEFAULT_MEDIA_SERVERS[0] }
+                    val defaultFileServer =
+                        async {
+                            val parsed = parseOrNull<ServerName>(defaultFileServerStr) ?: DEFAULT_MEDIA_SERVERS[0]
+                            if (parsed.type == ServerType.Originless) ORIGINLESS_UPLOAD_TARGET else parsed
```
vitorpamplona: why is this change needed? shouldn't the upload targed be saved as prefered server and decompiled correctly when parsed?

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewMediaView.kt
```diff
@@ -218,52 +218,56 @@ fun ImageVideoPost(
         )
     }
 
-    Column(
```
vitorpamplona, line 221: why are you not showing the media quality? Almost every upload should be compacted and exif removed before upload.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/images/IpfsFetcher.kt
```diff
@@ -0,0 +1,90 @@
...
+
+            return IpfsFetcher(candidates) { url ->
+                NetworkFetcher(
+                    url = url,
+                    options = options,
+                    networkClient = lazy { networkClient(url).asNetworkClient() },
+                    diskCache = lazy { imageLoader.diskCache },
+                    cacheStrategy = cacheStrategyLazy,
+                    connectivityChecker = lazy { connectivityCheckerLazy.get(options.context) },
+                    concurrentRequestStrategy = lazy { DeDupeConcurrentRequestStrategy() },
```
vitorpamplona: This is no good.. you are recreating the cache in every request. Just folow the way the others work

amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
```diff
@@ -3631,6 +3632,14 @@ class Account(
     init {
         Log.d("AccountRegisterObservers", "Init")
 
+        // Originless nodes always resolve `ipfs://`, even when Originless uploads are off.
+        IpfsGatewayResolver.currentServerBases = settings.originlessServerUrls.value
+        scope.launch {
+            settings.originlessServerUrls.collect { urls ->
+                IpfsGatewayResolver.currentServerBases = urls
```
vitorpamplona: Why saving a static IpfsGatewayResolver instance if you can just access the direct flow?  Remember, we have multiple account running at the same time. They are just going to replace the url of whatever the last account is using.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
```diff
@@ -195,6 +199,20 @@ class AccountSettings(
...
     var localRelayServers: MutableStateFlow<Set<String>> = MutableStateFlow(setOf()),
     var defaultFileServer: ServerName = DEFAULT_MEDIA_SERVERS[0],
+    /**
+     * Local Originless node URLs (NIP-96-simple: `POST /upload` or `POST /media`,
+     * `GET /ipfs/{cid}`). Not a kind-10063 list. Uploads pin to every entry;
+     * there is no default upload node. `ipfs://` fetches try each gateway until
+     * one succeeds, falling back to OriginlessUrls.DEFAULT_SERVER when this list
+     * is empty.
+     */
+    val originlessServerUrls: MutableStateFlow<List<String>> = MutableStateFlow(emptyList()),
```
vitorpamplona: This should be turned into a Nostr Event, just like the Blossom Server List event (you can just copy from the Blossom spec. In that way, we don't need to save them locally.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
```diff
@@ -195,6 +199,20 @@ class AccountSettings(
...
+    /**
+     * When true, compose uploads go to Originless (`ipfs://`) and Blossom/NIP-96
+     * targets are hidden from the picker. `ipfs://` fetches always use
+     * [originlessServerUrls], even when this is false.
+     */
+    val originlessUploadsEnabled: MutableStateFlow<Boolean> = MutableStateFlow(false),
```
vitorpamplona, line 206: There shouldn't be a separate flag to enable or disable. If the user has an originless event kind, then you should show side by side with the Blossom relays (if the user has any setup).

besoeasy:
@vitorpamplona works beautifully for images, audio.

M sure you will love it over time too. 

The one with originless - can't be taken down, no DCMA, no blocks 

https://njump.to/naddr1qqjrsdf3x3skycec943xxcmp956ryeps95urqdfk95urwdnyxv6kgvej8qexvqgawaehxw309ahx7um5wgknqvfw09skk6tgdahxuefwvdhk6tczyqmthj5ppc4w47ynjzl7x4dzdrdrr5ps5j0hc3eyl8g65e8glt70cqcyqqqglvccglmxa

The one with blossom - easy to take down, DCMA, domain is liable for content hosted

https://njump.to/naddr1qq8hgunpvd4j66rrwajhwvn0d43qz8thwden5te0dehhxarj95crztnev94kj6r0dehx2tnrdakj7q3qsf3zslxxvmxckewwsxmzj60a4drdf85d2nw78nyj977udcuu43hqxpqqqz8mxcf65lr

From my side, its stable enough to merge. Thank you for being reasonable and having a broad vision 🙏

amethyst/src/main/java/com/vitorpamplona/amethyst/service/okhttp/EncryptionKeyCache.kt
```diff
@@ -36,8 +41,11 @@ class EncryptionKeyCache {
         url: String?,
         decryptInformation: DecryptInformation,
     ) {
-        if (cache.get(url) == null) {
-            cache.put(url, decryptInformation)
+        if (url == null) return
+        IpfsGatewayResolver.decryptionKeyUrls(url).forEach { alias ->
```
vitorpamplona: This class should not even know which transport layer the user is using. Let's rethink about this. We need to move this out.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
```diff
@@ -81,12 +82,29 @@ class ImageDownloader {
...
+        }
+        return null
+    }
 
     suspend fun waitAndGetImage(
         imageUrl: String,
         okHttpClient: (url: String) -> OkHttpClient,
-    ): Blob? = retryWithDelay { tryGetTheImage(imageUrl, okHttpClient) }
+    ): Blob? {
+        val urls = IpfsGatewayResolver.httpFetchUrls(imageUrl)
```
vitorpamplona: Why does this need IPFS treatment and not Blossom treatment?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
```diff
@@ -323,10 +371,22 @@ class UploadOrchestrator {
 
         updateState(0.6, UploadingState.Downloading)
 
-        // Use streaming verification for memory efficiency with large files
-        val verification =
-            ImageDownloader().waitAndVerifyStream(uploadResult.url, okHttpClient)
-                ?: return error(R.string.could_not_download_from_the_server)
+        val urlsToVerify =
```
vitorpamplona: Why does this need a special handler for IPFS but not blossom: uris?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/hls/HlsBlobUploaderFactory.kt
```diff
@@ -124,4 +130,30 @@ object HlsBlobUploaderFactory {
                 context = context,
             )
         }
+
+    private fun originlessAdapter(
```
vitorpamplona, line 134: Did you test this video streaming setup?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
```diff
@@ -81,12 +82,29 @@ class ImageDownloader {
     suspend fun waitAndVerifyStream(
         imageUrl: String,
         okHttpClient: (url: String) -> OkHttpClient,
-    ): StreamVerification? = retryWithDelay { tryStreamAndVerify(imageUrl, okHttpClient) }
+        maxAttempts: Int = 15,
+    ): StreamVerification? {
+        val urls = IpfsGatewayResolver.httpFetchUrls(imageUrl)
```
vitorpamplona: Why does this need IPFS treatment and not Blossom treatment? Was blossom: uris failing?

commons/src/jvmAndroid/kotlin/com/vitorpamplona/amethyst/commons/originless/OriginlessGatewayFailoverInterceptor.kt
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * When a GET/HEAD to a configured Originless `{base}/ipfs/{cid}` fails, retries
+ * the same CID on the rest of the user's Originless nodes.
+ *
+ * Notes store `ipfs://CID`; Coil/PdfFetcher rewrite that to the first gateway.
+ * This interceptor is how a later node in the list still serves the file when
+ * the first is down or never pinned that CID.
+ */
+class OriginlessGatewayFailoverInterceptor(
```
vitorpamplona, line 37: Why do you need this interceptor if you already loop through the sever URLS on other places in the code? Isn't this duplicating checking of the same servers?

vitorpamplona (commented):
Just pick a number for the event kind of your server list. Otherwise, we will need to remove all of this later to redo with an event and migrate all the users to the new scheme..

vitorpamplona:
> The one with originless - can't be taken down, no DCMA, no blocks

That is def not correct: 

Screenshot 2026-08-26 at 11 40 40 PM

besoeasy:
> vitorpamplona

thats for public node, not for your node, it runs unfiltered network, the screenshot you showed, evem they have to delete the CID manually using API.

besoeasy:
Yes. Originless node URLs should stay local. Publishing them as a public Nostr event is the wrong model, and it is unsafe.

Originless is a private pin node, not a public media host like Blossom. Umbret listing is "private, decentralized file sharing." The API is unauthenticated POST /upload. A published list is an open dump onto someone's Umbrel/localhost:3232/home IP. Kind 10063 is worse: other apps would treat those URLs as full Blossom servers.

Vitor is still thinking in Blossom terms. In Blossom the note stores `https://that-server/sha256`, so other clients need your server URL, which is why kind 10063 exists. Originless notes store `ipfs://CID`. That is a content address, not a host. Readers do not need the uploader's node. They resolve the CID through their Originless list, then a public Originless gateway (originless.gupt.app, originless.besoeasy.com, …). That is why fetch already falls back to DEFAULT_SERVER and why the node list is not a protocol event.

vitorpamplona:
> Publishing them as a public Nostr event is the wrong model, and it is unsafe.

You can use the encrypted part of the event. It will save a list just for the user to use in all Nostr apps. We have tons of those settings already.. This is just one more. Super easy, and then it syncs between all the user's Nostr clients.

besoeasy:
> > Publishing them as a public Nostr event is the wrong model, and it is unsafe.
> 
> You can use the encrypted part of the event. It will save a list just for the user to use in all Nostr apps. We have tons of those settings already.. This is just one more. Super easy, and then it syncs between all the user's Nostr clients.

you are right, i slipped out of mind, that we can encrypt that.

vitorpamplona:
I didn't say that others would be using any of the author's Originless nodes. But it would have been much faster if they could, since the app would hit the source node faster. The current images you were testing take about 5 seconds to load from a regular dweb.link pull from here

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
besoeasy: Blossom uploads return https://…. blossom: is Coil/BlossomFetcher. Originless uploads return ipfs://CID, which is not HTTP, so those two paths have to expand gateways. blossom: was not failing.

vitorpamplona: hum.. I am trying to figure out a better way to do this... since testing if it is IPFS everywhere in the code smells... Its late here but I will keep thinking.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/utils/ChatFileUploadDialog.kt
besoeasy: We didn't remove hasCompressible() — it's still there. We added usesClientMediaCompression so Originless doesn't show Media Quality.

Originless pins the original bytes as ipfs://CID. We don't transcode unless the user asks (Originless settings → Optimize media → POST /media). Blossom/NIP-96 still show the slider whenever the file is compressible.

vitorpamplona: But you should keep the option, right? It's the user's choice to compress or not, not the server's.

amethyst/src/main/java/com/vitorpamplona/amethyst/LocalPreferences.kt
besoeasy: You're right. We already save ORIGINLESS_UPLOAD_TARGET (originless://all) as the preferred server, so it should just parse back. Dropped the remap.

Originless still isn't one HTTPS host — uploads fan out to every node in the local list, and that sentinel is what the picker stores. If an older build saved a per-node URL, [UploadServerListState.resetTargetOrNull](https://github.com/vitorpamplona/amethyst/pull/3997#) snaps it to the sentinel

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/screen/loggedIn/chats/utils/ChatFileUploadDialog.kt
vitorpamplona: Otherwise, even if I run my own Originless it's either compress everything or compress nothing. We had that before, and nobody liked it. That's why we offer the option on the client-side and by default is compressed because otherwise people's data plans will go faster than their salary by just scrolling the feed :)

amethyst/src/main/java/com/vitorpamplona/amethyst/service/images/IpfsFetcher.kt
besoeasy: Agreed — lazy { DeDupeConcurrentRequestStrategy() } was a private map per fetch, so IPFS never de-duped with the rest.

[IpfsFetcher.Factory](https://github.com/vitorpamplona/amethyst/pull/3997#) now takes ImageLoaderSetup's strategy and keeps it in lazyOf(concurrentRequestStrategy), same as BlossomFetcher / OkHttpFactory / ProfilePictureFetcher.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/Account.kt
besoeasy: Agreed — that collect was last-writer-wins across logged-in accounts.

Dropped it. Coil/OkHttp are process-wide, so they still can't take the Account flow per request. AppModules binds [IpfsGatewayResolver.serverBasesProvider](https://github.com/vitorpamplona/amethyst/pull/3997#) to sessionManager.loggedInAccount()?.settings?.originlessServerUrls and reads it at fetch time. Switching accounts changes the gateway list; two [Account.init](https://github.com/vitorpamplona/amethyst/pull/3997#)s no longer overwrite each other

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/ImageDownloader.kt
besoeasy: Blossom uploads return https://…. blossom: is Coil/BlossomFetcher. Originless uploads return ipfs://CID, which OkHttp can't fetch as-is, so we rewrite once with toHttpUrl.

The multi-URL loop is gone — ImageDownloader now uses OkHttp, so OriginlessGatewayFailoverInterceptor handles the other nodes. blossom: wasn't failing.

commons/src/jvmAndroid/kotlin/com/vitorpamplona/amethyst/commons/originless/OriginlessGatewayFailoverInterceptor.kt
besoeasy: Yes, that was duplicating the same node list.

Coil and ExoPlayer only have one HTTP URL (toHttpUrl → first gateway), so they need the interceptor when that node is down or never pinned the CID.

ImageDownloader / UploadOrchestrator were walking the same list again. They now rewrite once and go through OkHttp, so failover is only the interceptor.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
besoeasy: It doesn't anymore. verifyHeader just calls waitAndVerifyStream(uploadResult.url) like everything else.

Blossom returns https://…, which downloads as-is. Originless returns ipfs://CID, so ImageDownloader rewrites that once to HTTP; OkHttp's interceptor failovers. blossom: never went through this method.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
besoeasy: We don't want that. Originless nodes are often private (Umbrel, LAN, unauthenticated POST /upload). A kind-10063-style event would publish those URLs, and other apps would treat them as a public Blossom API, which they aren't.

Blossom's list is meant to be shared. This one isn't — it's a local fetch/upload target for ipfs://. Notes only store ipfs://CID; other clients can use any gateway. Empty list still fetches from https://originless.gupt.app/

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
besoeasy: There's no Originless event, and we don't want one (those URLs can be private).

The flag is the opt-in for uploading via Originless. Nodes in the list always resolve ipfs://, even with the switch off. Turning it on hides Blossom/NIP-96 from the picker and pins to every Originless node. Side-by-side would mix an unauthenticated POST /upload API with Blossom, which other apps would then treat as Blossom.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/okhttp/EncryptionKeyCache.kt
besoeasy: Agreed — this class shouldn't know about IPFS.

add / get are exact URL lookups again. Kind 15 still stores ipfs://CID while OkHttp fetches {gateway}/ipfs/{CID}, so the aliases are registered at the call site (addForMediaUrl from RenderEncryptedFile / Account). The interceptor never imports IpfsGatewayResolver

besoeasy:
@vitorpamplona check and merge

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/hls/HlsBlobUploaderFactory.kt
besoeasy: yes i did

vitorpamplona:
So, I am testing and only media uploaded to originless.gupt.app can be downloaded from it. It doesn't seem to see any other IPFS node. Even if I run my own originless, I can't use it to see everybody else's images. 

Is that how it is supposed to be? If so, then we definitely need the Server list event that other people can access and then the resolver needs to use each author's own IPFS server. If not, then there is something broken in your gateway

besoeasy:
> So, I am testing and only media uploaded to originless.gupt.app can be downloaded from it. It doesn't seem to see any other IPFS node. Even if I run my own originless, I can't use it to see everybody else's images.
> 
> Is that how it is supposed to be? If so, then we definitely need the Server list event that other people can access and then the resolver needs to use each author's own IPFS server. If not, then there is something broken in your gateway

It will take atleast 12 hours before your node is well connected, completely normal.

Make sure your ISP don't block tor/libp2p. 

Do you want me to share a debian server for you to prototype.

vitorpamplona:
> It will take atleast 12 hours before your node is well connected, completely normal.

Might need to add that warning for users, since most IPFS things will be broken until then

vitorpamplona:
12 hours from the moment you started your server or from the moment the picture was sent to an IPFS node that is not yours?

besoeasy:
> 12 hours from the moment you started your server or from the moment the picture was sent to an IPFS node that is not yours?

See due to this

ipfs init --profile=lowpower

It saves so much CPU and memory, but one side effect is peer learning is slow.

12 hours from cold start, doesn't happen everywhere only on home network and arm based devices. 

Once node has good peers, it's almost instant.

https://railway.com/ deployed one here was able to fetch file in 8 minutes after deployment and after it was instant.

vitorpamplona:
I just checked again... they are still inaccessible from your server. And it has been more than 12 hours you have it running, no?

vitorpamplona:
Tested this on a Samsung SM-T220 (Android 14), with a second device running a pre-IPFS build as a control for "what other clients see". Builds clean, and `:amethyst`, `:commons` and `:quartz` unit tests are all green.

## What works

- **Upload path, end to end.** Picked an image in the composer, it uploaded to Originless and produced `ipfs://QmYhhHT8DtLdwm82oPhJFf4Rm9acod5KuNY53T5u3VPFJD`. The `imeta` tag is complete and correct — hash, size, mime, dim, blurhash.
- **Rendering.** An `ipfs://CID` pinned on the configured node renders inline as media.
- **Content survival holds.** A file uploaded through Originless is retrievable from `ipfs.io` afterwards, so the pin really does reach the public network.
- **Settings screen is genuinely nice.** Live "Online" health check, Primary badge, drag-ordering, upload opt-in **off** by default, and an EXIF/GPS-strip toggle that is also off by default with a clear explanation of what it changes.
- **Tor is respected.** IPFS fetches go through the shared `callFactory`, and `OriginlessGatewayFailoverInterceptor` retries via `chain.proceed` rather than building its own client. That's the right call.

## Findings

### 1. Every note with IPFS media gets a corrupt `r` tag

Published event from the app:

```json
["r","ipfs://qmyhhht8dtldwm82ophjff4rm9acod5kuny53t5u3vpfjd/"]
["imeta","url ipfs://QmYhhHT8DtLdwm82oPhJFf4Rm9acod5KuNY53T5u3VPFJD", ...]
```

The `r` tag is lowercased and has a trailing slash; content and `imeta` keep the correct mixed case. CIDv0 is base58btc and case-sensitive, so the `r` value is not a valid CID:

```
GET /ipfs/QmYhhHT8DtLdwm82oPhJFf4Rm9acod5KuNY53T5u3VPFJD  -> 200
GET /ipfs/qmyhhht8dtldwm82ophjff4rm9acod5kuny53t5u3vpfjd  -> 400
    invalid cid: selected encoding not supported
```

Cause: `Url.kt:130` lowercases the host component, and in `ipfs://CID` the CID is parsed as the host — which also explains the added `/`. `findURLs` itself is fine because it returns `originalUrl`; the `r`-tag path uses the normalized form. That's why `findsIpfsUrisForImetaMatching` passes and this still ships broken — the test asserts on `originalUrl`, not on what actually reaches the tag.

### 2. Default configuration silently routes to `originless.gupt.app`, and the UI says otherwise

With nothing configured (confirmed — no `originless` key in shared prefs), the settings page says:

> You have no Originless nodes set. Add one below so ipfs:// links can load.

But they do load. `IpfsGatewayResolver.fetchBases()` falls back to `OriginlessUrls.DEFAULT_SERVER`, so every `ipfs://` fetch goes to `https://originless.gupt.app` — sending the user's IP and IPFS viewing activity to a third-party service they never opted into, while the UI states nothing is configured. Either the fallback should go, or the UI needs to say plainly which server is in use and let people turn it off.

### 3. That default node cannot fetch from the IPFS network

It only serves its own pins. Its own error:

```
failed to resolve /ipfs/QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o: block was not found
locally (offline): ipld: could not find QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o:
retrieval: no providers found for the CID (phase: path resolution)
```

Four public CIDs, following redirects:

```
CID                                                           ipfs.io       originless.gupt.app
QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o                200 /  12 B   404
  (canonical IPFS "hello world")
bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi   200 / 119 KB  404
  (IPFS docs site root)
QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc                200 / 255 KB  404
QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG                200 / 254 KB  404
```

So out of the box, any `ipfs://` link **not** uploaded to that exact node fails to load.

### 4. The title describes a different design than the code

The title says "using public gateways (dweb.link & ipfs.io)". Neither string appears anywhere in the shipped code — only in a test, as an arbitrary custom-gateway argument. The described design would have avoided finding 3; the implemented one doesn't.

### 5. Unresolvable IPFS media degrades to raw text

An explicit `.png` that the gateway 404s renders as a bare link — no broken-image placeholder and no indication that media was intended.

### 6. Interop: this media is invisible to every other client

On a build without `ipfs://` support — which is every current Nostr client and every shipped Amethyst release — all these notes render as plain, unlinked `ipfs://…` text, including images uploaded through the app. An `https://` URL renders everywhere; this doesn't. Worth being a deliberate decision rather than a side effect.

## Not a problem

`114cb13073` deletes `EncryptionKeyCacheTest`, which looks alarming in the log but is fine — that test was added earlier in this same PR, and `android.util.LruCache` is stubbed in JVM unit tests so it could never have passed. No coverage regression.

## Summary

Finding 1 is a straightforward bug and should be fixed before merge — it corrupts a tag on every post with IPFS media.

Findings 2 and 3 are the substantive question: as it stands, the default configuration points at a contributor-operated node that can't serve general IPFS content, without telling the user. That needs a decision on defaults before this can go in.

besoeasy:
@vitorpamplona making a video for you

vitorpamplona:
> @vitorpamplona making a video for you

I don't need a video; I just want it to work :)

besoeasy:
> > @vitorpamplona making a video for you
> 
> I don't need a video; I just want it to work :)

showing you how to do it, just in case u missing something,

besoeasy:
fixing 

5. Unresolvable IPFS media degrades to raw text
An explicit .png that the gateway 404s renders as a bare link — no broken-image placeholder and no indication that media was intended.


this is good reported bug, but i am thinking how to make it better, since ipfs content should be stateless and contain its own media.

vitorpamplona:
I am using your server at https://originless.gupt.app, since you made it the default for all users in the newest commits.... Everyone of our 30,000 users will be hitting that server

besoeasy:
> I am using your server at https://originless.gupt.app, since you made it the default for all users in the newest commits.... Everyone of our 30,000 users will be hitting that server

not a problem if CID is cached and fetched once, i am loving the bugs you reported, i will be pushing some code in 20-30 minutes. i like how how tested this one. 

i was able to fetch file from gupt node i uploaded it from https://console.filebase.com/

image

test on cursor cloud everything work fine. 
swarm_hello_cid_via_local_gateway
dashboard_peers_running


let me push few minor tweaks, and everything should be 10/10

besoeasy:
6. Interop: this media is invisible to every other client

is known limitation as of now.

besoeasy:
@vitorpamplona check now

test now and it should be fine. use latest docker image

vitorpamplona:
Re-tested on the SM-T220 against `45732cb4b2`, and both fixes hold up:

- **CID case in `r` tags** — a new post now publishes `r`, `imeta` and content byte-identical: `ipfs://QmfTYi1cXuRju9g6CJa6wdSP8GRxtdrFMqzrqaQuFyBpU9`. That value fetches 200 `image/png`, while the old lowercased form still 400s. Asserting on `HttpUrlFormatter.normalize()` rather than `originalUrl` closes the hole that let it ship.
- **Silent default gateway** — verified with a CID the device had never cached (freshly uploaded, then posted from a second account, so a cache hit couldn't fake it). With no nodes configured it renders as a plain link and nothing is fetched; re-add a node and it renders. Removing the last node also correctly auto-disables the upload toggle.

Also worth recording, since it came up: the node is genuinely retrieving from the network now, not serving a pin list. A CID harvested from a stranger's Nostr post *after* the change — `QmRA3NWM82ZGynMbYzAgYTSXCVM14Wx1RZ8fKP42G6gjgj`, which I never published — returns 200/184 KB. And the failure mode changed in a way a pin list can't imitate: unavailable content used to give an instant `404 "block was not found locally (offline)"` and now gives a slow timeout, i.e. the node actually goes looking.

## Gateway throughput

Method: one 119 KB file (`bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`) that all four serve, real GETs, escalating load.

| Load | ipfs.io | dweb.link | w3s.link | originless.gupt.app |
|---|---|---|---|---|
| 40 sequential | 40/40 · 8 req/s | 40/40 · 6 req/s | 40/40 · 4 req/s | 40/40 · **1.5 req/s** |
| 60 @ concurrency 20 | 60/60 · 1 s | 60/60 · 2 s | 60/60 · 3 s | 60/60 · 6 s |
| 300 @ concurrency 50 | 300/300 · 9 s (33 req/s) | 300/300 · 9 s (33 req/s) | **63 s** · 4 req/s · 1 drop | not run (single hobby node) |
| 600 @ concurrency 100 | 595/600 · 66 s · 5 drops | 592/600 · 68 s · 5 drops | — | — |

**No rate limiting exists to speak of.** Not one HTTP 429 and no `x-ratelimit-*` / `retry-after` headers at any level. They are Cloudflare-fronted and degrade by slowing down and dropping connections, never by refusing. Feed-level concurrency is comfortably inside the flat part of every curve, so throttling is not a concern for this feature.

## The actual problem: uncached media

Everything above is **cache-hit** measurement (`cf-cache-status: HIT`, one popular CID fetched repeatedly). That is the best case and it flatters every gateway. On a cache miss the numbers are a different story:

- An uncached or unavailable CID takes **28–30 seconds** before failing with 504 — the same on ipfs.io, dweb.link and w3s.link.
- Of **21** `ipfs://` CIDs I harvested from real Nostr posts, only **3** were retrievable from ipfs.io at test time. The other 18 all timed out at ~28–30 s.

A 504 does not strictly prove the content is gone — one of the timeouts was my own image, uploaded minutes earlier and definitely pinned — but that distinction does not help the user. Either way they wait half a minute and get nothing.

That is the risk worth weighing here, not throttling: a feed of `ipfs://` media degrades into long hangs followed by silence, and per finding 5 the failure currently renders as bare `ipfs://` text with no placeholder or error. Whatever the underlying cause, `https://` media does not behave this way.

It also cuts against the "better content survival" premise. On this sample, most real-world `ipfs://` links already do not resolve.

## On the recommended default node

`originless.gupt.app` is **~5× slower than ipfs.io on cache hits** (1.5 vs 8 req/s sequential). During testing it also intermittently returned 504 on its *own* 198-byte pin — a file it had served in 0.5 s a few hours earlier. It handles feed-level concurrency fine (60 requests at concurrency 20 in 6 s), so it is usable, but as the node suggested by default it is the weakest of the four and its behaviour moved around under light load.

I deliberately kept load on it well below the public gateways — it is a single self-hosted node rather than Cloudflare-backed infrastructure, so pushing it to failure would have measured my bandwidth rather than anything useful.

vitorpamplona:
Finished testing the rest of the feature surface, on an SM-T220 (Android 14) and a Pixel 8 (Android 17) with a fresh account. Both earlier fixes hold up, and almost everything works — but I found one crash you'll want before merge.

## 🔴 Crash loop with 2+ Originless nodes

Configuring a second node makes the app die on launch, repeatedly.

```
FATAL EXCEPTION: OkHttp Dispatcher
java.lang.IllegalStateException: cannot make a new request because the
previous response is still open: please call response.close()
  at OriginlessGatewayFailoverInterceptor.intercept(...:73)
```

`intercept()` keeps the failed `Response` open so it can be returned if every gateway misses, then calls `chain.proceed()` again for the next node. OkHttp forbids that.

**Repro:** two nodes where the first 404s a CID, then load any `ipfs://` image.

**Severity:** the feed refetches on start, so it dies on **3 of 3 launches** — and a user in that state can't reach Settings to undo it. The only recovery short of clearing app data (and losing the account) is to disable the network, launch, delete the extra node, re-enable. I verified that recovery works.

It only fires with 2+ nodes — `intercept()` early-returns at `servers.size <= 1` — which is why single-node use never hits it.

Fix:

```diff
                 if (response.code == 404 || response.code == 410 || response.code in 500..599) {
                     lastFailure?.close()
-                    lastFailure = response
+                    // Buffer the error body and release the connection before looping. OkHttp
+                    // throws IllegalStateException ("cannot make a new request because the
+                    // previous response is still open") if chain.proceed() is called while a
+                    // response is open, and this loop has to keep the failure around in case
+                    // every gateway misses. Body size is capped: these are small error pages.
+                    lastFailure =
+                        response
+                            .newBuilder()
+                            .body(response.peekBody(MAX_BUFFERED_ERROR_BYTES))
+                            .build()
+                    response.close()
                     continue
                 }
```

plus `private const val MAX_BUFFERED_ERROR_BYTES = 64L * 1024` in the companion.

### The tests covered this path and still passed

`retriesNextOriginlessNodeOn404` and `retriesNextOriginlessNodeOn500` walk exactly the crashing code, and both were green. `FakeChain` hands back canned responses without enforcing OkHttp's "previous response is still open" rule, so the fake permits what the real client rejects.

I made `FakeChain` enforce it (a `Source` that records whether the body was closed; `proceed` throws OkHttp's exact error otherwise). Mutation-checked both ways: with the old code those two tests now **fail** with the same `IllegalStateException` seen on device; with the fix, all five pass.

Worth noting as a pattern — this is the third defect here that had a passing test over the broken path. The `r`-tag test asserted `originalUrl` rather than the published tag value; these asserted through a fake that didn't model the contract. The production code is decent; the test doubles are too permissive.

## ✅ Verified working

| Area | Result |
|---|---|
| Multi-node **fetch failover** | Works — a CID present only on node 2 still loads while the primary 404s it |
| Multi-node **upload fan-out** | Succeeds even with a broken node in the list |
| **NIP-17 encrypted DM media** | Sent encrypted, decrypted and rendered on the receiving account |
| **Tor** honours "Force Tor when loading images" | Verified: zero direct connections to either gateway, Tor guards active, images still render. Control (regular Blossom image) behaves identically |
| **HLS upload → Originless** | Full path works on the Pixel: real transcode (27 MediaCodec hits), published `ipfs://Qmcd4H…`, resolves 200 as `application/vnd.apple.mpegurl`, plays back in the feed |
| **EXIF `/media` strip** | Genuinely strips — `/upload` keeps Exif+GPS+APP1, `/media` has all three gone (216→126 B). Correctly off by default |
| **Profile picture** via `ipfs://` | Renders |
| **Save to disk** from `ipfs://` | Valid PNG, CID-named |
| **Blossom/NIP-96 not regressed** | Picker reverts correctly, Media Quality slider returns, upload produces a valid Blossom URL |
| `r`-tag CID case / silent gateway | Both earlier fixes confirmed on device |

The Originless settings screen deserves a mention — live health chip, drag-ordering, opt-in upload, EXIF toggle off by default. It's good work.

## Two smaller notes

**`ipfs://` video without `imeta` renders as a still image.** Isolated properly: same app, same account, same CID, only the tag differs. With `imeta m video/mp4` (the app's own upload path) it plays; hand-pasted with no `imeta` it's a frozen frame with no player, while the identical file over `https` gets full controls. A CID is opaque — no extension to classify on — so `imeta` is the only type signal, and the fallback is image. This bites `ipfs://` video from clients that don't emit `imeta`, or a CID re-shared later detached from its original note — which is the "content outlives the post" case this PR is for. Consulting the gateway's `Content-Type` when `imeta` is absent would fix it; Originless already returns `video/mp4` correctly.

**HLS manifests reference `ipfs://` internally**, not http:

```
#EXT-X-STREAM-INF:BANDWIDTH=500000,RESOLUTION=360x644,CODECS="hev1.1.6.L93.B0,mp4a.40.2"
ipfs://QmaQT6iat6hxivvaC9X9kyKATW4Z49fBDxYY7bygHFPwRU

#EXT-X-MAP:URI="ipfs://QmaMREDoGD27SUVv6nDquMqyFn8S7WCiFuHZ8C5xrVxRgC",BYTERANGE="1176@0"
```

No standard HLS player resolves `ipfs://`, so these are playable only by a client that rewrites it through a gateway. Arguably the point of content-addressed HLS, but it means the output is Amethyst-only in a way the UI doesn't convey.

## Things that are *not* PR problems

Recording these so they don't get pinned on this PR:

- **HLS Upload is unusable on the SM-T220** — every rendition checkbox is `enabled="false"`, so Publish never enables. Same with File Server set to Nostr.Build, so it's a device/codec limitation (that tablet reports "H.265 not available"). It works fine on the Pixel 8.
- **Notes containing a mention (`p` tag) don't reach relays** from the app on my test accounts — including uploads to blossom.band, whose npub-subdomain URL gets parsed as a mention. Reproduces identically on plain `main` without this PR, and `amy` publishes the same content fine, so it's pre-existing and unrelated. Flagging separately.

vitorpamplona:
> you are right, i slipped out of mind, that we can encrypt that.

Can you please pick a number and create the settings as the event? I don't want to merge with local settings. In the end... all local settings will be migrated to events anyway. So better start right with this work.

amethyst/src/main/java/com/vitorpamplona/amethyst/model/AccountSettings.kt
vitorpamplona: We need to move to a model where the user can choose IPFS or Blossom in each upload. That's why this setting should not exist. If the user picks a Blossom server, it publishes to the primary server and mirrors to the others. If the user picks an IPFS server, then it publishes to just that one.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/UploadOrchestrator.kt
vitorpamplona: Blossom uses a blossom: uri or a https url. Both work.

vitorpamplona: blossom has two ways: you can do the `blossom:` uri just like IPFS, OR an `https://` url.. Both work. In case of the `blossom:` URI, the server is dynamically found. Only the sha256 is on the post.

amethyst/src/main/java/com/vitorpamplona/amethyst/ui/actions/NewMediaView.kt
vitorpamplona: We need local compression active... this is not optional

besoeasy:
@vitorpamplona done

commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
```diff
@@ -595,9 +599,16 @@ class RichTextParser {
             if (videoExtensions.any { removedParamsFromUrl.endsWith(it) }) return MediaContentKind.VIDEO
             if (pdfExtensions.any { removedParamsFromUrl.endsWith(it) }) return MediaContentKind.PDF
 
+            // Originless pins are content-addressed (`ipfs://CID`) and have no filename.
+            // Without imeta `m` or an extension, default to image so the note renders;
+            // a declared MIME above still wins for video/pdf.
+            if (isExtensionlessIpfsUri(url)) return MediaContentKind.IMAGE
```
vitorpamplona, line 618: this seems like a hack. Why only image? We can send any file to the IPFS server, right?

amethyst/src/main/java/com/vitorpamplona/amethyst/service/okhttp/EncryptionKeyCache.kt
```diff
@@ -36,7 +38,7 @@ class EncryptionKeyCache {
         url: String?,
         decryptInformation: DecryptInformation,
     ) {
-        if (cache.get(url) == null) {
+        if (url != null) {
```
vitorpamplona: Was this change on purpose? Something is off

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
```diff
@@ -149,3 +150,5 @@ enum class CompressorQuality {
     VERY_HIGH,
     UNCOMPRESSED,
 }
+
+fun CompressorQuality.forServer(type: ServerType): CompressorQuality = if (type.usesClientMediaCompression) this else CompressorQuality.UNCOMPRESSED
```
vitorpamplona: Why are you overriding the user's choice? If they want to compress, the app should upload a locally compressed file to IPFS.

vitorpamplona:
Tested the new encrypted node list (kind 10062) end-to-end on an SM-T220 with a fresh account, driving the relay side with `amy`. The format itself is right — servers are NIP-44 private tags, nothing leaks in the public tags — but **the app cannot read the list back**, including the one it just published itself.

## 🔴 kind 10062 is never consumed — the list empties on restart and never syncs

`LocalCache.justConsumeInner` has no branch for `OriginlessServersEvent`, so every 10062 falls through to the `else`:

```
W Event Not Supported: From wss://nos.lol/: {"kind":10062,"tags":[["client","Amethyst"]],...}
W Event Not Supported: From wss://nostr.mom/: {"kind":10062,"tags":[["client","amy"]],...}
```

The first of those is **Amethyst's own event**, published seconds earlier from the settings screen — `justConsumeMyOwnEvent` goes through the same dispatch, so the write path is dropped too.

Everything downstream reads that addressable note:

- `OriginlessServersListState` takes the list off `cache.getOrCreateAddressableNote(...)`, so a list edited elsewhere never arrives.
- `AccountSettings.updateOriginlessServersList` is fed from `originlessServers.getOriginlessServersList()` (Account.kt:3569), i.e. from the same always-null note — so the persisted backup is never written either.

Measured on device: I added a node, confirmed the relay had it, restarted, and got **"You have no Originless nodes set."** The in-session list had come from the settings backup; once the process died there was nothing left. A three-node list published from `amy` was likewise invisible.

Since this commit makes the relay the source of truth, that means a user's node list currently does not survive a launch.

Fix is two lines, next to the sibling list kinds:

```diff
                 is BlockedRelayListEvent,
                 is BlossomServersEvent,
+                is OriginlessServersEvent,
                 is NestsServersEvent,
```

Opened as t3nklabs/amethyst#6 with a regression test (`OriginlessServersIngestionTest`) that asserts the dispatch handles the kind; it fails with the branch removed.

Worth noting the subscription side was already correct — `FilterAccountInfoAndListsFromKey` requests 10062 and the events do arrive. Only the consume dispatch was missing, which is why this looks like a network problem and isn't one.

## ✅ After the fix

| Check | Result |
|---|---|
| Load list published by another client | 3 nodes render in published order, "Primary" on the first |
| Health resolution | The two unreachable hosts show Offline, `originless.gupt.app` Online |
| Edit in app → relay | Deleting a node republished `[["server","https://beta-from-amy.example.org"],["server","https://originless.gupt.app"]]`, correctly NIP-44 encrypted, order preserved |
| Survives cold restart | List is intact (was empty before) |
| Public tags | Only `client`; no server URL in plaintext |

## Two small things

**The `client` tag is inherited from whoever wrote last.** After the app edited a list originally published by `amy`, the republished event still carried `["client","amy"]` — `updateServerList` copies the earlier version's public tags. Preserving unknown tags is defensible, but `client` probably ought to be restamped.

**The settings screen doesn't live-update.** With the screen open, a list changed from another client doesn't appear — I held it for 180 s, then left and re-entered and the new list was there instantly. `OriginlessServersViewModel.refresh()` reads `account.originlessServers.flow.value` once at `load()` and never collects. Flagging it only for completeness: `BlossomServersViewModel` does exactly the same thing (`obtainFileServers()`), so this matches existing behaviour and is **not** a regression from this PR — and snapshotting is arguably right on a screen with pending edits.

besoeasy:
@vitorpamplona thanks for doing this fix, i was going to ask you for help regarding this, since i had no experience with this.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/uploads/MediaCompressor.kt
besoeasy: you are right, i did this to test something, since modifying the file changed its CID.

besoeasy: FIXED

commons/src/commonMain/kotlin/com/vitorpamplona/amethyst/commons/richtext/RichTextParser.kt
besoeasy: You're right — Originless can pin any file. That IMAGE default is only a render heuristic for a bare ipfs://CID with no imeta m and no filename; we guessed image so the note would still show something. MIME/m already wins when it's present, and a path like ipfs://CID/clip.mp4 already classifies from the extension.

amethyst/src/main/java/com/vitorpamplona/amethyst/service/okhttp/EncryptionKeyCache.kt
besoeasy: Not on purpose — sorry. cache.get(url) == null was your first-write-wins guard; we only meant to stop this class from expanding IPFS aliases. We'll put it back: skip a null URL, and don't overwrite an existing cipher.

besoeasy:
@vitorpamplona please check

PR #4002, opened 2026-08-28
fix(zaps): deduplicate LNURL endpoint fetches; stop NWC NIP-04 downgrade on a cold cache

davotoula:
Two independent fixes to zap-related fetching, both found by operating a NIP-57/NIP-47 wallet service against Amethyst and reading the client source to explain what the server saw.

1. **LNURL endpoint fetches were not deduplicated**, so a burst of zap receipts for one lightning address produced one HTTP request per receipt to that provider's `/.well-known/lnurlp/` endpoint. Measured at 211 requests in 3.18 seconds for a single address.
2. **NWC requests silently downgraded to NIP-04** on the first transaction to each wallet after every app launch, even against a wallet advertising `nip44_v2`.

They share no files and can be reviewed independently — commits 1–2 are the LNURL fix, commits 3–5 the NWC one.

---

## 1. LNURL endpoint fetch stampede

`LocalCache.consume(LnZapEvent)` validates each incoming zap receipt against the recipient's LNURL provider `nostrPubkey` (NIP-57 Appendix F). It launches one coroutine per receipt, and each one calls `LnurlEndpointResolver.resolve()`.

The resolver's cache is read-through, so it only helps once a fetch has landed. A burst of receipts for one address therefore missed together: N receipts became N coroutines, N cache misses and N HTTP requests to the same document, followed by N redundant writes of the same value.

This costs the LNURL provider, not Amethyst. The endpoint is unauthenticated and unthrottled by design, and a feed or profile carrying many receipts for one popular address is exactly the burst shape.

### Fix

`LnurlEndpointCache` gains `getOrFetch(url, fetch)`: one in-flight `CompletableDeferred` per URL, with concurrent callers awaiting it rather than starting their own. The eviction policy, capacity and key normalisation are unchanged.

The deduplication lives on the cache rather than in the OkHttp resolver deliberately. The flight map and the store must agree on what counts as "the same address"; keeping them apart meant the resolver had to call `LnurlForm.normalizeUrl` purely to mirror a keying detail private to a class in another module, with nothing enforcing that. Together, the key is computed once and shared, so they cannot drift.

Two properties worth noting for review:

- The cache is populated **before** the in-flight slot is released, so a caller arriving in between reads the result instead of starting a second fetch.
- The slot is released in a non-suspending `finally`, so awaiters are unblocked even if the coroutine that owns the fetch is cancelled. A failed fetch caches nothing, so a provider having a bad minute is retried by the next caller rather than remembered as unresolvable.

### Effect on provider load

For a burst of N receipts for one address, requests to the provider drop from N to 1. The cache has no TTL, so after the first fetch the address is answered from memory for the life of the process.

This is not a client-side latency change. The fetches were already asynchronous and off the render path; the receipts were displayed immediately either way.

### Deliberately not included

- **Negative caching.** A failed fetch still caches nothing, so an endpoint that is down or rate-limiting is re-fetched once per receipt. Deduplication collapses each burst, but not the drip. Choosing a TTL is a policy decision with a real downside (briefly refusing to credit legitimate zaps), so it is left out of a bugfix.
- **`LightningAddressResolver`.** The zap-*sending* path writes to `LnurlEndpointCache` but never reads it. That looks like an omission, but the sending path needs the document's `callback` field and `LnurlEndpointInfo` stores only `nostrPubkey` and `allowsNostr`. Making it read the cache would require widening what that cache is for, and would raise a staleness question it currently avoids: a stale `nostrPubkey` costs an uncredited zap, but a stale `callback` costs a failed payment on a deliberate user action.

---

## 2. NWC NIP-04 downgrade on a cold info cache

NIP-47 says a client "should always prefer nip44 if supported by the wallet service". `NwcSignerState.prefersNip44()` returning false therefore has to mean "this wallet does not offer NIP-44". It meant two different things: that, and "we have not asked yet".

`NwcInfoCache` is per-account and held in memory only, so it starts empty on every app launch. `prefersNip44` read it without waiting, so a cold read returned null, which is indistinguishable from "no NIP-44". The first transaction to each wallet after each launch went out as NIP-04 against a wallet advertising `nip44_v2` — a silent downgrade to deprecated encryption on a payment request.

The existing startup warm-up narrows the window but does not close it: it is asynchronous and races the user's first tap.

### Fix

`NwcInfoCache.currentOrFetch()` waits only when nothing at all is cached, and returns a stale entry as-is. Staleness never caused the downgrade — a stale entry already says which schemes the wallet advertises — so waiting on it would buy nothing and would add a round-trip every TTL period. `prefersNip44` becomes `suspend` and uses it; both call sites were already suspend.

Three properties matter for review, and each is pinned by a test:

- **The wait is capped at 3s.** It sits in front of a payment the user has already tapped, and the fetch itself is bounded only by the relay accessory's default 30s idle window, with the no-response timer not starting until `prefersNip44` returns. On expiry the request goes out as NIP-04 and the fetch keeps running, so the next request gets the negotiated scheme. The cap is on the wait and never on the fetch: a null from the fetch is cached as a definitive "no info event" for the whole TTL, so bounding the fetch would pin the wallet to NIP-04 for days — the opposite of the intent.
- **The fetch runs in the cache's own scope, not the caller's.** A caller on the payment path is a `viewModelScope` coroutine; owning the fetch there meant backing out of the payment screen cancelled it, leaving the cache cold so the next attempt paid the whole cost again. Every caller now only awaits, so giving up cancels one await and nothing else.
- **A dead scope releases its awaiters.** `Account.scope` is cancelled on log-off (`AccountCacheState.removeAccount`), which runs on its own coroutine and can race a payment already in flight. Launching into an already-cancelled scope never runs the body, so the `finally` that completes the deferred never runs either — the caller would await forever and the abandoned in-flight slot would make every later call for that wallet do the same. The deferred is now completed from the job's own completion when the body never got to do it.

Every fetching path in `NwcInfoCache` funnels through one request per wallet. Previously only the background refresh was guarded, and by a plain key set that could not be awaited.

### Effect on latency and load, stated precisely

This is a correctness fix, not a latency optimisation, and it is worth being exact about the direction:

- **It adds latency, bounded and once.** On a genuinely cold cache the first request to a wallet waits up to 3s for its kind 13194 info event instead of proceeding immediately with NIP-04. That is per wallet, per app launch, and only when nothing is cached. Subsequent requests and stale entries do not wait.
- **It reduces relay load slightly.** At most one kind 13194 subscription per wallet is open at a time, regardless of how many paths ask concurrently. This is deduplication of a relay REQ, not of an NWC RPC.
- **It does not change NWC request volume.** The number of `get_balance` / `get_info` / `pay_invoice` requests reaching the wallet service is unchanged. There is no direct reduction in wallet-service load.

---

## Field measurement

A wallet-service operator measured a patched build against stock from the server side, using a self-hosted NIP-47 service (LND on a Raspberry Pi 4). Same nostr account in both clients, and separate NWC pairings configured identically — same backend, same single relay `wss://nos.lol`, same limits.

Method: builds run alternately inside one ten-minute window, three cycles, so time-varying relay conditions apply equally to both. Identical script each arm — force-close, open the wallet five times ~10 s apart, request transactions once, no payments. Requests attributed by the service's per-connection id, with marker requests separating runs.

| Arm | Answered | Unanswered | Wall clock | Gaps ≥8 s | Service handling median/max |
| --- | --- | --- | --- | --- | --- |
| cycle 1 — stock | 42 | 0 | 53 s | — | 18.2 / 74 ms |
| cycle 1 — patched | 33 | 0 | 28 s | — | 47.1 / 113 ms |
| cycle 2 — stock | 31 | 0 | 41 s | 17.6 | 30.4 / 105 ms |
| cycle 2 — patched | 33 | 0 | 26 s | — | 17.0 / 87 ms |
| cycle 3 — stock | 33 | 0 | 104 s | 26.9, 17.4, 11.6, 9.0 | 18.4 / 80 ms |
| cycle 3 — patched | 33 | 0 | 26 s | — | 19.0 / 147 ms |

Totals: stock 106 answered, 0 unanswered, five gaps ≥8 s. Patched 99 answered, 0 unanswered, no gaps ≥8 s.

The service answered 100% of requests in every arm with median handling of 17–47 ms. In cycle 1 the patched build was served *more slowly* than stock (47.1 ms against 18.2 ms median) and still completed the script in half the time, which rules out service latency as the explanation.

### What this data does establish

What it does establish is the thing most relevant to reviewing section 2: **the added cold-cache wait does not degrade responsiveness in practice.** The natural objection to making a payment path wait on a relay round-trip is that it shows up as a slower first interaction. Across three alternating cycles the patched build completed an identical script in 28 s, 26 s and 26 s with no gap over 8 s, against 53 s, 41 s and 104 s for stock.

That evidence applies *a fortiori* to what is proposed here. The measured build carried the first three commits — the wait it contained was **uncapped**, bounded only by the fetch's own 30s idle window. The 3s cap and the scope-ownership fix were written after the measurement and only reduce the worst case, so the shipped behaviour cannot be slower than what was measured.

The separate LNURL measurement — 211 requests for one address in 3.18 s, from a stock client — is direct evidence for section 1, taken from the same operator's logs.

---

## Tests

New tests, each of which fails without the change it covers:

- `LnurlEndpointCacheTest` — 20 concurrent callers for one URL produce one fetch; a failed fetch is not cached and the next caller retries; two spellings of one address (host case, trailing slash) share one fetch.
- `OkHttpLnurlEndpointResolverTest` — a burst of callers produces one HTTP request through a real OkHttp client, proving the resolver routes through the cache.
- `NwcInfoCacheTest` — concurrent callers share one fetch; a caller joins a fetch already started by the background refresh; `currentOrFetch` waits when nothing is cached and returns a stale entry without waiting (the refresh is made to hang, so waiting on it would hang the test); cancelling a caller does not abort the shared fetch; a cache built on an already-cancelled scope releases its callers instead of hanging.

Concurrency tests release their callers through a shared gate rather than sleeping, so they assert "these callers were in flight together" rather than depending on timing.

Each guard was verified by mutation — removing the single-flight, keying the flight map on the raw URL, never releasing the in-flight slot, making the resolver bypass the cache, reverting `currentOrFetch` to the old non-waiting read, putting the fetch back on the caller's coroutine, and removing the dead-scope safety net each fail exactly the test that covers them.

`prefersNip44` itself is a two-line call-site swap covered by the cache tests. `NwcSignerState` has no test harness and constructing one (it needs `LocalCache`, `AccountSettings` and the filter assembler) was out of proportion to the change.

Green: `:quartz:jvmTest`, `:commons:jvmTest`, `:amethyst:testPlayDebugUnitTest`, `:quartz:verifyKmpPurity`, `spotlessCheck`, and compiles for `:amethyst` (fdroid and play), `:desktopApp` and `:commons:compileAndroidMain`.

## Follow-ups deliberately left out

- The single-flight mechanism is now implemented twice — `LnurlEndpointCache` in `quartz` and `NwcInfoCache` in `amethyst`. Extracting a shared `SingleFlight<K, V>` into `quartz/utils` is the right next change, but it adds public API to a shared module and is better reviewed on its own.
- `NwcInfoCache.getFresh` has no production caller. That predates this branch; it is documented as such rather than removed, since deleting it would rewrite pre-existing tests.
- `NwcInfoEvent?` conflates "never fetched" with "wallet published no info event", and the three production readers each encode that differently. Making the distinction explicit in the type is the deeper fix; this PR fixes the call site that was getting it wrong.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip47WalletConnect/NwcInfoCache.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip47WalletConnect/NwcInfoCache.kt
@@ -25,6 +25,7 @@ import com.vitorpamplona.quartz.nip47WalletConnect.Nip47WalletConnect
 import com.vitorpamplona.quartz.nip47WalletConnect.events.NwcInfoEvent
 import com.vitorpamplona.quartz.utils.TimeUtils
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -37,16 +38,23 @@ import java.util.concurrent.ConcurrentHashMap
  * supported RPC methods, and whether it emits notifications.
  *
  * Entries expire after [ttlSeconds] (default 2 days) so a wallet that later
- * changes its advertised capabilities is eventually re-checked. Reads never block
- * on the network:
+ * changes its advertised capabilities is eventually re-checked. Four entry
+ * points, in increasing order of how much they will wait:
  *
  *  - [current] returns whatever is cached (possibly stale, possibly null) with no
- *    side effect — for the payment hot path.
+ *    side effect and never blocks — for callers that can act on "don't know".
  *  - [refreshIfStale] triggers a background fetch when the entry is missing or
  *    expired, and returns immediately — call it right before using a wallet so a
  *    stale entry self-heals without holding up the transaction.
- *  - [getFresh] is the suspending variant for callers that can await (e.g. the
- *    notification watcher deciding whether to open a subscription).
+ *  - [currentOrFetch] waits only when nothing at all is cached, and returns a
+ *    stale entry as-is — for callers where "don't know" and "no" are different
+ *    answers, such as NIP-44 negotiation.
+ *  - [getFresh] waits whenever the entry is missing *or* expired, for a caller
+ *    that must not act on a stale answer. No production caller needs that today.
+ *
+ * Every fetching path funnels through one request per wallet, so the startup
+ * warm-up, a payment waiting on a cold cache and the notification watcher join
+ * the same call rather than racing each other.
  *
  * A completed fetch — including a definitive "wallet published no info event"
  * (null) — is cached with a timestamp. A *failed* fetch (network error/timeout)
@@ -65,7 +73,11 @@ class NwcInfoCache(
     )
 
     private val cache = ConcurrentHashMap<HexKey, Entry>()
-    private val inFlight = ConcurrentHashMap.newKeySet<HexKey>()
+
+    // Fetches in progress, keyed like [cache]. Every path that fetches goes through
+    // [fetchOnce], so the startup warm-up, a payment waiting on a cold cache and the
+    // notification watcher all join one request per wallet instead of racing.
+    private val inFlight = ConcurrentHashMap<HexKey, CompletableDeferred<NwcInfoEvent?>>()
 
     private fun isFresh(entry: Entry): Boolean = now() - entry.fetchedAt < ttlSeconds
 
@@ -80,15 +92,9 @@ class NwcInfoCache(
     fun refreshIfStale(uri: Nip47WalletConnect.Nip47URINorm) {
         val entry = cache[uri.pubKeyHex]
         if (entry != null && isFresh(entry)) return
-        if (!inFlight.add(uri.pubKeyHex)) return
+        if (inFlight.containsKey(uri.pubKeyHex)) return
 
-        scope.launch(Dispatchers.IO) {
-            try {
-                fetchAndStore(uri)
-            } finally {
-                inFlight.remove(uri.pubKeyHex)
-            }
-        }
+        scope.launch(Dispatchers.IO) { fetchOnce(uri) }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


davotoula:
the android job dies at minifyFdroidBenchmarkWithR8 on a resource ceiling, while :quartz:jvmTest and :commons:jvmTest passed on CI and :amethyst:testPlayDebugUnitTest passes locally at 1360/1360.

PR #4011, opened 2026-08-29
fix: rapid settings toggles are silently discarded

davotoula:
## The bug

Toggling rows quickly in **Settings → Side Menu** silently discards some of them. The switch flips, the drawer updates, and then the edit is gone on the next app start — never written to preferences, and only arbitrarily present on relays.

On a device, flipping 16 switches at a normal tapping pace (~2/second) lost **7 of them**.

## Root cause

Synced settings live in one NIP-78 app-specific data event (kind 30078, `d=AmethystSettings`), and that published event is also their *only* local copy: `AccountSettings.backupAppSpecificData` is written when the event comes back through the `LocalCache` collector, and that is what reaches SharedPreferences.

`created_at` has one-second resolution. The Side Menu picker republishes on every discrete toggle, so two toggles a few hundred milliseconds apart stamp the *same* `created_at`. `LocalCache.consumeBaseReplaceable` keeps only the strictly newest version per address, so it drops the second one — the collector never fires, `backupAppSpecificData` keeps the earlier edit, and the restart restores that. Relays are no safer: NIP-01 breaks a `created_at` tie by lowest id, so which edit survives there is effectively random.

The picker was modelled on the bottom-bar picker added three days earlier (`7a41e21272` → `dda0f0d9e8`), and inherited its publish-on-every-change contract along with the bug. `BottomBarSettingsState.togglePin` persists per change in exactly the same way, so **both pickers are affected**; the bottom bar's `commit()` only defers its drag gesture, not its switches.

## The fix

**Publish once the toggles stop, not once per toggle.** The two navigation pickers are the only settings surfaces that write in bursts — a configuring session is a run of discrete taps over ~50 rows, and each one signed, encrypted and published the whole blob to every outbox relay, which on an external signer is an IPC round trip per switch. They now share a 2s debounce, flushed when the screen leaves the composition, when the app stops, and from `onCleared`. The other settings call sites are deliberately untouched: they change one setting at a time, and the published event is that setting's only durable copy, so delaying them would buy nothing and cost durability.

The publisher runs on the account's scope rather than `viewModelScope`, because AndroidX closes `viewModelScope` *before* it invokes `onCleared` (`ViewModel.clear()` closes the keyed closeables, and `viewModelScope` is one of them). A pending publish tied to it would already be cancelled by the time any teardown hook could rescue it. The account's scope is cancelled only when the account is removed, so a pending edit simply completes across an account switch or an Activity teardown.

**Wait for the clock rather than stamping the future.** One second is the real floor on how often an address can be replaced, so a client that replaces one faster has to wait rather than invent a timestamp. `awaitCreatedAtToSupersede` suspends until the second the previous version claimed has passed, then stamps the real time: the new version still wins, and no event is ever dated ahead of the clock. The wait is bounded by `MAX_SUPERSEDE_WAIT_SECONDS` — a version further ahead than that came from another device's skewed clock rather than this client's own burst, and sleeping it out could take hours, so past the bound out-stamping is still the only way to supersede.

It lives in quartz next to `supersedes`: that function decides who wins an address, this one says what a new version must stamp in order to be that winner. The same rule was already written out twice in the app (`RoomParticipantActions`, `AppRecommendationsState`), each with its own comment re-deriving the same NIP-01 argument. `AppRecommendationsState` and the settings blob use the waiting form, since both can accumulate drift across repeated edits and both already suspend under a mutex. `RoomParticipantActions` keeps the non-suspending form — it is reached from Compose click handlers, and its stamp derives from the single event being acted on, so it sits at most one second ahead and cannot drift.

**Stop reporting cancellation as a signer failure.** `java.util.concurrent.CancellationException` extends `IllegalStateException`, so the trailing `catch (e: IllegalStateException)` in the signer error reporter was swallowing every cancelled signer coroutine and toasting it as "signer not found" with the raw exception text. Latent since that arm was written — nothing cancelled those jobs — but the debounce cancels a superseded publish on every rapid edit, which made it fire on essentially every fast toggle. Swallowing it also meant cancellation never propagated.

### Automated

| Suite | Covers |
| --- | --- |
| `AwaitCreatedAtToSupersedeTest` (quartz) | waits out the claimed second instead of stamping ahead; a five-deep burst never runs ahead of the clock; no wait when the clock is already past; out-stamps a version too far ahead to wait out |
| `NextCreatedAtToSupersedeTest` (quartz) | the non-suspending rule kept for callers that cannot suspend |
| `AppSpecificDataRepublishTest` (amethyst) | against the real `LocalCache`: a same-second republish is dropped, a stamped one lands |
| `DebouncedPublisherTest` (amethyst) | a burst publishes once, spaced edits publish separately, flush skips the wait, double-flush does not double-publish, a pending publish dies with the scope it was launched on |

Full `:amethyst` unit suite and `:quartz:jvmTest` pass; `spotlessCheck` clean.

### Manual, on a Pixel 9a (Android 17)

Instrumented the publish path end to end and ran the same test on both branches — flip switches in Settings → Side Menu at a normal tapping pace, force-stop, reopen.

| | before | after |
| --- | --- | --- |
| Toggles | 16 | 23 |
| Events signed | 16 | 3 |
| Rejected by the cache | **7** | **0** |
| Persisted | 9 | 3 |

Before, every rejected event shared its `created_at` with the one immediately preceding it, and no collector or persist step followed it — the edit existed nowhere.

After, the set the UI ended on, the set that was published, and the set restored after the cold start were byte-identical:

```
BADGES BLOSSOM_DATA BOOKMARKS CALENDARS COMMUNITIES DISCOVER EMOJI_PACKS
FAVORITE_ALGO_FEEDS FOLLOW_PACKS INTEREST_SETS LONGS MESSAGES MUSIC_PLAYLISTS
MUSIC_TRACKS NAPPLETS NOTIFICATIONS PICTURES PODCASTS RELAY_GROUPS SHORTS
SOFTWARE_APPS WEB_BOOKMARKS
```

Every published event carried the true wall-clock second — with the debounce in place the sub-second path never ran, so nothing was future-dated even before the waiting change landed. No decrypt or decode failures, and no signer toast under fast clicking.

## License

      original license header.

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip78AppSpecific/AppSpecificState.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip78AppSpecific/AppSpecificState.kt
@@ -25,11 +25,10 @@ import com.vitorpamplona.amethyst.model.AccountSyncedSettingsInternal
 import com.vitorpamplona.amethyst.model.LocalCache
 import com.vitorpamplona.amethyst.model.NoteState
 import com.vitorpamplona.quartz.nip01Core.core.JsonMapper
-import com.vitorpamplona.quartz.nip01Core.core.nextCreatedAtToSupersede
+import com.vitorpamplona.quartz.nip01Core.core.awaitCreatedAtToSupersede
 import com.vitorpamplona.quartz.nip01Core.signers.NostrSigner
 import com.vitorpamplona.quartz.nip78AppData.AppSpecificDataEvent
 import com.vitorpamplona.quartz.utils.Log
-import com.vitorpamplona.quartz.utils.TimeUtils
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
@@ -75,11 +74,7 @@ class AppSpecificState(
         val (toInternal, createdAt) =
             stampOrder.withLock {
                 val snapshot = settings.syncedSettings.toInternal(settings.mutedPublicChats.value)
-                val stamp =
-                    nextCreatedAtToSupersede(
-                        newestKnown = maxOf(lastPublishedAt, amethystSettingsNote.event?.createdAt ?: 0L),
-                        now = TimeUtils.now(),
-                    )
+                val stamp = awaitCreatedAtToSupersede(maxOf(lastPublishedAt, amethystSettingsNote.event?.createdAt ?: 0L))
                 lastPublishedAt = stamp
                 snapshot to stamp
             }
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip78AppSpecific/AppSpecificState.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/nip78AppSpecific/AppSpecificState.kt
@@ -28,11 +28,14 @@ import com.vitorpamplona.quartz.nip01Core.core.JsonMapper
 import com.vitorpamplona.quartz.nip01Core.signers.NostrSigner
 import com.vitorpamplona.quartz.nip78AppData.AppSpecificDataEvent
 import com.vitorpamplona.quartz.utils.Log
+import com.vitorpamplona.quartz.utils.TimeUtils
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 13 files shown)


vitorpamplona:
Does this actually work? `created_at`'s precision is seconds. So nothing on Nostr can update a replaceable event multiple times per second. If you keep pressing and +1-ing each created_at, you will start publishing events in the future, and some relays might block it. 

Maybe we should instead just have a 1s sample/debounce rate to save? So the screen can go crazy and update multiple times a second, the app will only have the last one

davotoula:
This was verified on device while capturing logs. 

Before fix, setting changes were lost during quick tapping. 

After fix all were preserved correctly.

I will do another test, would appreciate any edge cases you can think of!

davotoula:
Added a "wait remainder of a second" fix. Updates are not sent below 1 second intervall.

Testing:

Clean pass:
30 toggles → 4 published events, 4 accepted, 4 persisted, 0 dropped, 0 failures.
@vitorpamplona

PR #4028, opened 2026-08-31
fix(video): use the HLS master playlist and its full quality ladder

vitorpamplona:
NIP-71 video events (kinds 34235/34236) carry a whole HLS ladder in their imeta tags: one entry for the master playlist, which enumerates every rendition, plus one per rendition, each locked to a single resolution. Two separate things kept playback at the bottom of that ladder.

## 1. Renderers took `imetaTags()[0]` blindly

That only works because our own publisher emits the master first (`HlsVideoEventBuilder`). A client that orders the tags differently pinned us to one rung — no adaptive bitrate, and no quality menu either, since a media playlist exposes a single video track and `VideoQualityButton` hides itself below two. The feed filters already accepted an event when *any* imeta was playable, so the tag the card rendered was not necessarily the one that made it pass.

New `VideoEvent.selectVideoTrack()` in `commons/model/nip71Video/`:

- prefers HLS over a progressive file — only HLS can adapt;
- picks the master among the HLS entries: largest declared `dim` with the earliest tag winning ties (our master-first layout), and a **dim-less** manifest ahead of dimensioned rungs, since a ladder-spanning master has no single resolution to declare;
- skips the separate audio track NIP-71 PR #2255 allows and any poster image, falling back to the first imeta when nothing looks like video, so an image-shaped event still renders;
- fills gaps in the chosen entry from its siblings — poster, blurhash, thumbhash, aspect ratio, alt are ladder-wide — so selecting a bare master doesn't cost the card its placeholder or its layout height.

Wired into `VideoCardCompose`, `Video`, `VideoDisplay`, and `NotificationContent.mediaImageUrl` (which now gets a poster instead of falling back to an `.m3u8` URL Coil cannot decode).

## 2. Rendition selection is sized to the player, not to a policy

`VideoViewInner` derived its rendition policy from `isFullscreen` alone, so everything outside the fullscreen dialog was pinned to the lowest rung — including a portrait short rendered full-width, which played at 360p.

The fix is not a second named policy. ExoPlayer already filters renditions by a viewport; its default is the **physical display size** (`TrackSelectionParameters.Builder.init` sets `isViewportSizeLimitedByPhysicalDisplaySize`), which is why adaptive selection on a thumbnail would fetch a display-resolution rung, and why a hard "lowest" pin had to exist as a counterweight. Handing it the *measured* size answers the question directly:

| player | before | after |
|---|---|---|
| full-width short / video card | lowest rung | top of the ladder |
| small inline attachment | lowest rung | the rung matching its pixels |
| fullscreen dialog | display-capped adaptive | unchanged in practice |
| PiP window | lowest rung | small, because the window is small |

All of them still adapt to the connection under that ceiling.

`setViewportSize` is safe where `setMaxVideoSize` would not be: `DefaultTrackSelector` derives its retain threshold from an actual rendition ("the smallest to exceed the maximum size at which it can be displayed within the viewport") and leaves the group untouched when nothing covers the viewport, so a small player can never filter every track away. Manual picks from the quality menu still win — overrides are re-applied after constraint-based selection. Both verified against media3 1.11.0 sources.

The measurement rides the `onSizeChanged` `RenderVideoPlayer` already had for double-tap seeking: no new layout observation, no recomposition, and a guard keeps a settling layout pass from re-running track selection over IPC.

`VideoQualityPolicy`, `ApplyInitialVideoQuality` and `findLowestResolutionTrackIndex` are gone; `VideoView` and `ZoomableContentView` are byte-identical with `main`.

## Trade-off worth a maintainer's eye

Feed video now streams at the rung matching the card's real pixel size instead of the bottom one. On a 3x-density phone a full-width card is ~1200px wide, so an autoplaying feed will pull noticeably more data than before. That is the point of the change, but `ExoPlayerBuilder.feedTunedLoadControl()` already notes that several players preload at once — if that proves too aggressive in the wild, the lever is now a single `setViewportSize` call rather than a policy threaded through every call site.

## Tests

14 new tests for the imeta selection (master first/last, dim-less master, sibling metadata fill, no master published, HLS vs mp4, extension-only detection, the `?ref=a.m3u8` false positive, audio/poster skips) and 5 for the viewport update guard. Full suite green: 1408 amethyst + 1710 commons.


vitorpamplona:
@davotoula let me know what you think about this. 

Closes: https://gitworkshop.dev/vitorpamplona.com/relay.ngit.dev/amethyst/issues/nevent1qy28wumn8ghj7un9d3shjtnwva5hgtnyv4mqqgxyeagm2untuafykspee06urhg0v6c6wf5hjf87fuhfj0mu3fzmfutdtvk9

vitorpamplona:
`test-and-build-android` is red on `421b25bb`, but the failure is inherited from `main`, not from this PR.

**Failing task:** `:amethyst:lintFdroidBenchmark` — `ExtraTranslation`, 30 errors, first one:

```
amethyst/src/main/res/values-th/strings.xml:515: Error: "route_video" is translated
here but not found in default locale [ExtraTranslation]
```

The unit tests in the same job passed: 9226 run, 9099 passed, 127 skipped, **0 failed**.

**Why it isn't this PR's:** the base commit (`2c4cf9d1`) is red on its own push build for the same reason — [run 33444776622](https://github.com/vitorpamplona/amethyst/actions/runs/33444776622) — and this PR touches no files under `res/`.

**Cause:** d6d5a72e ("stop labelling the mixed media feed Shorts") retired `route_video` → `route_media` and `new_short` → `new_media` in the default locale. Its commit message notes *"Crowdin drops the retired keys on its next sync and the label falls back to English until retranslated"* — but `ExtraTranslation` is a lint **error** and the project keeps no lint baseline, so it fires the moment a locale carries a key the default locale lacks, before any Crowdin sync gets a turn. The sync that did follow (#4027, now `main`) did not remove them.

Current orphan count across `values-*/strings.xml`:

| retired key | locales still carrying it |
|---|---|
| `new_short` | 55 |
| `route_video` | 46 |

**Proposed fix**, not applied here so this PR isn't widened — delete both retired keys from every locale file, which is what the next Crowdin sync would have done anyway:

```bash
find amethyst/src/main/res/values-*/ -name strings.xml \
  -exec sed -i -E '/<string name="(route_video|new_short)"/d' {} +
```

Same shape as #4017 ("fix(i18n): unbreak main's lint"). Happy to open that as its own PR if wanted — it unblocks every open PR, not just this one. Not re-running the job: the failure is deterministic and reproduces on the base branch.

---

davotoula:
Can I get more context? 

The feature was built so that in feed videos and pip (usually small viewport) are locked to the lowest quality but fullscreen uses exoplayer built in adaptive choosing based on Internet speed. 

Is there a bug are do you want to improve on the behaviour? I saw some tests for matching selected quality based on viewport size

davotoula:
From my clanker to your clanker:

Thanks for the writeup — I dug through both halves and pulled the media3 1.11.0 sources to check the claims. Short version: the imeta half is a clear win, the quality half is replacing a deliberate design and I'd like a couple of things back before it lands. I'd rather see them as two PRs.

## 1. `selectVideoTrack()` — agreed, with one ordering fix

This is orthogonal to anything I built and it's the right call. `HlsVideoEventBuilder` emits `listOf(masterVideoMeta) + renditionMetas` with the master's `dim` set to the largest rendition, so for our own uploads the new selector is a no-op — which is exactly what makes it safe to merge. The value is interop with publishers who order tags differently, and the notification poster fix is a nice extra.

One thing I think is wrong, though. Rule 4 is checked before rule 3:

```kotlin
hls.any { it.dimension == null } -> hls.first { it.dimension == null }
else -> hls.maxByOrNull { it.pixelCount() } ?: hls.first()
```

A dim-less entry wins unconditionally. But the common sloppy-publisher shape isn't "master without dim, renditions with dim" — it's the reverse: a master that declares its top resolution and a rendition or two that forgot `dim`. In that ladder this picks a single-rung media playlist, which is the exact failure the PR exists to fix, and it does it silently.

Suggest only preferring a dim-less entry when *no* HLS entry declares a dim:

```kotlin
hls.none { it.dimension != null } -> hls.first()
else -> hls.maxByOrNull { it.pixelCount() } ?: hls.first()
```

Also minor: `withLadderMetadataFrom(candidates)` fills from the filtered list, so a poster carried on an `image/*` imeta (which `canBeTheVideo()` excludes) never reaches the chosen entry. Probably fine given the `?: meta?.url` fallback, but worth a look.

## 2. Viewport sizing — the mechanism is right, the policy is a product decision

To be clear about what the old code was for, since the PR frames the `LOWEST` pin as a "counterweight": that's accurate, and I'd checked the same thing. `TrackSelectionParameters.Builder`'s init sets `isViewportSizeLimitedByPhysicalDisplaySize = true` with `viewportWidth = Integer.MAX_VALUE`, so every player is told it has the whole display. Pinning the lowest rung was the only lever I had that didn't require measuring the player. Handing media3 the measured size is strictly better than what I did, and I verified the safety argument too — `getMaxVideoPixelsToRetainForViewport` returns `Integer.MAX_VALUE` when no rendition exceeds the viewport, so a small player can't filter the group empty. No objection to the mechanism.

My concern is what it does in this app specifically. Almost all inline video in Amethyst renders at roughly content width, not just the media-card feeds — so "the rung matching its pixels" is near the top of the ladder for nearly every player, not only the full-bleed ones. The honest summary of the behaviour change is: autoplaying feeds go from pinned-lowest to adaptive-up-to-about-1080p, everywhere. The PR says as much in the trade-off section, and I don't think it's wrong to want that; I think it needs the counterweight rebuilt rather than removed.

Four asks:

**Metered connections.** `UiSettingsState` already carries `isMobileOrMeteredConnection` and uses it to gate autoplay and image loading. A viewport ceiling is orthogonal to that signal, so right now the change lands the full increase on mobile data with nothing to hold it back. Capping the viewport (or keeping something like the old pin) when metered would keep the "quality proportional to the player" idea intact while preserving the reason the low pin existed.

**Keep the diagnostics.** `2ed1b17f` added `Log.d("VideoQuality")` lines listing the available renditions and the one chosen, and they go away with `InitialVideoQualitySelector.kt` with nothing replacing them. Without them, a user reporting "Amethyst is eating my data" isn't a debuggable claim. A log on each viewport push and on `onTracksChanged` showing the selected format would cover it.

**Measure it, don't only reason it.** The five new tests exercise `needsViewportUpdate`'s integer arithmetic, which is worth having, but nothing in the suite shows a 1080px box actually resolving to the top rung or a small box to a small one. The before/after table is derived from reading `DefaultTrackSelector`, which matches my own reading — but given this is a bandwidth change, a real measurement (bytes over a fixed feed scroll, before and after) would settle it in a way source-reading can't.

**PiP.** `PipVideoActivity`'s explicit `LOWEST` is dropped in favour of `RenderPipVideo`'s measured size, but `processIntentForPiP` calls `enterPictureInPictureMode` from composition, so the first layout pass can measure full-screen and push a full-screen viewport before the window shrinks. On the `makeLaunchIntoPip` path that's probably invisible; on the pre-T `makeBasic()` path I'm less sure. Worth a check on device that PiP doesn't grab a high rung for the first few seconds.

## CI comment nit

The diagnosis is right — `ExtraTranslation` from `d6d5a72e`, inherited from main, not from this PR, and the 30-error count matches. The per-key table doesn't, though. At `2c4cf9d1` I count 15 locale files carrying each key, not 55 and 46:

| retired key  | locales carrying it at 2c4cf9d1 |
| ------------ | ------------------------------- |
| `new_short`  | 15                              |
| `route_video`| 15                              |

15 files x 2 keys = the 30 errors, so the total is right and the breakdown isn't. The cleanup PR is still worth opening on its own — it unblocks everything.

## Suggestion

Split it. Part 1 with the dim-less ordering fix I'd merge as-is. Part 2 I'd like to see with the metered-connection cap and the logging back in, plus a byte measurement, and then I'm happy with it — the goal is right, it's just that removing the only bandwidth lever and not replacing it is a bigger change than the diff makes it look.

davotoula:
In the meantime, I fixed the i18n lint issue on main and updated this brach with the fix...

vitorpamplona:
Thanks — and thanks for unbreaking main. Three of your asks are pushed in `d6913174`; three need a decision or a device.

**First, your context question:** yes, there was a real bug independent of the quality policy. Renderers took `imetaTags()[0]` blindly, so a publisher that doesn't put the master first pinned playback to one rung — and the quality menu disappeared too, since a media playlist exposes a single track. Part 2 is the behaviour change; Part 1 is the bug.

## Pushed

**Ordering.** You're right, and the failure mode is worse than "suboptimal": with a master declaring `dim` beside a rendition that forgot one, the old order picked the rendition — the exact bug the selector exists to fix. Taken as you wrote it: tag order decides only when *no* HLS entry declares a dim. Regression test `aDimlessRenditionNeverBeatsADimensionedMaster` pins it. The cost is the case rule 4 was for — a genuinely dim-less master beside dimensioned rungs now yields the top rung, so we keep the quality and lose the adaptation. That's the better way to fail, since the other direction lands on the bottom rung.

**Poster fill.** Real miss, fixed. Fills from every imeta rather than the playable candidates, and when no entry carries one in `image` it takes an `image/*` sibling's own url — which is where the poster actually lives for publishers that ship it as a separate entry, and what `mediaImageUrl` wants.

**Diagnostics.** Restored under the `VideoQuality` tag: every viewport push, and the rung selection landed on marked against the whole ladder (`1080x1920*, 720x1280, 360x640`). The listener registers only when `Log.minLevel <= DEBUG`, so debug builds get the trace and release keeps the no-listener-per-player property that dropping the old selector bought.

## Needs a call from @vitorpamplona

**Metered connections.** I agree with the principle and won't pretend otherwise: removing the only bandwidth lever without replacing it is the weak part of this PR. But what replaces it is a product decision, not a review nit, so I'd rather Vitor pick than have me quietly pick for him. Concretely I'd propose: when `isMobileOrMeteredConnection` is true, clamp the pushed viewport to a fixed short side (480p-ish) instead of the measured size — one `min()` at the single `setViewportSize` call, keeps "quality proportional to the player" on wifi, keeps the old restraint on data, and stays one lever rather than a policy per call site.

## Needs a device, which I don't have here

**Byte measurement.** Fair, and I can't do it from this environment — no emulator or handset, and the numbers in the PR are read off `DefaultTrackSelector`, not measured. I'm not going to dress that up as verification. If you can run a fixed feed scroll on debug builds either side of `d6913174`, the `VideoQuality` trace now gives per-player rungs to go with the byte totals. I can add an instrumented test that asserts a chosen rung against a synthetic ladder if that's more useful than a manual run.

**PiP.** Confirmed by construction, not just plausible: `PiPFromIntents.kt:45` calls `processIntentForPiP` inside `remember { mutableStateOf(...) }`, so `enterPictureInPictureMode` runs during first composition — before any layout pass. The first `onSizeChanged` can therefore measure full-screen and push a full-screen viewport before the window shrinks. Proposed fix: don't push a viewport from `RenderPipVideo` until `isInPictureInPictureMode` is true, so the pre-shrink measurement never reaches the selector. I can't confirm the pre-T `makeBasic()` timing without a device, so I'd want your eyes on it after.

## Correction

My orphan-key table was wrong and yours is right: 15 locales each, not 55/46. I measured my checkout, which sat at `b335e299` — before the Crowdin sync in `7a4dc1b3` cleaned 32 of the 47 — and then cited the count against `2c4cf9d1`, where it was already 15/15. 15 × 2 = your 30 errors. The diagnosis held; the number didn't, and I should have measured the commit I was naming.

## Split

Your call and Vitor's, not mine — I'll do whichever. If you want them separate, Part 1 is `9639f97c` + `d6913174` and Part 2 is `421b25bb`, so it's a clean cherry-pick either way.

---

vitorpamplona:
@vitorpamplona says keep it in one PR, so no split — and the last two code asks are pushed in `7ed36811`.

**Metered clamp.** While `isMobileOrMeteredConnection` is true the pushed viewport is capped to a 480px short side, aspect preserved so it still describes the player's shape rather than just its area. That keeps "quality proportional to the player" on wifi and restores something like the old restraint on data, as one `min()`-shaped step at the single `setViewportSize` call rather than a policy per call site. Connectivity changes don't relayout, so a `LaunchedEffect` re-pushes with the last measured size when the ceiling flips; before the first measurement the existing zero-size guard makes that a no-op. `clampViewportShortSide` rounds *up*, so a rounding artifact can't ask for 479 and drop a rung sitting exactly at 480.

480 is a starting value, not a considered product decision — it's one constant (`METERED_MAX_SHORT_SIDE_PX`) if you want it somewhere else.

**PiP.** `RenderPipVideo` now withholds the push until `isInPictureInPictureMode` is true, so the pre-shrink full-screen measurement never reaches the selector; the shrink relayouts and pushes the real size. Evaluated at measure time rather than composition, since that's when the answer has to be current. As said before, I can't check the pre-T `makeBasic()` timing without a device — worth your eyes there.

4 new tests on the clamp (uncapped passthrough, portrait and landscape scaling by short side, the round-up floor, unmeasured/zero). 1412 amethyst + 1712 commons green.

Still outstanding and unchanged: the byte measurement over a fixed feed scroll, which I can't run from here. The `VideoQuality` trace gives per-player rungs to pair with byte totals if you take a debug build either side of this branch — and it'd now be worth running once on wifi and once on data, since those are two different behaviours after this commit.

---

vitorpamplona:
Ran a full audit over this PR's own diff before asking you to look again. Five defects, four confirmed with throwaway probe tests against the branch. Fixed in `54c6521f`, each with a regression test.

**1. An HLS master labelled `audio/mpegurl` was read as an audio track and dropped.** The worst of the five, and squarely my fault: `canBeTheVideo()` tested `!isAudio` before asking whether the entry was a playlist, but two of the four playlist MIMEs this repo already recognises in `isHlsMimeType`/`MediaItemCache` are spelled `audio/x-mpegurl` and `audio/mpegurl` — legacy aliases naming the manifest format, not the content. Probe: master(`audio/mpegurl`, 1080×1920) + rung(`vnd.apple.mpegurl`, 360×640) selected the 360p rung. When *every* entry used that MIME the candidate list emptied and selection fell through to `imetas.first()` — the poster JPEG, handed to the video player.

**2. The poster could set the video's aspect ratio.** Broadening the metadata fill to all imetas, as you asked, let `dimension` come from the poster too. Probe: dim-less master + poster `dim 1600×900` produced a 1600×900 master, and `JustVideoDisplay` sizes its box from that — a vertical short laid out at 16:9. Only entries that could be the video may describe its shape now; the poster still supplies the still.

**3. `master.m3u8` served as `application/octet-stream` was not recognised as HLS** and lost to a correctly labelled 360p rung. A server default is not a claim about the file, so the path extension gets a say again when the declared type is that uninformative.

**4. The metered cap had no fullscreen exemption** — tapping into fullscreen on mobile data pinned 480p, and put a ceiling the quality menu's "Auto" could not exceed, contradicting the table in this PR's description. The cap is there to hold back feeds that autoplay unasked; someone who tapped fullscreen asked.

**5. The PiP gate could leave a stale viewport.** Skipping the push until `isInPictureInPictureMode` had no retry path, and `ExoPlayerPool.demoteToCold` clears track overrides but not the viewport — so if PiP was never actually entered (per-app PiP off, no `FEATURE_PICTURE_IN_PICTURE`), a pooled player kept whatever its previous view pushed, for the session. It now caps the pre-shrink measurement rather than skipping it, so a viewport is always pushed and can never be a stale full-screen one.

1412 amethyst + 1716 commons green.

On the last run, `test-quartz-ios` failed on `PoolEventOutboxScaleTest.publishing stays flat as the outbox fills` — 1 of 3491, a timing/scale test in a module this PR changes zero files in, with main green at both `1590dc85` and `725bbc55`. I read that as a flake; a targeted re-run was refused with "workflow is already running" while the Android job was still going, and the new push supersedes it anyway. If it fails again on `54c6521f` I'll treat it as real rather than re-running.

---

PR #4093, opened 2026-09-12
feat(media): play YouTube links in the app, on YouTube's own IFrame player

vitorpamplona:
> **Stacked on #4092.** Base is `fix/og-media-playback`, so the diff here is only the YouTube work. GitHub will retarget this to `main` when #4092 merges. Please review #4092 first.

> **Draft:** nothing here has been run on a device — see *Not verified* below. That gap matters much more for this PR than for #4092, because the whole feature is a WebView interaction.

## Summary

YouTube publishes no media file, so ExoPlayer has nothing to open — #4092 deliberately refuses its `og:video`, which is an `/embed/` page typed `text/html`. The one supported way to play a video outside youtube.com is its IFrame player page, which needs a browser. Rather than embed a second WebView stack, this uses the one Amethyst already ships.

Tapping a YouTube link calls `FavoriteAppLauncher.launchUrl` on the player page, which opens `NappletBrowserActivity`: the `:napplet` process, the site's remembered Tor choice, and a per-account storage partition. **YouTube's scripts never share a process with the account or `LocalCache`.** The page is loaded from `youtube-nocookie.com`, which sets no tracking cookie until playback starts.

This is also the only ToS-compliant path — the IFrame embed keeps YouTube's controls and branding, and there is no background playback. Stream extraction (NewPipeExtractor) was considered and rejected: I checked its LICENSE, and it is **GPLv3 with no linking or classpath exception**, which under CLAUDE.md § Dependency Licensing is a stop for an MIT-licensed APK — the same reason TarsosDSP was removed.

Wired at the two places a YouTube link renders: the link-preview card, and `Video.kt`'s existing NIP-71 YouTube branch (which previously showed a thumbnail and handed off externally). Plain text links are left alone — they go through the shared `SharedClickableUrl`, and desktop has no in-app browser to route them to.

**`playYouTubeInApp` defaults to on.** `YouTubeInApp.open` returns false when it declines — setting off, or not a video link — and the caller then opens the URL exactly as before. Existing installs have no stored value, so **they get in-app playback on upgrade** and turn it off in Settings → Media.

Also adds **HTML5 fullscreen to the sandboxed browser**, which was broken for every site, not just YouTube: `BrowserChromeClient` had no `onShowCustomView`, so `requestFullscreen()` did nothing and the fullscreen button on any embedded player was a dead tap. Back now leaves the video rather than the page, and the surface is handed back on teardown — holding one of the WebView's views across `destroy()` is the renderer corruption that file already documents at length.

`autoplay=1` is passed but will usually be ignored: the browser keeps `mediaPlaybackRequiresUserGesture = true`, so the viewer taps play once inside the player. That setting is shared with every other site there and is not worth relaxing to save a tap — it is what stops any page from playing sound on open.

### Feature test plan

Verified with `curl` against the live hosts:

- `https://www.youtube-nocookie.com/embed/dQw4w9WgXcQ` → `200`, `text/html`, ~142 KB, `<title>YouTube</title>` — a real standalone player page, which is what makes loading it as a top-level URL work.
- YouTube `watch?v=…` declares `og:video:type: text/html`, which is why #4092 refuses it and this PR exists.

`YouTubeLinkTest` (7 tests) pins the parser against those spellings: watch, `youtu.be`, shorts, embed, live, `/v/`, `m.`, `music.`, `youtube-nocookie`; noise (`&list=`, `?si=`, `#fragment`, trailing slash); start offsets in all three formats YouTube writes (`t=90`, `t=90s`, `t=1h2m3s`, `start=`); and rejection of channel pages, `/feed/subscriptions`, bad ids and look-alike hosts (`youtube.com.evil.example`). The 11-character id rule is what keeps a channel page from opening a player showing nothing.

```
./gradlew :commons:jvmTest :amethyst:testFdroidDebugUnitTest :amethyst:compileFdroidDebugKotlin
BUILD SUCCESSFUL in 2m 52s
```

### Regression test plan

Touch points, failure mode, verification:

- **Every link-preview card** — `UrlPreviewCard`'s tap now tries `YouTubeInApp.open` first; if that mis-claimed a URL, no link would open in a browser again. Verified: `open` returns false unless the setting is on *and* `YouTubeLink.parse` returns a video, and the parser's rejection cases are tested above.
- **NIP-71 YouTube notes** — `Video.kt`'s existing branch changed from `openUri` to the same try-then-fall-back. Verified by inspection; the fallback is the original call, unchanged.
- **The in-app browser, all sites** — `onShowCustomView` is new shared behaviour in `NappletBrowserActivity`. Failure mode: a page requesting fullscreen leaves a black overlay you cannot dismiss, or the WebView is corrupted on teardown. Mitigated by exiting fullscreen in `onDestroy` before the existing WebView detach, refusing a second request while answering its callback, and routing back through `exitFullscreen`. **Not verified at runtime** — see below.
- **Settings persistence** — a new `BooleanType` was threaded through `UiSettings`, `UiSettingsFlow` (constructor, `listOfFlows`, the positional `propertyWatchFlow` mapping, `toSettings`, `update`, the `build` factory), and `UISharedPreferences`. Failure mode: a wrong index silently swaps two settings for everyone. Verified: the field was appended last everywhere, so every existing index is unchanged, and it reads `flows[30]`; amethyst's 1471 unit tests pass.
- **Rotation during playback** — `NappletBrowserActivity` already declares `configChanges="orientation|screenSize|…"`, so fullscreen will not recreate the activity.
- **Icon font** — `SmartDisplay` was already referenced in `MaterialSymbols.kt`, so no subset regeneration is required (CLAUDE.md § Icons).
- **Strings** — added only to `values/strings.xml`, leaving Crowdin to handle translations.

### Not verified, and why

**Nobody has watched a video play.** No emulator image and no device in this environment. Specifically unverified:

1. That the IFrame player renders and plays in that WebView, with Safe Browsing on, `MIXED_CONTENT_COMPATIBILITY_MODE`, and the per-account storage partition.
2. The whole fullscreen cycle — enter, rotate, back, exit, hand back. View-lifecycle code is exactly what only misbehaves at runtime.
3. Whether a feed video playing in the main process yields audio focus when the browser activity takes over.

That is why this is a **draft**. I would not merge it without someone installing `installFdroidDebug`, tapping a YouTube link, and exercising fullscreen.

Both flavours were built on the base branch (#4092); this PR adds no Google-proprietary touch points, but `installPlayDebug` / `installFdroidDebug` were not run because installing needs a device.

## Known gaps

Listed rather than hidden — none are blockers for review, all are follow-ups:

- **Two taps to start**, per `mediaPlaybackRequiresUserGesture` above.
- **Links with previews turned off don't route in-app** — they render through commons' `SharedClickableUrl`, shared with desktop. Needs a deliberate decision, not a quiet change.
- **The Video / Shorts / Longs feeds still blocklist `youtube.com`**, so YouTube NIP-71 events never appear there. That predates this work and made sense when YouTube couldn't play; worth revisiting now that it can, but it is a product call.
- **Nothing signals a card will play in-app** — a YouTube preview card looks like any other link card. `YouTubeInApp.handles()` exists to back such an affordance and currently has no caller; it should either get one or be deleted before merge.
- `YouTubeInApp` reads `Amethyst.instance.uiState` directly, which puts it out of reach of unit tests. Taking the flag as a parameter would make the routing decision testable.

## Interop suites

No events published, no kinds or tags, no MLS/DM/QUIC path. The audio it affects is decoded by YouTube's player in a WebView, not by us.

## AI assistance

Written with Claude Code. The code-review pass required by CONTRIBUTING-WITH-AI was run against the base commits and its six findings are fixed in #4092; this PR's own diff has **not** had an independent review pass yet, and given the *Not verified* section above I would rather that happen alongside on-device QA than before it.

## License

No new dependencies. NewPipeExtractor (GPLv3, no linking exception) was explicitly rejected — see Summary.

https://claude.ai/code/session_01S8uqrrzDpqUYxV9R7CJgxW

---
_Generated by [Claude Code](https://claude.ai/code/session_01S8uqrrzDpqUYxV9R7CJgxW)_

### DIFF
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettings.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettings.kt
@@ -65,6 +65,12 @@ data class UiSettings(
     // be bank or Venmo handles carrying legal names, and this puts them one tap
     // from every note in the feed.
     val showPayToZapChip: Boolean = true,
+    // Whether a YouTube link plays inside Amethyst -- in the sandboxed in-app browser, on
+    // YouTube's IFrame player page -- instead of being handed to the external YouTube app.
+    // Defaults to NEVER: the handoff is the behavior every existing user already has, the
+    // external app plays better (their account, quality, downloads), and an in-app YouTube
+    // page is YouTube's scripts running on a page we opened. Opt-in, not opt-out.
+    val playYouTubeInApp: BooleanType = BooleanType.NEVER,
 )
 
 enum class ThemeType(
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettings.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettings.kt
@@ -67,10 +67,12 @@ data class UiSettings(
     val showPayToZapChip: Boolean = true,
     // Whether a YouTube link plays inside Amethyst -- in the sandboxed in-app browser, on
     // YouTube's IFrame player page -- instead of being handed to the external YouTube app.
-    // Defaults to NEVER: the handoff is the behavior every existing user already has, the
-    // external app plays better (their account, quality, downloads), and an in-app YouTube
-    // page is YouTube's scripts running on a page we opened. Opt-in, not opt-out.
-    val playYouTubeInApp: BooleanType = BooleanType.NEVER,
+    // Defaults to ALWAYS: staying in the app is the point of the feature, and a link that
+    // bounces the reader out to another app loses them mid-feed. The page still runs in the
+    // `:napplet` process behind the account's Tor choice, so playing it in-app costs the
+    // reader nothing they were not already spending by opening it at all. Anyone who prefers
+    // the external app (their account, quality, downloads) turns this off in Settings > Media.
+    val playYouTubeInApp: BooleanType = BooleanType.ALWAYS,
 )
 
 enum class ThemeType(
--- a/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettingsFlow.kt
+++ b/amethyst/src/main/java/com/vitorpamplona/amethyst/model/UiSettingsFlow.kt
@@ -57,6 +57,7 @@ class UiSettingsFlow(
     val composeSignature: MutableStateFlow<String> = MutableStateFlow(""),
     val showOnchainWallet: MutableStateFlow<Boolean> = MutableStateFlow(true),
     val showPayToZapChip: MutableStateFlow<Boolean> = MutableStateFlow(true),
+    val playYouTubeInApp: MutableStateFlow<BooleanType> = MutableStateFlow(BooleanType.NEVER),
 ) {
     val listOfFlows: List<Flow<Any?>> =
         listOf<Flow<Any?>>(
@@ -90,6 +91,7 @@ class UiSettingsFlow(
             composeSignature,
             showOnchainWallet,
             showPayToZapChip,
+            playYouTubeInApp,
         )
 
     // emits at every change in any of the propertyes.
@@ -127,6 +129,7 @@ class UiSettingsFlow(
                 flows[27] as String,
                 flows[28] as Boolean,
                 flows[29] as Boolean,
+                flows[30] as BooleanType,
             )
         }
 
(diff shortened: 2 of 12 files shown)


davotoula:
Not doing this?

vitorpamplona:
It got too complicated. We are basically playing YouTube web (which kinda sucks and has no info on the users login with YouTube) on a new process and transferring the video feed to a webview in the activity.

davotoula:
Youtube links open quickly in Newpipe so maybe better to leave it to external player...

Issue #91, opened 2023-02-07
Umbrel Private Relay 


craigderington:
I'm using the new Umbrel private relay (nostr-rs-relay).  I have my Amethyst client relay list configured with my private relay address.  This address is only prefixed with ws://<address> and not a secure web socket.  All of my activity in the Amethyst app creates errors on this private relay.  It worked as expected when I initially saved the new relay, then started failing.

vitorpamplona:
Amethyst will force it to be a secure connection. So wss is forced for now when you edit your relay list in the app. 

Other clients might have similar restrictions.

craigderington:
Yeah I noticed it appended the string with the secure socket.  Thanks for your reply.

mayankchhabra:
@vitorpamplona it'd be great if you can add `ws://` support to Amethyst so it can connect to local relays, such as the one on Umbrel. We're getting lots of requests from users who want to connect Amethyst to their relay on Umbrel. Damus recently updated their client to support it too: https://github.com/damus-io/damus/issues/344 and https://github.com/damus-io/damus/pull/383. Keep up the great work!!

lukechilds:
Also @vitorpamplona just to add some more context, when connecting to a private relay on Umbrel users are either connecting directly over the local network or remotely via WireGuard (Tailscale). So although it's plaintext `ws://` from the apps perspective no insecure communication is happening over untrusted networks. It's either all local local plaintext, or if remotely then WireGaurd is providing an encrypted tunnel over the public internet.

vitorpamplona:
Is this issue only when adding relays through Amathyst settings? Does it work when you add them in another interface? 

Because I don't think we actively block the connection if the relay list is already setup in another platform. Unless we have a bug somewhere.

craigderington:
Yes, I am using Amethyst with my private relay.  When I initially added the private relay to amethyst when viewed in another client it appeared like this.  wss://was://

PR #4127, opened 2026-09-15
fix(blossom): send the auth token as standard padded Base64 again

davotoula:
## Problem

Since 1.15.0, roughly two Blossom uploads in three fail with the server error `invalid "Authorization": invalid base64 token`, on every Blossom server the reporter uses, and mirroring fails silently. Reported on nostr on 14 Sep 2026 (`nevent1qqsqqqqd3626guc7w2s92c93mnrx5a5ddews8pzx7mzm6lztj0445tsp…`, by setto@basspistol.org).

## Cause

f1c461dcfa (PR #3991) switched `BlossomAuthorizationEvent.rawToken()` from standard padded Base64 to BUD-11's "Base64 URL-safe without padding". Deployed Blossom servers decode the token with a strict standard decoder, which requires `=` padding and rejects the `-`/`_` alphabet. Padding is needed whenever the event JSON's byte length is not a multiple of three, which is two lengths in three and moves with every upload's `alt`, size and hash, so the failure looked file-dependent.

Probed on the reporter's server with hand-made tokens:

| Token | Server reply |
|-------|--------------|
| standard, padded | `broken event` (decoder passed; fake event rejected, as expected) |
| standard, unpadded | `invalid base64 token` |
| URL-safe, padded | `invalid base64 token` |
| URL-safe, unpadded | `invalid base64 token` |

A wider survey found blossom.primal.net, nostr.download and files.sovbit.host equally strict; nostr.build, blossom.band, 24242.io and others accept either form.

## Change

`rawToken()` goes back to standard Base64 with padding, the exact bytes 1.14.x sent and the same encoder NIP-98's `HTTPAuthorizationEvent` already uses. Every Blossom path (upload, media, mirror, list, delete, BUD-11 read auth, static-site publisher, desktop, `amy`) flows through this one function, so no call sites change. The KDoc records the conflict with the BUD-11 draft and the probe evidence so it is not "fixed" back to the spec without re-probing a strict server.

This deliberately does not follow BUD-11's MUST. The draft and the reference servers disagree; until they converge, the encoding every deployed decoder accepts wins. The spec-side question belongs on hzrd149/blossom (PR #91 introduced the MUST) and khatru's blossom package, and will be raised there separately.

## Also: Sovbit default server

The bundled Sovbit entry pointed at `cdn.sovbit.host`, which no longer resolves (DNS failure, `curl` exit 6), so uploads to it failed regardless of the token. The Blossom service lives at `files.sovbit.host`, a strict decoder like Primal. The default in `ServerName.kt` and the matching instrumentation test now point there. Users who added Sovbit under the old host will need to re-add it; the default list only affects new selections.

## Tests

The quartz test that pinned "no padding, URL-safe" is replaced by one that decodes every token through the strict standard decoder across five `alt` lengths, covering all three `length mod 3` cases, and asserts that at least one of them carries padding. It fails on `main` for two of the five. The Android read-auth provider test is flipped the same way.

| Suite | Result |
|-------|--------|
| `BlossomAuthorizationEventTest` | 9/9 |
| `BlossomReadAuthTokenProviderTest` | 10/10 |

## Field verification

- `amy blossom upload` with a throwaway key against the reporter's server, six file sizes covering every `mod 3` case: 6/6 answered `only media signed by whitelisted pubkeys are allowed`, i.e. the token decoded and the event parsed every time. Before the fix the same call fails with `invalid base64 token` on about two in three.
- Debug APK on a Pixel 9a: three composer uploads to blossom.primal.net, nostr.download and a lenient WordPress Blossom server, each blob downloaded back and its SHA-256 matched to the URL.
- Auto-mirror after upload (the same `BlossomClient.mirrorOrUpload` call "Mirror to missing" makes): one upload to the WordPress server, then the same hash verified present on blossom.primal.net and nostr.download with nothing else tapped. That is the path the reporter saw fail silently.


davotoula:
@vitorpamplona Bugfix...

Sakrecoer:
Thank youuuu

